PR #47, opened 2017-07-17
[CMake] Use CMake's FindMatlab.cmake as opposed to our own

### LINKED ISSUE #41: FindMATLAB.cmake fails for MATLAB 2017a
See https://github.com/UCL/STIR/issues/81

KrisThielemans:
The version supplied since CMake 3.3 seems to work fine, so we set this now as minimum CMake version. This should be fine as the SuperBuild needs 3.7 anyway.

Note that it uses Matlab_ROOT_DIR, not MATLAB_ROOT.

Fixes #41


KrisThielemans:
Needs testing on multiple platforms. don't merge yet! Add a comment where tested.

I've tested on CentOS7 (UCL CS system)

bathomas:
I ran the following (macOS 10.12.6):
```shell
git clone https://github.com/CCPPETMR/SIRF-SuperBuild
cd SIRF-SuperBuild/
git checkout FindMatlab
cmake -DSIRF_TAG=FindMatlab -DMatlab_ROOT_DIR=/Applications/MATLAB_R2016b.app
make
```
This built successfully. However, ```make -j``` failed, so maybe there is an issue with dependencies somewhere. I also got errors  that ```pyconfig.h``` does not exist. I deleted the directory to start afresh but unfortunately sourceforge appears down now. Apologies for being a bit vague.

KrisThielemans:
thanks Ben. No idea about the `make -j`. I can't see why this would have happened (certainly not in the PR) nor how to debug the CMake files...

The `pyconfig.h` problem is because CMake's FindPython stuff is largely broken on Mac. Our wiki says to use
```
    cmake -DPYTHON_LIBRARY=$(python-config --prefix)/lib/libpython2.7.dylib \
          -DPYTHON_INCLUDE_DIR=$(python-config --prefix)/include/python2.7 .
```
not sure how that would have to be for Python 3, but I can guess.

PR #62, opened 2017-09-27
Selecting shared_ptr by SIRF engines.

evgueni-ovtchinnikov:
Switched to the use of std::shared_ptr in xGadgetron and stir::shared_ptr in xSTIR. Generally, different SIRF engines can now use different shared_ptr (as specified in *_shared_ptr,h).


KrisThielemans:
@casperdcl  do you have any idea why github claims that this PR contains so many commits? This PR is based on master, which was merged to devel recently, from which this branch was created. I guess the git merge won't be a problem, but I'd like to understand this a bit more.

KrisThielemans:
@evgueni-ovtchinnikov I had a brief look at this. It seems a bit of a hack to me. I guess it works so I'm ok with merging it, but I'm generally quite warry of solving this stuff with `#define`. Also, it seems we have twice `object_handle.inl` (identical?) .

I'm not 100% sure, but I guess what you'd need is to have an extra template arg for your `ObjectHandle` class (and others?)

evgueni-ovtchinnikov:
Yes, files object_handle.inl are identical – I should have had just one of them in src/common, but I just do not know how to edit CMakeLists so that it is found.

Extra template arg would just be too much editing, with namespace it is just two lines per file.

evgueni-ovtchinnikov:
Could it be because I branched it from devel?

KrisThielemans:
FYI, modifying CMake files:
1. create src/common/include/SIRF/common or something
2. in xSTIR/CMakeLists.txt do something like
```
include_directories(${PROJECT_SOURCE_DIR}/src/common/include)
```
3. in your file
```
#include "SIRF/common/object_handle.inl"
```

I wouldn't do this right now, as we want to avoid the `#define` trick later (it's not good programming style, and I'm worried that we'll have a conflict at some point).

can you merge with master locally first. looks like you'll have conflicts.

evgueni-ovtchinnikov:
Merged locally, no conflicts.

KrisThielemans:
ok. please push

evgueni-ovtchinnikov:
Done

evgueni-ovtchinnikov:
Merged manually

PR #68, opened 2017-11-29
Update UserGuide.md

ckolbPTB:
Removed SimpleReconGadgetSet, because I could not find it in Gadgetron.

GenericReconImageArrayScalingGadget has more parameters than currently available/listed in SIRF.

### DIFF
--- a/UserGuide.md
+++ b/UserGuide.md
@@ -622,6 +622,8 @@ internal2 | internal4 | debug_folder | ""
 | | average_all_ref_S | "true"
 | | prepare_ref_always | "true"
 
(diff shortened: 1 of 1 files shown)


evgueni-ovtchinnikov:
It is actually a small gadget chain performing fully sampled reconstruction - see SIRF/src/xGadgetron/cGadgetron/gadget_lib.h. I explained it in DeveloperGuide, will copy explanation to UserGuide.

Thanks for spotting omissions in GenericReconImageArrayScalingGadget. I noticed also two parameters that this gadget actually does not have: perform_timing and verbose, and the default value of scalingFactor is wrong (must be 4.0). Will correct.

PR #152, opened 2018-03-13
PLS prior

rijobro:


KrisThielemans:
this PR seems very problematic. It has SIRFReg stuff, PLS Prior, and some files commited with conflict markers.
I'm guess this should be closed. @rijobro, is there anything here that needs to be saved?

PR #197, opened 2018-08-03
Resolves namespace ambiguity in use of shared_ptr

### LINKED ISSUE #195: Shared Pointer Conflict due to using namespace
In the file $SIRF_PATH/src/xGadgetron/cGadgetron/gadgetron_data_containers.h
the statement in line 46 "using namespace gadgetron;" calls a "using std::shared_ptr;" from the included file "cgadgetron_shared_ptr.h".

This causes a conflict with  the useage of stir shared pointers in $SIRF_PATH/src/xSTIR/cSTIR/stir_types.h and thus to a compile error

johannesmayer:
Closes #195 

Inclusion of cstir and cgadgetron files resulted in ambiguous use of shared_ptr due to 
using namespace statements in header files.

Replaced >using stir::shared_ptr; by >using namespace stir; in .cpp files. 
These .cpp are not included anywhere, so no importing of STIR object into the global namespace happens.

### DIFF
--- a/src/xSTIR/cSTIR/cstir.cpp
+++ b/src/xSTIR/cSTIR/cstir.cpp
@@ -26,7 +26,7 @@ limitations under the License.
 #include "stir_x.h"
 #include "stir_data_containers.h"
 
-using stir::shared_ptr;
+using namespace stir;
 
 static void*
 unknownObject(const char* obj, const char* name, const char* file, int line)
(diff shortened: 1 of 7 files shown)


evgueni-ovtchinnikov:
merged and edited the gadgetron side accordingly

@KrisThielemans: is `USING_NAMESPACE_STIR` deprecated?

KrisThielemans:
> is USING_NAMESPACE_STIR deprecated?
not officially, as it's everywhere really. However, I do prefer `using namespace stir` (as anyone with a compiler without namespace support should just go give up)

evgueni-ovtchinnikov:
adding
    #define WIN32_LEAN_AND_MEAN
eliminated the `WinSock.h` error

PR #201, opened 2018-08-09
brew reinstall boost instead of upgrade

rijobro:


paskino:
I'm seeing all OSX failures on travis due to boost being already installed. 
Shouldn't the same fix go in the SuperBuild?

PR #207, opened 2018-08-29
Add weighted average for SIRFRegImageDeformation class

johannesmayer:
New class `SIRFRegImageDeformationWeightedMean`.
Allows for weighted mean of type `SIRFImageDataDeformation`.
`SIRFRegImageDeformationWeightedMean` inherits from `SIRFRegImageWeightedMean`.

To use the existing methods, the members in the base class  `SIRFRegImageWeightedMean` were declared private to avoid being passed to the child class.
Keeping the member names of the child the same but changing the tpe allows to use base class methods without the need to redefine them.

Also adjusted std::cout output of  check function if averaging the data is justified to include the displacement field dimensions.

### DIFF
--- a/src/Registration/cReg/SIRFRegImageWeightedMean.cpp
+++ b/src/Registration/cReg/SIRFRegImageWeightedMean.cpp
@@ -112,8 +112,8 @@ void SIRFRegImageWeightedMean::check_can_do_mean() const
 
     // Print all the info
     std::cout << "\n\nChecking that images can be combined...\n\n";
-    std::cout << "im || datatype | ndim | nvox     | nx  | ny  | nz | dx      | dy      | dz  \n";
-    std::cout << "---++----------+------+----------+-----+-----+----+---------+---------+------\n";
+    std::cout << "im || datatype | ndim | nvox | nx  | ny  | nz | nt | nu | dx | dy | dz | dt | du \n";
+    std::cout << "---++----------+------+------+-----+-----+----+----+----+----+----+----+----+----\n";
     for (unsigned i=0; i<_input_images.size(); i++) {
         std::cout << i << "  || ";
         std::string data_type = nifti_datatype_string(_input_images[i].get_image_as_nifti()->datatype);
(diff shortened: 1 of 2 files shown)


johannesmayer:
Sure your method looks better, I'll implement that.
It'll take me a couple of days since I got other stuff on my plate.

As chance would have it everything I do is in float precision.

johannesmayer:
Yo, please never accept that PR haha,
bad segfaul action is going on, I'll come back at you tomorrow with the 
fix grrrr

jo

--
Johannes Mayer
Quantitative MRI
Physikalisch-Technische Bundesanstalt
Abbestr. 2-12
10587 Berlin, Germany

johannes.mayer@ptb.de
phone: +49  30 3481 7233

Please visit www.ptb.de 




Von:    "Richard Brown" <notifications@github.com>
An:     "CCPPETMR/SIRF" <SIRF@noreply.github.com>
Kopie:  "johannesmayer" <johannes.mayer@ptb.de>, "Author" 
<author@noreply.github.com>
Datum:  29.08.2018 16:24
Betreff:        Re: [CCPPETMR/SIRF] Add weighted average for 
SIRFRegImageDeformation class (#207)



Thanks for this! I wonder if it's necessary to have two separate classes 
to do this though. Since we just loop over all voxels in the image, it 
shouldn't matter if ndim=3 or ndim=5 (deformation images etc):
for (unsigned j=0; j<_output_image.get_image_as_nifti()->nvox; j++) {
    // Add in the weighted contribution of the jth voxel of the ith image
    output_data_ptr[j] += input_data_ptr[j] * normalised_weights[
}

If we replaced std::vector<SIRFImageData> with 
std::vector<shared_ptr<SIRFImageData> >, then we should be able to do the 
average regardless of whether they are SIRFImageData or 
SIRFImageDataDeformation. Would you be able to give that a go?
I've also noticed something rather bad. I assume that all data are of type 
float, such that I do this cast:
float *input_data_ptr = 
static_cast<float*>(_input_images[i].get_image_as_nifti()->data);

I'll open an issue on this and fix it ASAP. For the meantime, the method 
will only work if your datatypes are floats!
?

rijobro:
Just had a go. Do `git pull` to get my changes. I only did limited testing, but there's a `ctest` that seems to work with no problems. (Fixed the data casting as well.)

If you're happy, feel free to close.

PR #193, opened 2018-07-30
GeometricalInfo - implementing Geometry (Hackathon)

ashgillman:
TODO:
- [ ] Add `ImageData` class between `aContainer` and `*ImageData`.
- [ ] Add `frame_of_reference_uid` to GeometricalInfo. Can only resample (without registration) between images if it is the same.

***

This is the preliminary work from the hackathon and continuation over the weekend.

PR #226, opened 2018-10-25
SIRF image classes hierarchy

evgueni-ovtchinnikov:
Bare bones of the new image data classes hierarchy. Need to discuss a proper place for geometrical info.


rijobro:
Great! Can geometrical info not just live in `SIRFImageData`?

evgueni-ovtchinnikov:
could add abstract base class for geometrical info, e.g. `SIRFGeometricalInfo`, and protected data member of type `std::shared_ptr<SIRFGeometricalInfo>`

evgueni-ovtchinnikov:
puzzled that all travis builds failed - built ok on my laptop

evgueni-ovtchinnikov:
@casperdcl : travis builds fail because of `git submodule update` failure - any idea what went wrong this time?

KrisThielemans:
> could add abstract base class for geometrical info, e.g. SIRFGeometricalInfo, and protected data member of type std::shared_ptr<SIRFGeometricalInfo> in SIRFImageData

#193 has this abstract class already. That PR isn't ready yet though.

We found out in STIR that it can be hard to have abstract base classes as members in other abstract base classes. It can lead to a lot of casting. I think a better solution is to have a member:
```c++
virtual GeometricalInfo& get_geometrical_info() const = 0;
```
That way, a derived class can store whatever geo-info derived class it needs and actually overload with
```c++
virtual VoxelisedGeometricalInfo& get_geometrical_info() const;
```
Obviously, you cannot avoid casting all the time (i.e. if you have a `SIRF::ImageData` and want access to the `VoxelisedGeometricalInfo`, you'll have to cast somehow.

another advantage is that you make enforce that the derived class implements this, and therefore actually knows it should initialise a geo-info object.

You could add this to the skeleton by creating a dummy `GeometricalInfo` class, such that you can do this before #193 makes progress.

rijobro:
> could add abstract base class for geometrical info, e.g. `SIRFGeometricalInfo`, and protected data member of type `std::shared_ptr<SIRFGeometricalInfo>` in `SIRFImageData`

Why would we need to have an abstract base for geometrical info? I thought the idea was that the geometrical info class should contain enough information to be able to switch between different types. In this fashion, we would have the same geometrical info class for all image types.

src/common/include/SIRF/common/mr_image_data.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#include "sirf_image_data.h"
```
KrisThielemans: would have to be `"SIRF/common/ImageData.h"`. Path is important, and filename has to be same as the class.

src/common/include/SIRF/common/mr_image_data.h
```diff
@@ -0,0 +1,12 @@
...
+
+namespace sirf {
+	template <typename T>
+	class MRImageData : public SIRFImageData<T>
+	{
+	public:
+		virtual ~MRImageData() {}
```
KrisThielemans: not necessary. we need a virtual destructor in the base class, therefore none in the derived classes (unless it needs to do something special)

src/common/include/SIRF/common/sirf_image_data.h
```diff
@@ -0,0 +1,17 @@
+#pragma once
+
+#include "data_container.h"
+
+/*
```
KrisThielemans:
```suggestion
/*!
```
for doxygen

src/common/include/SIRF/common/sirf_image_data.h
```diff
@@ -0,0 +1,17 @@
...
+\ingroup SIRF Image Data classes
```
KrisThielemans: doxygen groups needs to be defined somewhere. i think needs to be a single word.

src/xGadgetron/cGadgetron/gadgetron_data_containers.h
```diff
@@ -418,7 +420,7 @@ namespace sirf {
 
 	Images are stored in an std::vector<shared_ptr<ImageWrap> > object.
 	*/
-	class ImagesVector : public MRImageData {
+	class ImagesVector : public GadgetronImageData {
```
KrisThielemans: if derived from `GadgetronImageData`, the class `ImagesVector` would have to change name.  Possibly could just be templated

src/common/include/sirf/common/image_data.h
```diff
@@ -0,0 +1,49 @@
...
+			virtual Iterator_const& operator++() = 0;
+			virtual const aNumRef& operator*() const = 0;
+			virtual bool operator==(const Iterator_const&) const = 0;
+			virtual bool operator!=(const Iterator_const&) const = 0;
+		};
+		virtual Iterator& begin() = 0;
+		virtual Iterator_const& begin() const = 0;
+		virtual Iterator& end() = 0;
+		virtual Iterator_const& end() const = 0;
+		void copy(Iterator_const& src, Iterator& dst, Iterator& end)
```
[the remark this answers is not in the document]
KrisThielemans: but fine to leave it for now of course.

src/common/include/sirf/common/num_ref.h
```diff
@@ -0,0 +1,323 @@
...
+			return type_;
+		}
+		virtual void set_ptr(void* ptr)
+		{
+			//std::cout << (size_t)ptr << '\n';
+			ptr_ = ptr;
+		}
+		virtual void copy(const aNumRef& a_ref)
+		{
+			const NumRef& ref = (const NumRef&)a_ref;
```
KrisThielemans: why do we now that `a_ref` is a `NumRef`? Would need to use `dynamic_cast` to be sure

DANAJK:
I think you are enabling the casting of complex ISMRMRD data to float, taking either the absolute value or the real part, depending on `abs_`, which in turn depends on the value of `m`? Going from complex to the absolute or real value is quite an important step and concealing it within a type cast might be confusing.

src/xGadgetron/cGadgetron/cgadgetron.cpp
```diff
@@ -94,6 +86,17 @@ fileNotFound(const char* name, const char* file, int line)
 	return (void*)handle;
 }
 
+bool file_exists(std::string filename)
```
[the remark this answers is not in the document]
evgueni-ovtchinnikov: makes no tangible difference imho

src/common/include/sirf/common/image_data.h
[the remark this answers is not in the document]
evgueni-ovtchinnikov: this copy is just an alternative to std::copy whereby the range is defined by destination

src/common/include/sirf/common/num_ref.h
evgueni-ovtchinnikov: actually should have been

    void copy(const NumRef& ref)

and no cast; will do after merging new `num_ref.h` by Richard

KrisThielemans:
> i believe abs is actually our only option as pet images obtained from mr images have to be non-negative

I haven't check the usage of `abs` (didn't find the time yet), but I don't understand what @evgueni-ovtchinnikov is saying above. First of all, PET images are not always non-negative (they aren't with FBP and also some recent iterative algos). but what does "obtaining a PET image from an MR" mean? We will never convert an MR image to a PET or vice versa. Instead, the MR image might be used as a prior, or basis for filtering the PET, or whatever.

evgueni-ovtchinnikov:
so far as i can see, PLSPrior.set_anatomical_image takes an argument of type `shared_ptr<DiscretisedDensity<3,elemT> >`, so conversion is needed

src/xSTIR/cSTIR/stir_data_containers.h
```diff
@@ -465,64 +483,188 @@ namespace sirf {
...
+			mutable FloatRef _ref;
+			std::shared_ptr<Iterator_const> _sptr_iter;
+		};
+		STIRImageData() {}
+		STIRImageData(const ImageData& id)
+		{
+			Dimensions dim = id.dimensions();
+			int nx = dim["x"];
+			int ny = dim["y"];
+			int nz = 1;
```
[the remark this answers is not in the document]
rijobro: I guess you did this because we don't necessarily know what the 3rd dimension of the MR data is?

So, if I had an image that had `z=100` for three time points, I would create a `STIRImageData` with 300 slices? Is this desired? 

Could we throw an error if `ndim > 3`?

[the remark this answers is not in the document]
evgueni-ovtchinnikov: images reconstructed by gadgetron normally have 4 dimensions: nx, ny, nz (never saw nz > 1) and number of slices

note also that i actually added that piece of code merely to test conversion (constructing from an abstract image using abstract iterators) - dimensions and other geometrical metadata will have to be handled thoroughly (i believe Ashley is working on it on his branch)

KrisThielemans:
> so far as i can see, PLSPrior.set_anatomical_image takes an argument of type shared_ptr<DiscretisedDensity<3,elemT> >, so conversion is needed

ah, yes, we would need to be able to get from `MRImageData` to `DiscretisedDensity<3,elemT>`, probably via `STIRImageData`. But that doesn't mean that the converted image is now a PET image. Ideally, we'd set the `ImagingModality` of `stir::DiscretisedDensity<3,elemT>` to `MR`. It certainly doesn't imply anything on positivity.

This does bring to light a problem with the current hierarchy: `ImageData->PETImageData->STIRImageData`. STIR images don't *have* to be PET. I suppose that ISMRMRD images don't have to be MR either. hmm

PR #230, opened 2018-11-04
[FIX] Build error with gcc 7.3.0

ashgillman:
Just a small one. I guess I am using a newer c++ standard? But build with this.

### DIFF
--- a/src/Registration/cReg/NiftiImageData.h
+++ b/src/Registration/cReg/NiftiImageData.h
@@ -31,6 +31,7 @@ limitations under the License.
 #define _NIFTIIMAGEDATA_H_
 
 #include <nifti1_io.h>
+#include <vector>
 #include <string>
 #include <memory>
 #include <iostream>


ashgillman:
Builds after the first commit. Import into Python after the second (haven't actually tested anything just yet).

The first is probably newer gcc. The second probably has to do with my build system, Nix.

KrisThielemans:
With CMake, dependencies should be carried recursively. If SIRFReg depends on SIRF's `common`, then we should need to add it only there (and depending libraries/ executables will know). It seems [commented out](https://github.com/CCPPETMR/SIRF/blob/daed8676e9ee2ea06478b58c75817869f0350812/src/Registration/cReg/CMakeLists.txt#L32) though.
Similarly for the `STIR_LIBRARIES` stuff. I'd guess that `cstir` should depend on them, but i haven't tracked that.

Note that executables currently would need to depend on `STIR_REGISTRIES` as otherwise they won't be able to use STIR IO.

ashgillman:
Does SIRF common depend on STIR though? The linker error was a missing SIRF lib. Do you mean it makes sense then for csirfreg to depend on cstir?

ashgillman:
Oh it looks like SIRFreg does target_link_libraries to cstir (sorry no link, on mobile). But still, this line fixes my error - so why?

rijobro:
I linked `cstir` to be able to use `PETImageData`, and I would link `cgadgetron` to get `MRImageData`. At one point I linked `common` to get the `GeometricalInfo` stuff, but I think I removed it because the PR hadn't been merged.

I don't link against `STIR_REGISTRIES` because none of my executables currently require STIR IO.

rijobro:
With the new data structure, I should be able to remove `cstir` and `cgadgetron`, since the `NiftiImageData` constructor will simply accept the abstract base class, `ImageData`, as an argument.

rijobro:
So @ash, do I really need the `${STIR_LIBRARIES}`? I'm happy to include vector obviously.

ashgillman:
Hmm, maybe I should split into two PRs.

Kris is right, it doesn't make sense. I will try with maybe cstir instead of ${STIR_LIBRARIES} when I get to work which might be marginally better.

ashgillman:
https://github.com/CCPPETMR/SIRF/pull/211#issuecomment-435642579

This is the linker error I added it for. Can you think of any reason SIRFReg depends on STIR::Interview?

KrisThielemans:
`Interfile` I guess.
some ideas

- Are you using shared libs? 
- I have no idea how nix and CMake are related, but @rijobro is using plain CMake.
- Due to circular dependencies in STIR libraries, it can sometimes be that strange linking errors appear, even with static libraries. This looks like it just needs an extra IO dependency. if cstir depends on STIR_LIBRARIES, then maybe the latter needs yet another IO occurence...

ashgillman:
Yes Interfile sorry - my phone's autocorrect.

1. I believe so, as in `-DCMAKE_BUILD_SHARED=ON` (or similar, going from memory). Yes this likely was enabled.
2. Nix just calls CMake with a few extra default flags (I think shared libs by default, and custom install dirs) and everything is in a different path, but otherwise nothing special. The possible complication is that Nix builds everything with the source in a temp dir, so the source is no longer available at runtime (has caused some STIR issues I worked around). But it doesn't fundamentally change anything.
3. Hmm, I see. Although, even without the above changes I /am/ able to use sirf.STIR without problems. Not sure if that fact is relevant though

KrisThielemans:
you report 2 errors on that comment, a run-time error (shared libs) and a compilation error (could be shared or static). The circular dependencies can cause run-time errors (not sure how to resolve those) and linking errors (just listing them again solves this).

it'd be weird if this would be the cause, but possibly you could give it a go (as cstir does depend on `STIR_LIBRARIES`).

by the way, a `make VERBOSE=On` would tell you what libraries it's currently linking with. If the STIR ones are there already, then it's the circular dependency that is causing it.

ashgillman:
Sorry, actually that is one error. All the linking works at build time, but the library is not found at runtime. In the second command I was trying to demonstrate the library wasn't linked, I meant to actually run `ldd` not `ld` I think (I'm not all over the theory here).

However, maybe I am actually static linking - ldd on the .so file shows links to boost, itk, vtk, etc., but not stir. `nm` though:

```
$ nm /home/gil2a4/dev/CCPPETMR@github.com/SIRF/install/python/sirf/_pysirfreg.so | grep InterfileHeader      
                 U _ZN4stir15InterfileHeaderC1Ev
00000000001f53b0 W _ZN4stir15InterfileHeaderD0Ev
00000000001f4e60 W _ZN4stir15InterfileHeaderD1Ev
00000000001f4e60 W _ZN4stir15InterfileHeaderD2Ev
00000000001f4bd0 W _ZN4stir22MinimalInterfileHeaderD0Ev
00000000001f4af0 W _ZN4stir22MinimalInterfileHeaderD1Ev
00000000001f4af0 W _ZN4stir22MinimalInterfileHeaderD2Ev
                 U _ZNK4stir22MinimalInterfileHeader17get_exam_info_ptrEv
00000000006a8a98 V _ZTIN4stir22MinimalInterfileHeaderE
0000000000446160 V _ZTSN4stir22MinimalInterfileHea$ nm /home/gil2a4/dev/CCPPETMR@github.com/SIRF/install/python/sirf/_pysirfreg.so | grep InterfileHeader      
                 U _ZN4stir15InterfileHeaderC1Ev
00000000001f53b0 W _ZN4stir15InterfileHeaderD0Ev
00000000001f4e60 W _ZN4stir15InterfileHeaderD1Ev
00000000001f4e60 W _ZN4stir15InterfileHeaderD2Ev
00000000001f4bd0 W _ZN4stir22MinimalInterfileHeaderD0Ev
00000000001f4af0 W _ZN4stir22MinimalInterfileHeaderD1Ev
00000000001f4af0 W _ZN4stir22MinimalInterfileHeaderD2Ev
                 U _ZNK4stir22MinimalInterfileHeader17get_exam_info_ptrEv
00000000006a8a98 V _ZTIN4stir22MinimalInterfileHeaderE
0000000000446160 V _ZTSN4stir22MinimalInterfileHeaderE
                 U _ZTVN4stir15InterfileHeaderE
00000000006a8ab0 V _ZTVN4stir22MinimalInterfileHeaderE
derE
                 U _ZTVN4stir15InterfileHeaderE
00000000006a8ab0 V _ZTVN4stir22MinimalInterfileHeaderE
```

So I'm not sure really what this shows, but I think the `U` means undefined, so most STIR definitions in the file are defined, but there are three undefined ones here. I am guessing that if `nm` shows a symbol is undefined, its expected that at runtime one of the linked libraries (the ones `ldd` shoes) is expected to provide it.

But since most of my STIR objects are defined, I guess we actually tried to statically link them? Could this perhaps be down to a missing import also somewhere?

ashgillman:
I split the PRs (#232) out, so potentially this can be merged to fix the build error, ignoring the runtime one

rijobro:
Merging this one, thanks.

PR #237, opened 2018-11-14
Add CIL methods to SIRF classes

paskino:
Follow up from https://github.com/CCPPETMR/Hackathon-SIRF/pull/13 
addresses requests from @evgueni-ovtchinnikov


examples/Python/PET/acquisition_model.py
```diff
@@ -150,6 +150,16 @@ def main():
 
     back_projected_image_as_array = back_projected_image.as_array()
     show_2D_array('Backprojection', back_projected_image_as_array[z,:,:])
+
```
evgueni-ovtchinnikov: remove trailing blanks here

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1193,6 +1201,18 @@ def backward(self, ad, subset_num = 0, num_subsets = 1):
...
+        am = type(self)()
+        am.set_up( self.acq_templ, self.img_templ )
+        return am
+    def direct(self, image, subset_num = 0, num_subsets = 1, ad = None):
+        return self.get_linear_acquisition_model().forward(image, \
+                                             subset_num=subset_num, \
+                                             num_subsets = num_subsets, \
+                                             ad = ad)
+    def adjoint(self, ad, subset_num = 0, num_subsets = 1):
+        return self.backward(ad, subset_num = subset_num,
```
evgueni-ovtchinnikov: remove trailing blanks here

paskino: I should have

examples/Python/PET/acquisition_model.py
```diff
@@ -151,6 +151,16 @@ def main():
     back_projected_image_as_array = back_projected_image.as_array()
     show_2D_array('Backprojection', back_projected_image_as_array[z,:,:])
 
+    acq_model.direct(image, 0, 4, simulated_data)
```
KrisThielemans: Please add a comment say what `direct` does. Non-obvious terminology for most of us.

examples/Python/PET/acquisition_model.py
```diff
@@ -151,6 +151,16 @@ def main():
...
+
+    # show simulated acquisition data
+    simulated_data_as_array_direct = simulated_data.as_array()
+    show_2D_array('Direct projection', simulated_data_as_array_direct[0,:,:])
+    back_projected_image_adj = acq_model.adjoint(simulated_data, 0, 4)
```
KrisThielemans: empty line above and add comment about `adjoint`

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -388,6 +388,16 @@ def axpby(a, x, b, y):
             (a.real, a.imag, x.handle, b.real, b.imag, y.handle)
         return z;
 
+    def copy(self):
+        '''alias of clone'''
+        return self.clone()
+    def conjugate(self):
+        '''Returns the complex conjugate of the data '''
```
KrisThielemans: seems to be modifying the object itself as well. Is this intentional? If so, needs to be in the docstring

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -1202,6 +1212,10 @@ def backward(self, ad):
             (self.handle, ad.handle)
         check_status(image.handle)
         return image
+    def direct(self, image):
```
KrisThielemans: add docstring

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -1202,6 +1212,10 @@ def backward(self, ad):
...
+        return self.forward(image)
+    def adjoint(self, ad):
```
KrisThielemans: add docstring

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1141,6 +1144,11 @@ def set_up(self, acq_templ, img_templ):
         '''
         assert_validity(acq_templ, AcquisitionData)
         assert_validity(img_templ, ImageData)
+
+        # temporary save the templates in the class
```
KrisThielemans: grammar: "temporarily". However, it doesn't seem to be temporarily but permanently. So delete "temporary".

This has potentially impact on memory though. Do we really need to do this? They don't seem to be used anywhere.

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1193,7 +1201,19 @@ def backward(self, ad, subset_num = 0, num_subsets = 1):
             (self.handle, ad.handle, subset_num, num_subsets)
         check_status(image.handle)
         return image
-
+    def get_linear_acquisition_model(self):
```
KrisThielemans: docstring. Is this a CIL function?

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1193,7 +1201,19 @@ def backward(self, ad, subset_num = 0, num_subsets = 1):
...
+        am = type(self)()
+        am.set_up( self.acq_templ, self.img_templ )
+        return am
+    def direct(self, image, subset_num = 0, num_subsets = 1, ad = None):
```
KrisThielemans: docstring. what's `ad` anyway?

examples/Python/PET/acquisition_model.py
mehrhardt: Also for me the terminology is not obvious. Especially with the background term being sometimes added and sometimes not. Perhaps this needs to be changed so that at least for some people this is intuitive?

KrisThielemans:
I guess the `direct` terminology is imposed on us by CIL. @paskino, is that the case? If so, we cannot change it I guess. It will need careful documenting.

As `direct` is (or should be) different from `forward` , and we need it to implement certain algorithms as far as I recall (correct?), it needs to be implemented in C++ and Matlab.  That'll need @evgueni-ovtchinnikov's help.

paskino:
I believe we should involve @jakobsj , @dkazanc and the mathematicians in UoM.

mehrhardt:
I think one problem we have here is that the naming comes in part from the underlying mathematical objects and in part from the PET imaging application.

To me, if we have an inverse problem, defined by some data `y \in Y` and a mapping `A : X -> Y`, then computing `A(x)` is called the `direct` or `forward problem`. This problem can be linear or non-linear. In ccpi, `A` is modelled as a linear operator so it makes sense to call `A.direct(x) := A(x)` and `A.adjoint(x) := A*(x)`. For the MRI part of SIRF, this remains the same. However, for the PET part, the model `A` has been chosen to be non-linear (affine linear) as it includes the background term, too. Now, a non-linear `A` can have both methods `direct` and `forward` but both should compute `A(x)` and it should not have an `adjoint` (as it is not linear).

I see two ways to resolve this conflict.

First, the PET model could be changed to be linear and all of these problems disappear. The PET model `A` could still have a method like `simulate_data` or `forward_affine` to perform the affine model and `get_background` to get the vector of background coefficients.

Second, we could leave the PET operator in SIRF non-linear `A(x) = Bx + c` (should not have an adjoint method! only "backward" as this is mathematically not well-defined) and instead has methods like `get_linear_model`=B, `get_offset`=c and these are used in algorithms which explicitly use the linearity of the operator such as SPDHG and all of ccpi.

Either way is fine but I suppose we need to be very clear what we are doing.

paskino:
We developed [`get_linear_acquisition_model`](https://github.com/CCPPETMR/SIRF/blob/add_to_sirf_classes/src/xSTIR/pSTIR/STIR.py#L1204) as base for [`direct`](https://github.com/CCPPETMR/SIRF/blob/add_to_sirf_classes/src/xSTIR/pSTIR/STIR.py#L1208).

paskino:
What about the `adjoint` or `backward`?

jakobsj:
I haven't followed this conversation, but would it be helpful to distinguish between nonlinear and linear operators? 

Could let `Operator` be the more general class (to represent operators for which linearity cannot be assumed, such as the affine PET model you are discussing) having relevant functions/methods, perhaps including `forward` or `direct` but definitely not `adjoint`. Then, perhaps as a subclass, there could be `LinearOperator`, having both `direct` and `adjoint` functions/methods. I don't know if it would be preferable to use the same name such as `direct` in both `Operator` and `LinearOperator` or not.

jakobsj:
Seems one would then be able to construct the affine SIRF `Operator` from a `LinearOperator` and a perhaps a `DataContainer` holding the offset.

paskino:
More than a `DataContainer` shouldn't it be a `ConstantOperator`? But I think it's a good idea to develop a general `Operator` and a subclass `LinearOperator`. Please foster discussion in UoM.

src/xGadgetron/pGadgetron/Gadgetron.py
paskino: I guess that's a mistake. We should return a new object like this
```python
out = self.clone()
out.fill(self.as_array().conjugate())
return out
```

jakobsj:
I don't have a good understanding of the nonlinear SIRT operator and what the "offset" refers to. But reading above it appears to be (chosen to be) affine `A = Bx + c`. Here I understand `B` as a `LinearOperator` and `x` as an `ImageData`. Then the result `Bx` I imagine will be an `AcquisitionData`. In order to add the offset `c` to that I guess it should be an `AcquisitionData` as well? I don't think it should be an operator, but it might make sense to have something like a `ConstantData` type holding essentially a constant which is then (by `+`) broadcast onto whatever size/datatype `Bx` has.  Not sure that is required though, as we might create `c` as an `AcquisitionData` using a method like `get_offset` maybe called on construction of the `Operator` `A`.

KrisThielemans:
@jakobsj you're correct that in PET `A = Bx + c` , `x` is an `ImageData` and `c` an `AcquisitionData`. 

If we'd want to be able to add ``Operator`s, then it'd make sense to have `ConstantOperator` (I think that terminology is more appropriate than `ConstantData`), but I thought CIL didn't go for adding of (objective) functions and all that. I don't see an urgent need for it at present.

KrisThielemans:
(caveat: github markdown doesn't support math notation.)

We've chosen in SIRF to use terminology for non-math people, as that's most of our target audience. Hence the `forward` and `backward` terminology.

I'd like to stress that we really don't want to be restricted to linear and affine operators. As soon as we introduce dynamics etc (or indeed transmission tomography), all forward models will be non-affine. In that case, many algorithms need an operation corresponding to the gradient, i.e. something like
```python
g = obj_func.gradient(x)
```
Supposing the objective function is something like `distance(acq_data, acq_model.forward(x))` (plus penalties), obviously this would lead to
```python
est_data = acq_data, acq_model.forward(x)
g = acq_model.backward(distance.gradient2(acq_data, est_data), x)
```
where the `backward(y,x)` is really the multiplication of the gradient at `x` with `y`. Obviously, in the affine case, `acq_model.backward(y,x)` doesn't depend on `x` (as it's just `A^t y`).

KrisThielemans:
I (somewhat belatedly) realise that we can resolve the affine question by shifting the constant into the "distance" function:
```
f=Ax
distance(acq_data, f) = KL(acq_data, f+ c)
```
This means you can stick to using a linear model. A similar trick will apply in the transmission problem, where a general model would be
```
f=Ax
distance(acq_data, f) = noise_model(acq_data, b exp(-f)+ c)
```
That's of course what @mehrhardt is doing in the algorithm derivation. I think this is quite neat, but on the other hand, I am convinced that physicists/engineers will expect an `AcquisitionModel` to actually model what the (noise-free) measured data is.

KrisThielemans:
So, going back to @mehrhardt 's suggestion
> I see two ways to resolve this conflict.
>
> First, the PET model could be changed to be linear and all of these problems disappear. The PET model A could still have a method like simulate_data or forward_affine to perform the affine model and get_background to get the vector of background coefficients.
>
> Second, we could leave the PET operator in SIRF non-linear A(x) = Bx + c (should not have an adjoint method! only "backward" as this is mathematically not well-defined) and instead has methods like get_linear_model=B, get_offset=c and these are used in algorithms which explicitly use the linearity of the operator such as SPDHG and all of ccpi.

The first option seems non-intuitive to me, and is going to fall-over once we get to kinetics etc. I strongly suggest to keep `PETAcquisitionModel` correspond to `forward(x)=Bx+c` and `backward(y)=B^t y` (with an ignored second argument I guess). Question is then how to accomodate SPDHG/CIL. `get_linear_model` could work, but in effect would just return a `PETAcquisitionModel` where the offset is 0. It seems a bit overkill (for SIRF) to create an extra class that. Given that in SIRF, the terminology for the offset is `additive_term`, i think I'd therefore would do something like
- add `get_linear_model` to `AcquisitionModel`, which returns a copy, but with additive term 0
- add `get_additive_term` to `AcquisitionModel`
- add `is_affine()` and `is_linear()` methods
- add `adjoint` which throws if not linear, but calls `backward` otherwise.


 linear_model = acq_m

mehrhardt:
@KrisThielemans Yes, this sounds all pretty good to me. Then one has a "mathy" version which is sometimes needed but can always fall back to the "forward" "backward" notation.

KrisThielemans:
If we do this, I think ideally most of these methods are in the C++ code, although we could/should implement them in Python only first to see if it works...

Not 100% sure what you'd want to do in CIL. `Operator->AffineOperator->LinearOperator`? And what about the adjoint and gradient stuff? @jakobsj, any ideas? What does ODL do?

jakobsj:
Re. CIL, having an AffineOperator as subclass of Operator could make sense
mathematically and then LinearOperator a subclass of that.

I'm not sure what you are asking in terms of adjoint and gradient stuff. In
general we try to keep things "where they belong". So adjoint would only be
available in the LinearOperator, gradient in functionals that are
smooth/differentiable, etc.

I also still don't understand the motivation to have a ConstantOperator for
the offset as discussed earlier. What do you have in mind there? Why an
operator?

src/xSTIR/pSTIR/STIR.py
paskino: They are used in `get_linear_acquisition_model` and `get_background_term`

src/xSTIR/pSTIR/STIR.py
paskino: `ad` seems to be `AcquisitionData`, and it is needed by the `forward` method

evgueni-ovtchinnikov: if `ad` is present, the forward projection is put there, otherwise it is returned (so it is like numpy's `out`)

paskino: can we name it `out`, then? It'll make it much clearer

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1015,6 +1028,66 @@ def set_background_term(self, bt):
...
+        '''Returns the sum of the additive and background terms of the AcquisitionModel
+           
+           PET acquisition model that relates an image x to the
+           acquisition data y as
+           (F)    y = S (G x + [a]) + [b]
+           where:
+           G is the geometric (ray tracing) projector from the image voxels
+           to the scanner's pairs of detectors (bins);
+           a and b are otional additive and background terms representing
+           the effects of noise and scattering;
```
KrisThielemans:
```suggestion
           the effects of accidental coincidences and scattering;
```

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1049,7 +1124,56 @@ def backward(self, ad, subset_num = 0, num_subsets = 1):
             (self.handle, ad.handle, subset_num, num_subsets)
         check_status(image.handle)
         return image
+    def get_linear_acquisition_model(self):
+        '''
+        Returns the linear part of the AcquisitionModel
```
KrisThielemans:
```suggestion
        Returns a new AcquisitionModel corresponding to the linear part of the current one.
```

src/xSTIR/pSTIR/STIR.py
KrisThielemans: yes please. However, this will need changing `forward` as well. Possibly has consequences for the examples etc. and would be a backwards compatibility break, so please create a separate issue for it, with a separate PR, such that it has some visibility.

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1049,7 +1124,56 @@ def backward(self, ad, subset_num = 0, num_subsets = 1):
...
+           Added for CCPi CIL compatibility
+           https://github.com/CCPPETMR/SIRF/pull/237#issuecomment-439894266
+        '''
+        if self.is_linear():
+            return self.backward(ad, subset_num = subset_num, 
+                             num_subsets = num_subsets)
+        else:
+            raise error('AcquisitionModel is not linear')
+    def is_affine(self):
+        '''Returns whether the AcquisitionModel is affine'''
```
KrisThielemans: see suggested change for MR.
possibly we should add in both that it is currently always True

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1049,7 +1124,56 @@ def backward(self, ad, subset_num = 0, num_subsets = 1):
...
+        return True
+    def is_linear(self):
+        '''Returns whether the constant term is zero'''
```
KrisThielemans: see suggested change in MR

KrisThielemans:
@paskino, please add something to the UsersGuide as well ( I guess a special section on CIL compatibliity for Python)

KrisThielemans:
Codacy found a few small issues as well.

KrisThielemans:
At present, `direct` throws an error if the model is not linear. I don't think this is necessary. `direct` can just be an alias for `forward`. Quoting @mehrhardt 
> if we have an inverse problem, defined by some data y \in Y and a mapping A : X -> Y, then computing A(x) is called the direct or forward problem

examples/Python/PET/acquisition_data.py
```diff
@@ -130,6 +130,10 @@ def main():
     print('norm of image*10: %f' % image.norm())
     diff = image.clone() - image
     print('norm of image.clone() - image: %f' % diff.norm())
+    
+    # test clone vs copy
```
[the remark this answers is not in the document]
paskino: It's in the docstring of the method. Do I have to put it here as well?

paskino:
That's right. It is the `adjoint` which doesn't exist if the `AcquisitionModel` isn't linear.

paskino:
As I tried to document [here](https://github.com/CCPPETMR/SIRF/blob/add_to_sirf_classes/doc/UserGuide.md#datacontainer) the `DataContainer` seems to be missing a number of methods. I thought I put them in but they may have disappeared when the `sirf.DataContainer` has been introduced.

I will implement them with calls to `as_array`, followed by the required operation in numpy and `fill`. It's not too efficient, but it'll work.

src/xSTIR/pSTIR/STIR.py
paskino: I used `out` in the `direct` method not in the `forward` so there won't be backward incompatibility.

KrisThielemans:
> As I tried to document here the DataContainer is missing a number of methods.

> I will implement them in sirf.DataContainer with calls to as_array, followed by the required operation in numpy and fill. It's not too efficient, but it'll work.

ok.please create an issue for it though. some of these would be easy to do in C++ I guess, but we can do that later. But we'll forget without issue :-)

examples/Python/PET/acquisition_data.py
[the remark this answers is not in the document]
KrisThielemans: you're assuming that people read the docstring... Yes, I'd put it here. it's an example. it'd confuse me if I see this code and I'd wonder "what's the difference between clone and copy now?"

paskino:
I believe that those missing methods should be implemented in this PR. First in Python and then in C++, for which I'll open an issue.

paskino:
Codacy fails mostly because of the grappa_and_cil.py example which is not finished. I'll move it to #238 when this PR is ready.

I need to implement the missing methods for `DataContainer` before merging this. I also discussed some changes to the CIL that should be done.

KrisThielemans:
> Codacy fails mostly because of the grappa_and_cil.py example which is not finished. I'll move it to #238 when this PR is ready.

Let's not start moving files between PRs. My head is hurting already with so many PRs (all good things have a bad side...). you'd have to `git rm` it here, and `git add` it there. Nobody will understand (as we're not squashing)

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -937,6 +948,32 @@ def backward(self, ad):
             (self.handle, ad.handle)
         check_status(image.handle)
         return image
+    def direct(self, image, out = None):
+        '''Projects an image into the (simulated) acquisition space, 
+           if the AcquisitionModel is linear.
```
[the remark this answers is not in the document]
paskino: What? "if the AcquisitionModel is linear"? 
That's not wrong, is it?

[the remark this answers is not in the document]
KrisThielemans: direct = forward, so will also do it if it's not linear. (best to say it's an alias) like for copy etc

src/xSTIR/pSTIR/STIR.py
```diff
@@ -964,7 +964,7 @@ def __del__(self):
         if self.handle is not None:
             pyiutil.deleteDataHandle(self.handle)
 
-class AcquisitionModel:
+class AcquisitionModel(object):
```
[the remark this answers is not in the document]
paskino: That's the [new style of class definition](https://wiki.python.org/moin/NewClassVsClassicClass) on Python (new since 2.2)

examples/Python/PET/acquisition_data.py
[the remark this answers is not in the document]
paskino: added doc that copy is an alias of clone

src/xSTIR/pSTIR/STIR.py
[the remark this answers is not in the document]
paskino: I guess I will remove it because the code is full of old style classes definition, but it's worth thinking to update.

KrisThielemans:
> I guess I will remove it because the code is full of old style classes definition, but it's worth thinking to update.

please create an issue for it then (for SIRF general) to use the new style

examples/Python/PET/interactive/spdhg_reconstruction.py
```diff
@@ -0,0 +1,310 @@
...
+    asm_attn.unnormalise(attn_factors)
+    print('done')
+    # use these in the final attenuation model
+    asm_attn = pet.AcquisitionSensitivityModel(attn_factors)
+    # chain attenuation and normalisation
+    asm = pet.AcquisitionSensitivityModel(asm_norm, asm_attn)
+    # update the acquisition model etc
+    am.set_acquisition_sensitivity(asm)
+
+    am.set_background_term(randoms)
```
KrisThielemans: I think this is incorrect for SPDHG. you will add it to the KullbackLeibler later.

Or, you add it here, and then use the `get_linear_model` below (which would be nice and clear in my opinion).

src/xSTIR/pSTIR/tests/tests_five.py
```diff
@@ -0,0 +1,102 @@
...
+
+{author}
+
+{licence}
+"""
+from sirf.Utilities import petmr_data_path, existing_filepath, \
+    pTest, RE_PYEXT , runner
+from sirf.STIR import MessageRedirector, AcquisitionData
+__version__ = "0.2.2"
+__author__ = "Casper da Costa-Luis, Edoardo Pasca"
```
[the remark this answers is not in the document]
paskino: This file is largely a plagiarism of another test in the same directory. For this reason I kept Casper's name. I just resolved to change the actual tests being run, not the way the script is run. I assumed these tests would be run in the ctest phase.

examples/Python/PET/interactive/spdhg_reconstruction.py
paskino: @KrisThielemans I'm not the most entitled to comment on this (@mehrhardt  help), however the KullbackLeibler fidelity [gets passed](https://github.com/CCPPETMR/SIRF/blob/add_to_sirf_classes/examples/Python/PET/interactive/spdhg_reconstruction.py#L236) the fully configured `AcquisitionMethod`.  

Then the `direct` method in the `AcquisitionModel`  [calls](https://github.com/CCPPETMR/SIRF/blob/add_to_sirf_classes/src/xSTIR/pSTIR/STIR.py#L1134) `self.get_linear_acquisition_model().forward()`
 
I had originally added this example to #238 so that I'd avoid this discussion on this particular PR, which deals with the additions of methods to SIRF classes in order to be able to use the CIL optimisation framework. The reason this example has appeared back here is the abstract @dkazanc is preparing. I'm happy to strip this example out of the PR so that we can merge the PR asap and then correct it.

KrisThielemans: well, let's not do crazy merges before the results are ok...

in your link, `AcquisitionModel,direct` just calls `forward`. I know we originally said that `direct` would do this, but we changed our mind...

For the future, I don't think it makes sense to let the user add the background to KL, and "hope" that `spdgh` does the appropriate thing. But I wouldn't worry about it now.

I still don't know if the current code is doing the wrong thing of course...

examples/Python/PET/interactive/spdhg_reconstruction.py
```diff
@@ -0,0 +1,310 @@
...
+f = [KullbackLeibler(op.sirf2sub(noisy_data), op.sirf2sub(background))
+     for op in A]
+
+#%%
+recon_noreg = spdhg(f, g_noreg, A, A_norms=Ls)
+
+#%%
+
+# set the number of iterations
+niter = 5
```
mehrhardt: As I commented via email, 5 "iterations" are not enough. Perhaps it is best to change this to:

nepochs = 5
nsubsets = 14 (see line 236)
niter = nsubsets * nepochs

examples/Python/PET/interactive/spdhgutils.py
```diff
@@ -0,0 +1,203 @@
...
+        self.ind = numpy.nonzero(y.flatten())[0]
+
+    def forward_sirf(self, x):
+        return self.__op__.forward(x, subset_num=self.__subset_num__,
+                                   num_subsets=self.__num_subsets__)
+
+    def __call__(self, x):
+        y = self.__op__.forward(x, subset_num=self.__subset_num__,
+                                num_subsets=self.__num_subsets__)
+        return self.sirf2sub(y)
```
mehrhardt: Here is a mistake. This class assumes that "forward", "direct" and "__call__" are all the same. I think we agreed that this is not the case. Thus, 

```
y = self.__op__.forward(x, subset_num=self.__subset_num__,
                                num_subsets=self.__num_subsets__)
```

should be changed to 

```
y = self.__op__.direct(x, subset_num=self.__subset_num__,
                                num_subsets=self.__num_subsets__)
```

mehrhardt: Or maybe we can just remove the "__call__" altogether as both CCPi and SIRF do not support the syntax "A(x)" anyways.

KrisThielemans: > This class assumes that "forward", "direct" and "call" are all the same. I think we agreed that this is not the case.

I believe we agreed that `forward` and `direct` were the same (after @jakobsj 's terminology post). I have directed @paskino accordingly. We introduced the `get_linear_model` for this.

I have no strong feelings about `__call__`. If we have it, it would be `forward` (but then what do we do for `adjoint` etc). Maybe it's an ODL thing?

mehrhardt: > I believe we agreed that forward and direct were the same (after @jakobsj 's terminology post). I have directed @paskino accordingly. We introduced the get_linear_model for this.

Ah, ok. I kind of lost track of our discussion.

> I have no strong feelings about __call__. If we have it, it would be forward (but then what do we do for adjoint etc). Maybe it's an ODL thing?

In ODL, we write `A(x)` and `A.adjoint(y)` but it might actually be better not to use it in order to avoid misunderstandings.

PR #238, opened 2018-11-15
SPDHG algorithm from CIL

paskino:
Adds an example on using the SPDHG algorithm with OS from [CIL](https://github.com/vais-ral/CCPi-Framework)

Requires https://github.com/CCPPETMR/Hackathon-SIRF-SuperBuild/pull/1 which I am going to port from the hackathon repo to this.


mehrhardt:
The PR looks good to me. 

A general question here is where in STIR/SIRF should all this code go? I.e. the algorithm `spdhg`, definition of KL divergence etc. Some of the files are in what was intended as a temporary folder "pCIL".

Also there are a few "hacks" to quickly code things up in examples/Python/PET/interactive/spdhgutils.py like the `SubsetOperator`. These are not `spdhg` specific and might be useful for other subset based algorithms. One of the purposes of the class `OperatorInd` is to "crop" the output of the STIR forward operator to a specific subset in order to minimize memory and cpu footprint of `spdhg`. At the moment STIR would output an array of the same size as the usual data but only with non-zeros for the actual subset. `spdhg` then takes this output and runs a few other operations on variables of the same size (square root, ...).

paskino:
The actual algorithm is in CIL and the [pCIL](https://github.com/CCPPETMR/SIRF/tree/spdhg_from_cil/src/pCIL) package should be deleted.

Also #237 should be merged first.

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -1202,6 +1212,10 @@ def backward(self, ad):
             (self.handle, ad.handle)
         check_status(image.handle)
         return image
+    def direct(self, image):
```
mehrhardt: Why does `CoilImageData` have a method `direct`? This is unintuitive to me.

examples/Python/PET/interactive/spdhgutils.py
```diff
@@ -0,0 +1,171 @@
...
+       print("calling FGP")
+       out = super(FGP_TV_SIRF, self).prox(x, sigma)
+       print("done")
+       y = x.copy()
+       y.fill(out.as_array())
+       return y
+   
+
+def PowerMethodNonsquare(op, numiters, x0=None):
+    # Initialise random
```
paskino: better remove all the comments

src/xGadgetron/pGadgetron/Gadgetron.py
paskino: what about `adjoint`? Should it be there?

mehrhardt: Well, I don't know. To me this is a data object and not an operator. So it should neither have `direct`, `forward` nor `adjoint`. But it could be that I did not see any specific CCP-PETMR point of view on this. @KrisThielemans ?

evgueni-ovtchinnikov: agree with Matthias - this class should have neither direct nor adjoint

examples/Python/PET/acquisition_data.py
```diff
@@ -130,6 +130,10 @@ def main():
     print('norm of image*10: %f' % image.norm())
     diff = image.clone() - image
     print('norm of image.clone() - image: %f' % diff.norm())
+    
+    # test clone vs copy
```
KrisThielemans: this is a test. should it be in an example? I'd rather put it in a test

examples/Python/PET/interactive/spdhg_reconstruction.py
```diff
@@ -0,0 +1,188 @@
...
+noisy_array = numpy.random.poisson(array.astype('float64'))
+max_counts = noisy_array.max()
+print(' Maximum counts in the data: {}'.format(noisy_array.max()))
+noisy_data = data.clone()
+noisy_data.fill(noisy_array);
+
+g_noreg = ZeroFun()
+
+
+g_reg = FGP_TV_SIRF(lambdaReg=.3,
```
KrisThielemans: some doc please what this does

examples/Python/PET/interactive/spdhg_reconstruction.py
```diff
@@ -0,0 +1,188 @@
...
+
+#%%
+recon_noreg = spdhg(f, g_noreg, A, A_norms=Ls)
+
+# %%
+for i in range(niter):
+    print(recon_noreg.iter)
+    recon_noreg.update()
+
+#%% does currently not work!
```
KrisThielemans: ah well...

examples/Python/PET/interactive/spdhgutils.py
```diff
@@ -0,0 +1,171 @@
...
+#
+# This demo is a 'script', i.e. intended to be run step by step in a
+# Python IDE such as spyder. It is organised in 'cells'. spyder displays these
+# cells nicely and allows you to run each cell on its own.
+#
+# First version: 8th of September 2018
+# Author: Matthias J Ehrhardt, Edoardo Pasca
+#
+## CCP PETMR Synergistic Image Reconstruction Framework (SIRF)
+## Copyright 2015 - 2018 Rutherford Appleton Laboratory STFC
```
KrisThielemans: this will need some change!

PR #244, opened 2018-11-27
Fixes #243 

### LINKED ISSUE #243: Sptr functinality for sirf::AcquisitionsVector
Currently the `sirf::AcquisitionsVector` carries a data member of type
`std::vector<gadgetron::shared_ptr<ISMRMRD::Acquisition> > acqs_;`

However, there is no function directly appending a `gadgetron::shared_ptr<ISMRMRD::Acquisition>` with call-by-value argument to have a memory-efficient adding of shared pointers. Currently adding an `ISMRMRD::Acquisition` involves `gadgetron::make_shared` which generates a duplication of memory (c.f. #204 ) .
 
Functions `get_sptr_acquisition(gadgetron::shared_ptr<ISMRMRD::Acquisition>)` and `append_sptr_acquisition(gadgetron::shared_ptr<ISMRMRD::Acquisition>)` are needed to allow constructions such as the following without memory duplication:
```
sirf::AcquisitionsVector all_acquis = some filled AcquisitionsVetor;

size_t num_acq = all_acquis.items();
size_t num_iterations = 1000;

std::vector< sirf::AcquisitionsVector > acq_vec_vector;
for(size_t i=0; i<num_iterations; i++)
	{
	std::cout << " NUM IT "<<i<<std::endl;
		
        sirf::AcquisitionsVector temp_vec;
	temp_vec.copy_acquisitions_info( all_acquis ) ;

	for(size_t i_acq=0; i_acq<num_acq; i_acq++)
	{	
		auto sptr_acq = all_acquis.get_sptr_acquisition(i_acq);
		temp_vec.append_sptr_acquisition(sptr_acq);
	}
	acq_vec_vector.push_back( temp_vec );
}
```

Also the index function `MRAcquisitionData::get_acquisition()` calling `this->index()` is not save as
```
int MRAcquisitionData::index(int i)
		{
			if (index_ && i >= 0 && i < (int)number())
				return index_[i];
			else
				return i;
		}
```
does not protect from i being out of range.

johannesmayer:
Functionality added to access shared pointers stored inside the `sirf::AcquisitionsVector` is added.

Appending them using the call by value function `append_sptr_acquisition(gadgetron::shared_ptr<ISMRMRD::Acquisition> sptr_acqu)` increases the internal counts of the shared pointers. This way, shared pointers to acquisitions can be extracted from one `AcquisitionsVector`, and appended to another one, without the data stored inside the objects pointed to being duplicated, as happens when the `get_acquisition(int, ISMRMRD::Acquisition& aq)` followed by `append_acquisition( const ISMRMRD::Acquisition& aq)` is used.

Also the `MRAcquisitionData::index()` function now is save to call for indices out of range.

### DIFF
--- a/src/xGadgetron/cGadgetron/gadgetron_data_containers.h
+++ b/src/xGadgetron/cGadgetron/gadgetron_data_containers.h
@@ -139,8 +139,8 @@ namespace sirf {
 		virtual void set_acquisition(unsigned int num, ISMRMRD::Acquisition& acq) = 0;
 		virtual void append_acquisition(ISMRMRD::Acquisition& acq) = 0;
 
-		virtual void append_sptr_acquisition( gadgetron::shared_ptr<ISMRMRD::Acquisition> sptr_acqu)=0;
-		virtual gadgetron::shared_ptr<ISMRMRD::Acquisition> get_sptr_acquisition(unsigned int num)=0;
+		virtual void append_acquisition_sptr( gadgetron::shared_ptr<ISMRMRD::Acquisition> sptr_acqu)=0;
+		virtual gadgetron::shared_ptr<ISMRMRD::Acquisition> get_acquisition_sptr(unsigned int num)=0;
 
 
 		virtual void copy_acquisitions_info(const MRAcquisitionData& ac) = 0;
@@ -261,8 +261,8 @@ namespace sirf {
 		}
 		virtual void append_acquisition(ISMRMRD::Acquisition& acq);
 
-		virtual void append_sptr_acquisition( gadgetron::shared_ptr<ISMRMRD::Acquisition> sptr_acqu){throw LocalisedException("Function undefined for sirf::AcuisitionsFile.", __FILE__, __LINE__);};
-		virtual gadgetron::shared_ptr<ISMRMRD::Acquisition> get_sptr_acquisition(unsigned int num){throw LocalisedException("Function undefined for sirf::AcuisitionsFile.", __FILE__, __LINE__);};
+		virtual void append_acquisition_sptr( gadgetron::shared_ptr<ISMRMRD::Acquisition> sptr_acqu){throw LocalisedException("Function undefined for sirf::AcuisitionsFile.", __FILE__, __LINE__);};
+		virtual gadgetron::shared_ptr<ISMRMRD::Acquisition> get_acquisition_sptr(unsigned int num){throw LocalisedException("Function undefined for sirf::AcuisitionsFile.", __FILE__, __LINE__);};
 
 		virtual void copy_acquisitions_info(const MRAcquisitionData& ac);
 		virtual MRAcquisitionData*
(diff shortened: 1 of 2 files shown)


src/xGadgetron/cGadgetron/gadgetron_data_containers.h
```diff
@@ -248,6 +260,10 @@ namespace sirf {
 			std::cerr << "AcquisitionsFile::set_acquisition not implemented yet, sorry\n";
 		}
 		virtual void append_acquisition(ISMRMRD::Acquisition& acq);
+
+		virtual void append_sptr_acquisition( gadgetron::shared_ptr<ISMRMRD::Acquisition> sptr_acqu){throw LocalisedException("Function undefined for sirf::AcuisitionsFile.", __FILE__, __LINE__);};
```
KrisThielemans: rename to `append_acquisition_sptr`. or do we have other functions with similar names?
also split into 2 lines

src/xGadgetron/cGadgetron/gadgetron_data_containers.h
```diff
@@ -248,6 +260,10 @@ namespace sirf {
...
+		virtual gadgetron::shared_ptr<ISMRMRD::Acquisition> get_sptr_acquisition(unsigned int num){throw LocalisedException("Function undefined for sirf::AcuisitionsFile.", __FILE__, __LINE__);};
```
KrisThielemans: as above `get_acquisition_sptr`. should also be a `const` member.

But do we really need this? Providing access to internal `shared_ptr` can be quite dangerous.

johannesmayer: I suggested this change becaues in the MR side of the simulation framework, arbitrarily many dynamics are enabled, e.g. one could add two motion and contrast variations for four different tissue types. Then I keep track of which readout line is acquired in which dynamic state and simulate the data which the `ISMRMRD::Acquisition` carries in the respective dynamic state.

So all `ISMRMRD::Acquisition` readout lines from some template file are read into memory, but I need to keep track of them with a pointer, otherwise the memory use explodes. 

We want to use files of size ~5GB, flexibly duplicating them during the simulation blows up the memory (of my laptop).

johannesmayer: Is there some other way to have objects just exist once and manipulate them in a safer way?

KrisThielemans: I see you have an example at #243. I'll add a question there soon.

src/xGadgetron/cGadgetron/gadgetron_data_containers.h
evgueni-ovtchinnikov: agree with Kris - all our methods returning shared pointers have sptr as suffix, and shared pointer properties as prefix

src/xGadgetron/cGadgetron/gadgetron_data_containers.h
johannesmayer: Making the `get_acquisition_sptr()` a `const` function is non-trivial unfortunately.
This requires `items()`and `number()`to be `const` as well and these methods come from the interface `aDataContainer` so this is a larger-scale operation I don't want to get into in this PR, haha, also since there is the new data hierarchy coming from @evgueni-ovtchinnikov I think which might involve changes there.

KrisThielemans:
@johannesmayer I believe you said we don't need this. Can you then close this PR?

PR #245, opened 2018-11-30
Remove ismrmrd dependency in sirf/common

rijobro:
Remove ismrmrd dependency in sirf/common

### DIFF
--- a/src/common/include/sirf/common/num_ref.h
+++ b/src/common/include/sirf/common/num_ref.h
@@ -3,7 +3,19 @@
 #include <complex>
 #include <typeinfo>
 
-#include <ismrmrd/ismrmrd.h>
+// Data Types from ismrmrd/ismrmrd.h (saves having to include it in this file, which should be independent of it)
+enum ISMRMRD_DataTypes {
+    ISMRMRD_USHORT   = 1, /**< corresponds to uint16_t */
+    ISMRMRD_SHORT    = 2, /**< corresponds to int16_t */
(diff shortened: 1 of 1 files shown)


KrisThielemans:
thanks. definitely a good idea to be independent of ISMRMRD.

I agree with @bathomas, we should drop ISMRMRD from the name. There doesn't seem to be any point in adding `SIRF` as it's already in the `sirf` namespace. I vote for `enum class NumberType`  (`DataType` seems to general to me). It then probably needs to be in its own header as I guess @rijobro will want to use it for SIRFReg (or has it already...)

rijobro:
Apparently I can't use a strongly typed `enum`, because the following switch is used:
```
switch (type_) {
    case NumberType::CXDOUBLE:
...
```
where `type_` is an `int`. This can't be done as strongly typed `enum`s can't be converted to `int`s. Are you happy with me using `enum NumberType` instead of `enum class NumberType`?

rijobro:
Done.

KrisThielemans:
of course, a merge is also fine (depending on what @rijobro wants)

rijobro:
Not fussed either way. Whatever the git gurus decided!

KrisThielemans:
if not merged elsewhere, squash merge. it's a small change. one commit would do the trick

evgueni-ovtchinnikov:
all travis builds failed with the message

fatal: Authentication failed for 'https://@github.com/ccp-petmr-codebot/github-stats/'

PR #250, opened 2018-12-05
include directory sirf/common in base CMakeLists

rijobro:
sirf/common is used throughout the code. makes sense to include it higher up, to avoid repetition.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -120,6 +120,11 @@ endif()
 
 ENABLE_TESTING()
 
+INCLUDE_DIRECTORIES(src/common/include)
+INCLUDE_DIRECTORIES(src/iUtilities/include)
+INCLUDE_DIRECTORIES(src/xSTIR/cSTIR/include)
+INCLUDE_DIRECTORIES(src/xGadgetron/cGadgetron/include)
+
 ADD_SUBDIRECTORY(src/iUtilities)
 ADD_SUBDIRECTORY(src/xSTIR)
 ADD_SUBDIRECTORY(src/xGadgetron)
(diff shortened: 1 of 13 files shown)


rijobro:
Any thoughts on this? There will at some point in the future be a folder for synergistic algorithms. They would need to access headers for Gadgetron, STIR, common, etc.

It seems to make sense to me to put `include_directories` for all sections, ie:
```
INCLUDE_DIRECTORIES(src/common/include)
INCLUDE_DIRECTORIES(src/iUtilities/include)
INCLUDE_DIRECTORIES(src/xSTIR/cSTIR/include)
INCLUDE_DIRECTORIES(src/xGadgetron/cGadgetron/include)
INCLUDE_DIRECTORIES(src/Registration/cReg/include)
```
In this way, all subdirectories of SIRF will be able to access all SIRF header files in different subdirectories elsewhere in the code.

KrisThielemans:
Instead of doing this, we really should move to [target_include_directories](https://cmake.org/cmake/help/latest/command/target_include_directories.html) for the libraries. That way, a user who says she needs the library, automatically gets the include file location. That'll also work once we install the include files. For instance, I think the following would be needed.
```
target_include_directories(common PUBLIC
  $<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include/>
  $<INSTALL_INTERFACE:include/
)
```
I'd drop this PR therefore, unless that causes problems for anyone who has merged it already...

rijobro:
Ok. Similarly, synergistic code will need to access `${STIR_INCLUDE_DIRS}`, which is created in `src/xSTIR/CMakeLists.txt`. Presumably, this can also be included with `target_include_directories`?

KrisThielemans:
can't remember if STIR uses this already, but it should...

rijobro:
Ok.

Issue #204, opened 2018-08-20
Memory Leak in sirf::AcquisitionsVector


johannesmayer:
The class `sirf::AcquisitionsVector` has a memory leak.
Loop constructions such as
```
AcquisitionsVector all_acquis = some acquisitions vector filled with data and acquisitions info.

AcquisitionsVector temp_dummy_vector;
temp_dummy_vector.copy_acquisitions_info(all_acquis);
ISMRMRD::Acquisition acq;
for(size_t i_acq=0; i_acq<all_acquis.number(); i_acq++)
{
    all_acquis.get_acquisition(i_acq, acq);
    temp_dummy_vector.append_acquisition(acq);
}
```
whereas replacing 
```
ISMRMRD::Acquisition acq;
for(size_t i_acq=0; i_acq<all_acquis.number(); i_acq++)
{
    all_acquis.get_acquisition(i_acq, acq);
    temp_dummy_vector.append_acquisition(acq);
}
```
by
```
for(size_t i_acq=0; i_acq<all_acquis.number(); i_acq++)
{
    ISMRMRD::Acquisition acq;
    all_acquis.get_acquisition(i_acq, acq);
    temp_dummy_vector.append_acquisition(acq);
}
```
does not generate a leak.

evgueni-ovtchinnikov:
just checked - the same happens if we use `std::vector<ISMRMRD_Acquisition>`

so it must be `std::vector` idiosyncrasy, not `shared_ptr`

evgueni-ovtchinnikov:
or else something wrong with `ISMRMRD_Acquisition` assignment

KrisThielemans:
has this been resolved?

evgueni-ovtchinnikov:
@johannesmayer: do you still have this leak?

johannesmayer:
I did not investigate any further into this matter. So I guess the bug is still present.

evgueni-ovtchinnikov:
i can no longer reproduce this leak - perhaps, ismrmrd was updated recently?

johannesmayer:
I let the my testjob run again (code below), and this leak is still present.

The following snippet should need a huge amount of memory, whereas setting `construct_inside_loop=true` will lead to freeing of memory after each iteration of the `for` loop.
(`ISMRMRD_H5_TEST_PATH` needs to point to a valid ISMRMRD file location of course).
[...]

PR #251, opened 2018-12-10
Add time_order MRAcquisitions

johannesmayer:
Added a function called `MRAcquisitionData::time_order()`.

The code is analogous to the already existing `MRAcquisitionData::order()` function.

In the `MRAcquisitionData::order` function the constructor of the `ISMRMRD::Acquisition` object was moved into the loop, to avoid the memory leak (c.f. #204 ).


johannesmayer:
I need `MRAcquisitionData::time_order`  in my simulation framework. I want to group together `ISMRMRD::Acquisition` s adjacent in time, so I thought I'd add this function.
It has nothing to do with reconstructing the data. 

Since `order()` has a hierarchy of sorting starting with respect to repetition, and then slices and phase encoding, I did not want to interfere with it that's why I wrote a new function just dedicated to time ordering.

KrisThielemans:
@evgueni-ovtchinnikov if order is crucial, should we set a private member to know if the data was ordered and how? In fact, I see we have this in Python/Matlab. I've openened https://github.com/CCPPETMR/SIRF/issues/270 for this.

Anyone can come up with a way to merge `order` and `time_order` into 1 function? (Supplying some arguments). could be a bit messy.

KrisThielemans:
@johannesmayer can we call this `sort_by_time` (and Evgueni latter on rename `order` to `sort` as in Python/MATLAB). how does that sound?

@johannesmayer please merge master onto here to resolve conflicts

PR #253, opened 2018-12-11
Mark DataContainer methods const where possible

rijobro:
Methods should be marked as `const` where possible (so they can be used by `const` objects)

PR #254, opened 2018-12-11
New data container sptr

rijobro:
In SIRFReg, we store shared pointers to `ImageData`. The output of the resampling/registration should be the same as the reference image, which means we need a virtual method that will create a shared pointer of any class that derives from `ImageData`.

Depends on #253.


src/common/include/sirf/common/DataContainer.h
```diff
@@ -41,11 +41,12 @@ namespace sirf {
 	class DataContainer {
 	public:
 		virtual ~DataContainer() {}
-		virtual DataContainer* new_data_container() = 0;
-		virtual ObjectHandle<DataContainer>* new_data_container_handle() = 0;
-		virtual unsigned int items() = 0;
-		virtual float norm() = 0;
-		virtual void dot(const DataContainer& dc, void* ptr) = 0;
+        virtual DataContainer* new_data_container() const = 0;
+        virtual std::shared_ptr<DataContainer> new_data_container_sptr() const = 0;
```
evgueni-ovtchinnikov: strongly against - i spent some time to make `DataContainer` independent of `shared_ptr` used (`std` or `boost`)

can you not instead create `shared_ptr` in your code from the pointer returned by `new_data_container()`?

rijobro: I thought about that, but I don't think I can do that (unless I try to dynamic_cast to all derived classes.

What is the rationale for getting rid of shared pointers? I know we wanted to get rid of `boost`, but `std` is OK, isn't it?

evgueni-ovtchinnikov: boost::shared_ptr is still an option in STIR

could you not use this function instead?

    std::shared_ptr<DataContainer> new_data_container_sptr(const DataContainer& dc)
    {
        return std::shared_ptr<DataContainer>(dc.new_data_container());
    }

KrisThielemans:
we agreed to no longer support `boost::shared_ptr` for STIR. We need C++-11 anyway, so we just say that STIR needs to be compiled with C++-11.

But the trick might work ok?

src/common/include/sirf/common/DataContainer.h
rijobro: And just put it in `DataContainer.h` instead of all the derived classes? Good point.

evgueni-ovtchinnikov: to cover `both shared_ptr`, i would do:
'''
void get_new_data_container_sptr(const DataContainer& dc, std::shared_ptr<DataContainer>& sptr)
{
    sptr = std::shared_ptr<DataContainer>(dc.new_data_container());
}
'''

rijobro: Fine by me. But I suppose we should only implement the first of those two for the time being? Since we're trying to move away from boost?

KrisThielemans: no please. no `boost`. 

a `get` method has to return something. Otherwise it's have to be called `fill` or so (in which case the `output` has to come first).

rijobro: `fill_shared_ptr_with_new_data_container()`. It just rolls of the tongue!

evgueni-ovtchinnikov: get: i for one always prefer to use a verb for void function (do_something_with(x);) and noun for a one that returns something (x = a_thing();) (cf. set_up vs setup)

KrisThielemans: I'm sorry. I see no need for the `fill`. I'm happy with the first proposal 
```
std::shared_ptr<DataContainer> get_new_data_container_sptr() const
{
    return std::shared_ptr<DataContainer>(this->new_data_container());
}
```
probably use `make_shared`

KrisThielemans: good convention with the verb and noun, but sadly `get` is so standard that we have to stick to it (I'm hoping we do in other places)

rijobro: Ok, reverted to `return std::shared_ptr<DataContainer>(new_data_container());`. Can't use `make_shared` as that calls the object's contructor.

src/common/include/sirf/common/DataContainer.h
```diff
@@ -41,11 +41,15 @@ namespace sirf {
 	class DataContainer {
 	public:
 		virtual ~DataContainer() {}
-		virtual DataContainer* new_data_container() = 0;
-		virtual ObjectHandle<DataContainer>* new_data_container_handle() = 0;
-		virtual unsigned int items() = 0;
-		virtual float norm() = 0;
-		virtual void dot(const DataContainer& dc, void* ptr) = 0;
+		virtual DataContainer* new_data_container() const = 0;
+		virtual std::shared_ptr<DataContainer> new_data_container_sptr() const
```
KrisThielemans:
```suggestion
		std::shared_ptr<DataContainer> new_data_container_sptr() const
```

src/common/include/sirf/common/ImageData.h
```diff
@@ -39,7 +39,7 @@ namespace sirf {
 		virtual Iterator_const& begin() const = 0;
 		virtual Iterator& end() = 0;
 		virtual Iterator_const& end() const = 0;
-		void copy(Iterator_const& src, Iterator& dst, Iterator& end)
+		void copy(Iterator_const& src, Iterator& dst, Iterator& end) const
```
KrisThielemans:
```suggestion
		static void copy(Iterator_const& src, Iterator& dst, Iterator& end)
```

rijobro:
Closing as superceded by #277.

PR #256, opened 2018-12-15
[TRAVIS] Fixes

KrisThielemans:
- use gcc-6 on Linux for DEVEL_BUILD (and gcc-5 otherwise)
- upgrade to CMake 3.13
- always build Boost as requirement too high for Trusty

See https://github.com/CCPPETMR/SIRF-SuperBuild/pull/176 for original commits


KrisThielemans:
still 2 failures with the Gadgetron server on OSX. This doesn't have anything to do with this PR though as far as I can see.

PR #266, opened 2018-12-19
added gitignore entries for clion and qtcreator ides, plus some libra…

ALEXJAZZ008008:
…ry files which were missing. can comment idea file as prefered.


.gitignore
```diff
@@ -184,6 +184,81 @@ UpgradeLog*.htm
...
+object_script.*.Debug
+*_plugin_import.cpp
+/.qmake.cache
+/.qmake.stash
+*.pro.user
+*.pro.user.*
+*.qbs.user
+*.qbs.user.*
+*.moc
+moc_*.cpp
```
KrisThielemans: somewhat reluctant to add these. who knows if we ever call a file that way?

.gitignore
```diff
@@ -184,6 +184,81 @@ UpgradeLog*.htm
...
+moc_*.h
+qrc_*.cpp
+ui_*.h
+*.qmlc
+*.jsc
+Makefile*
+*build-*
+
+# Qt unit tests
+target_wrapper.*
```
KrisThielemans: same here

.gitignore
```diff
@@ -184,6 +184,81 @@ UpgradeLog*.htm
...
 FakesAssemblies/
 
+# Covers JetBrains IDEs: CLion
+# Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
+
+# Ignore entire idea
+# Comment as required
+*.idea
+
+# User-specific stuff
```
KrisThielemans: what about just saying: `.idea`?

bathomas: For CLion I tend to use:

```
.idea/*
cmake-build-debug/*
cmake-build-release/*
```

KrisThielemans:
@ALEXJAZZ008008 could you finish this one please?

ALEXJAZZ008008:
Sorry, I wasn't certain if we were still interested in merging the edits I had made to the .gitignore.
I've removed the first two blocks you've mentioned and edited the third block based on Ben's comment.
Regarding Ben's comment, why is it more advantageous to ignore the contents of the .idea directory rather than just ignore any directory with the prefix .idea?

bathomas:
I had a problem a while back that the IDE was still showing local changes to files within `.idea` even when `.idea` had been added to `.gitignore`. Ignoring the contents appeared to fix it.

Thinking about it, possibly it would be better to have `cmake-build-*/` ?

ALEXJAZZ008008:
I agree that cmake-build-*/* would be better, I've edited the lines.

.gitignore
```diff
@@ -184,6 +184,73 @@ UpgradeLog*.htm
...
+.idea/**/dataSources/
+.idea/**/dataSources.ids
+.idea/**/dataSources.local.xml
+.idea/**/sqlDataSources.xml
+.idea/**/dynamic.xml
+.idea/**/uiDesigner.xml
+.idea/**/dbnavigator.xml
+
+# CMake
+cmake-build-*/
```
KrisThielemans: repetition

.gitignore
```diff
@@ -184,6 +184,73 @@ UpgradeLog*.htm
...
 
+# Covers JetBrains IDEs: CLion
+# Reference: https://intellij-support.jetbrains.com/hc/en-us/articles/206544839
+
+# CLion
+.idea/*
+cmake-build-*/*
+
+# User-specific stuff
+.idea/**/workspace.xml
```
KrisThielemans: seems we don't need all of these anymore as far as I know

.gitignore
```diff
@@ -184,6 +184,73 @@ UpgradeLog*.htm
...
+*.iws
+
+# mpeltonen/sbt-idea plugin
+.idea_modules/
+
+# JIRA plugin
+atlassian-ide-plugin.xml
+
+# Editor-based Rest Client
+.idea/httpRequests
```
KrisThielemans: not necessary anymore

.gitignore
KrisThielemans: git doc says to use `.idea/` etc. Please check

Issue #262, opened 2018-12-18
GadgetronImageData::order() only correct for axial slices


rijobro:
Because of the following code in `gadgetron_data_containers.cpp`:
```
for (int i = 0; i < ni; i++) {
      ImageWrap& iw = image_wrap(i);
      ISMRMRD::ImageHeader& head = iw.head();
		t[0] = head.repetition;
		t[1] = head.position[2];
		t[2] = head.slice;
		vt.push_back(t);
	}
```

evgueni-ovtchinnikov:
Richard already fixed this

PR #269, opened 2018-12-19
GadgetronImageWrap::order()

rijobro:
This addresses #262. Sort so that contrast varies slowest, then repetition and then lastly sort by dot product between `position` and `slice_dir`.

### DIFF
--- a/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
+++ b/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
@@ -553,14 +553,15 @@ MRAcquisitionData::clone()
 void
 MRAcquisitionData::order()
 {
-	typedef std::array<int, 3> tuple;
+	typedef std::array<int, 4> quad;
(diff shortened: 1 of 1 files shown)


KrisThielemans:
Should [AcquisitionData::order](https://github.com/CCPPETMR/SIRF/blob/029f1f2e8890cc1cb22815507948b652f08cb1ad/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp#L563) include contrast as well?

DANAJK:
It wasn't clear to us if we should sort on all the possible parameters, e.g. contrast, or not. I assume that for parameters that are not sorted, the natural order will be the temporal order of acquisition and this seems like a reasonable way to leave it.

KrisThielemans:
makes sense to me

KrisThielemans:
> I'd suggest adding also the head.phase

agreed. let's then merge this

rijobro:
Already done.

PR #274, opened 2018-12-20
Pure virtual method write for ImageData

rijobro:
This needs to be here so that the registration/resampling can return any image type, and that image type can be saved to file.

### DIFF
--- a/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
+++ b/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
@@ -1008,7 +1008,7 @@ GadgetronImageData::read(std::string filename)
 }
 
 void
-GadgetronImageData::write(std::string filename, std::string groupname) const
+GadgetronImageData::write(const std::string &filename, const std::string &groupname) const
 {
 	//if (images_.size() < 1)
 	if (number() < 1)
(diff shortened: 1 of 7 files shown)


src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
```diff
@@ -1008,7 +1008,7 @@ GadgetronImageData::read(std::string filename)
 }
 
 void
-GadgetronImageData::write(std::string filename, std::string groupname)
+GadgetronImageData::write(std::string filename, std::string groupname) const
```
KrisThielemans:
```suggestion
GadgetronImageData::write(const std::string& filename, const std::string& groupname) const
```
while we're modifying it...

src/xGadgetron/cGadgetron/include/sirf/cGadgetron/gadgetron_data_containers.h
```diff
@@ -407,7 +407,8 @@ namespace sirf {
 		virtual void get_real_data(float* data) const;
 		virtual void set_real_data(const float* data);
 		virtual int read(std::string filename);
-		virtual void write(std::string filename, std::string groupname);
+		virtual void write(std::string filename, std::string groupname) const;
```
KrisThielemans:
```suggestion
		virtual void write(const std::string& filename, const std::string& groupname) const;
```

KrisThielemans:
will need to expose to Python/MATLAB. Please raise an issue to do this later.

PR #275, opened 2018-12-21
Use target_include_directories

rijobro:
Following from #250, if `target_include_directories` is used, then the header files will be available when a library is linked.

Some `.h` files are still at the same level as the `.cpp` files, whereas the majority are in `include/sirf/engine`. This means that for the time being, we need to include to directories per library.

### DIFF
--- a/src/xGadgetron/cGadgetron/CMakeLists.txt
+++ b/src/xGadgetron/cGadgetron/CMakeLists.txt
@@ -32,20 +32,17 @@ if (SIRF_INSTALL_DEPENDENCIES AND WIN32)
 	endforeach()
   endif()
 	
-include_directories(${PROJECT_SOURCE_DIR}/src/common/include)
-include_directories(${PROJECT_SOURCE_DIR}/src/iUtilities/include)
-include_directories(${PROJECT_SOURCE_DIR}/src/xGadgetron/cGadgetron/include)
-
 add_library(cgadgetron cgadgetron.cpp gadgetron_x.cpp gadgetron_data_containers.cpp gadgetron_client.cpp ismrmrd_fftw.cpp)
 
-set (cGadgetron_INCLUDE_DIR "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>$<INSTALL_INTERFACE:include>")
+set (cGadgetron_INCLUDE_DIR "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>$<INSTALL_INTERFACE:include>")
+target_include_directories(cgadgetron PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>$<INSTALL_INTERFACE:include>")
 # copy to parent scope
 #set (cGadgetron_INCLUDE_DIR "${cGadgetron_INCLUDE_DIR}" PARENT_SCOPE)
 
 target_include_directories(cgadgetron PUBLIC "${cGadgetron_INCLUDE_DIR}")
 target_include_directories(cgadgetron PRIVATE "${FFTW3_INCLUDE_DIR}")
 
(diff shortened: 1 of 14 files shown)


src/xGadgetron/cGadgetron/CMakeLists.txt
```diff
@@ -32,20 +32,17 @@ if (SIRF_INSTALL_DEPENDENCIES AND WIN32)
...
 	
-include_directories(${PROJECT_SOURCE_DIR}/src/common/include)
-include_directories(${PROJECT_SOURCE_DIR}/src/iUtilities/include)
-include_directories(${PROJECT_SOURCE_DIR}/src/xGadgetron/cGadgetron/include)
-
 add_library(cgadgetron cgadgetron.cpp gadgetron_x.cpp gadgetron_data_containers.cpp gadgetron_client.cpp ismrmrd_fftw.cpp)
 
-set (cGadgetron_INCLUDE_DIR "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>$<INSTALL_INTERFACE:include>")
+set (cGadgetron_INCLUDE_DIR "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>$<INSTALL_INTERFACE:include>")
+target_include_directories(cgadgetron PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>$<INSTALL_INTERFACE:include>")
```
KrisThielemans: repeated below

src/xSTIR/CMakeLists.txt
```diff
@@ -28,6 +28,9 @@ if (STIR_FOUND)
   ADD_SUBDIRECTORY(pSTIR)
   ADD_SUBDIRECTORY(mSTIR)
 
+  # Going to need this elsewhere
```
KrisThielemans: arguably, this means we should find STIR one level up (in src/CMakeLists.txt). This needs a tiny bit of reorg of the CMakeLists.txt. I guess raise an issue and do it later

rijobro: I think it's ok here, as it keeps the engines separated. The only code that should need it elsewhere is the synergistic stuff. 

And arguably, the synergistic CMake code should be more intellegent (ie. only build against engines that can be found).

src/xGadgetron/cGadgetron/CMakeLists.txt
rijobro: No repetition as far as I can see. We need both:
```
target_include_directories(cgadgetron PUBLIC ...>
```
and
```
target_include_directories(cgadgetron PUBLIC .../include...>
```
because some files (e.g., `cstir.h` are still at the same level as the `.cpp` files.

rijobro:
>2 comments here, but at least one (install csirf) in #193 that should have been here. please merge this asap

I don't understand. You want me to merge in #193? I would have thought that this one would be accepted first, and then I'll merge master into that PR.

KrisThielemans:
> I would have thought that this one would be accepted first, and then I'll merge master into that PR.

sure, but I had made some comments there already (I think some of this is in that PR?). Anyway, feel free to accept this one first.

> And arguably, the synergistic CMake code should be more intellegent (ie. only build against engines that can be found).

exactly, so it needs to know which engines exist, so the search for the engines should be a level up.

src/xGadgetron/cGadgetron/CMakeLists.txt
KrisThielemans: ok

rijobro:
> exactly, so it needs to know which engines exist, so the search for the engines should be a level up.

Ok, good point.

> I think some of this is in that PR?

Ah, yes. I'd merged this one into that branch. I'll put the changes suggested over there into this PR to keep things separated.

KrisThielemans:
feel free to merge. I won't check this one again.

PR #277, opened 2018-12-21
Clone as sptr

rijobro:
We need this so that we can create copies of derived image classes, stored as the base ImageData class. 

We'll need this for registration etc., as the output will be a copy of one of the inputs (which will be stored as the abstract base class, ImageData).

### DIFF
--- a/doc/DeveloperGuide.md
+++ b/doc/DeveloperGuide.md
@@ -21,6 +21,7 @@
 	1. [Naming](#conventions_naming)
 	2. [File conventions](#conventions_file)
 	3. [Others](#conventions_others)
+6. [Cloning data container](#clone_datacontainer)
 
 # Overview <a name="Overview"></a>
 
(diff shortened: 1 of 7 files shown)


rijobro:
I not overly fussed about which name we choose, but `clone_as_sptr` seems to make the most sense to me.

If you agree, I'll rename `gadgetron::shared_ptr<MRAcquisitionData> clone()` as `gadgetron::shared_ptr<MRAcquisitionData> clone_as_sptr()` and then they would all be consistent. 

> I think most of these could use `make_shared`. 

We could, I can do that if you prefer (seems the same to me since they both call the constructor once).

> However, shouldn't we be using @evgueni-ovtchinnikov's strategy of having a virtual function that creates a ptr (using covariant return types), and just making the shared_ptr at base class level?

I'd rather not do that, since we'd be creating double the number of methods, and I'd err away from using raw pointers anyway.

Thoughts?

KrisThielemans:
This seems to overlap with #254. Doesn't that give you what you need? You might need of course some ugly casting. But that can be resolved implementing a `clone_as_sptr` in terms of `new_data_container_sptr` (doing the casting there). No other code needed.

In fact, somewhat surprisingly, this trick can be used with the same function name (using "hiding").

I've created a separate issue for the longer term design, see #278. I guess you don't need this PR if you can use `new_data_container_sptr`. (Note however that there's an extra constructor for `GadgetronImagesVector` in this PR.)

(By the way, I think `make_shared` is a tiny bit more conservative in memory usage than `shared_ptr(new T(...))` but it's probably the least of our worries).

rijobro:
`new_data_container_sptr` returns a blank copy of that image type. In hindsight, I don't need `new_data_container_sptr`, but I do need a `copy` method.

So, could we:
- close #254, 
- reopen #253 (which we had closed as it was merged into #254), and 
- use this PR?

If that's ok, I'll update this PR as per #278 so that `clone_as_sptr` is renamed as `copy` and returns a `std::unique_ptr`. I'll also use `make_shared`.

rijobro:
Implemented as per #278. Travis fine (as much as master).

PR #284, opened 2019-01-08
Write STIRImageData to different file formats

rijobro:
- Specify the format file (as you would in STIR).
- This is backwards compatible.
- Not exported to Python or Matlab (but would be simple to do so).


rijobro:
> looks good, but could use a default arg for the single-argument version to make code a bit more concise?

Don't think we can do that for inherited methods (inherited from `ImageData`).

> would need a doxygen with the file format of the text file. That should then be copied to the Python/MATLAB docstring as well (could do that later, if so, raise an issue).

Ok.

KrisThielemans:
> I know the data submodule is a pain, but if we accept this PR, it would be useful to put a few examples in `data/examples/PET/output_formats/` (or similar).
> 
> They could then be used in ctest as well as by users.
yeah. we're all afraid of updating it. However, we're going to have to do it for SIRFReg anyway I guess (or maybe that was already done?). If so, we can do this stuff now, do the SIRFReg one later, and then tell people they should update their copy (i.e. not just `git pull` I guess)

a ctest could (probably should?) write its own file easily enough of course.

By the way, these are going to be STIR specific, so I would make it clear in the name `STIR_output_format_xxx.par`. And by the way, nifti needs ITK of course, so would somehow have to say that.

rijobro:
> a ctest could (probably should?) write its own file easily enough of course.

Ok should I cancel the PR since changing the submodule is probably more effort than it's worth?

rijobro:
Ok, all done.

src/xSTIR/cSTIR/include/sirf/cSTIR/stir_data_containers.h
```diff
@@ -652,6 +652,17 @@ namespace sirf {
 		}
         /// Write to file
         virtual void write(const std::string &filename) const;
+        /// Write to file using format file. Example of the nifti format is below (which in turn requires ITK).
```
KrisThielemans: this will look rather horrible with Doxygen. Better to do something like
```doxygen
/// Write to file using format file
/*! This allows speciyfing the output file format used by STIR using a text file. 
     Keywords are specified by STIR.
    
    Example format...
   \verbatim
    OutputfileFormat Parameters:=
    ...
   \endverbatim
*/
```

KrisThielemans:
I guess I'd also put the example file in `examples/parameter_files` or so. I assume you will have some for SIRFReg as well.

I think we should call it `STIRbla.par` though (as that's what we call parameter files in STIR)

rijobro:
up to date again.

rijobro:
updated again.

PR #285, opened 2019-01-08
Improve MR write images

rijobro:
- Save the output when running `fully_sampled_recon.py`
- When saving MR images, use current date and time as default group

The latter point is preferable to current functionality (using 'dataset' by default), as this will not cause accidental concatenation of different reconstructions.

### DIFF
--- a/examples/Python/MR/fully_sampled_recon.py
+++ b/examples/Python/MR/fully_sampled_recon.py
@@ -11,6 +11,7 @@ Options:
   -p <path>, --path=<path>    path to data files, defaults to data/examples/MR
                               subfolder of SIRF root folder
   -e <engn>, --engine=<engn>  reconstruction engine [default: Gadgetron]
+  -o <outp>, --output=<path>  output file name [default: output.h5]
 '''
 
 ## CCP PETMR Synergistic Image Reconstruction Framework (SIRF).
@@ -82,6 +83,7 @@ def main():
     
     # retrieve reconstruced image data
     image_data = recon.get_output()
+    image_data.write(args['--output'])
 
     # show reconstructed image data
     image_array = image_data.as_array()
--- a/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
+++ b/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
@@ -38,6 +38,23 @@ using namespace sirf;
 std::string MRAcquisitionData::_storage_scheme;
 shared_ptr<MRAcquisitionData> MRAcquisitionData::acqs_templ_;
 
+static std::string get_date_time_string()
+{
+    time_t rawtime;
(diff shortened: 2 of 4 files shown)


rijobro:
This was suggested by @DANAJK elsewhere, and I just watched a Gadgetron tutorial in which they used the current date and time, so this seems like an acceptable approach.

DANAJK:
Yes, OK, but somehow gadgetron uses date and time by default I think - hence do we need to calculate the date and time groupname? (Prefer not to duplicate functionality).

rijobro:
Thanks @DANAJK. I've just had a look through @evgueni-ovtchinnikov 's wrapper code, but I can't figure out which ISMRMRD method is called eventually - and therefore I can't look for any default behaviour.

Any help @evgueni-ovtchinnikov ?

rijobro:
I got lost at this line of code: `IMAGE_PROCESSING_SWITCH_CONST`...

evgueni-ovtchinnikov:
i am not sure i understand what is the problem, but this is how sirf write mr images:

```
void
GadgetronImageData::write(std::string filename, std::string groupname)
{
	//if (images_.size() < 1)
	if (number() < 1)
		return;
	Mutex mtx;
	mtx.lock();
	ISMRMRD::Dataset dataset(filename.c_str(), groupname.c_str());
	mtx.unlock();
	for (unsigned int i = 0; i < number(); i++) {
		const ImageWrap& iw = image_wrap(i);
		iw.write(dataset);
	}
}
```

evgueni-ovtchinnikov:
just checked - ismrmrd dataset constructor does not have default groupname

rijobro:
So is everyone ok with this? Doesn't look like there's duplication, but it's only a few lines of code so it's not the end of the world if there is...

DANAJK:
Sorry, I was probably thinking of `gadgetron_ismrmrd_client`. It would be sensible to do the same as them in case people want to read in datasets from either SIRF or this program. Their way is:

std::string get_date_time_string()
{
    time_t rawtime;
    struct tm * timeinfo;
    time ( &rawtime );
    timeinfo = localtime ( &rawtime );

    std::stringstream str;
    str << timeinfo->tm_year+1900 << "-"
        << std::setw(2) << std::setfill('0') << timeinfo->tm_mon+1 << "-"
        << std::setw(2) << std::setfill('0') << timeinfo->tm_mday << " "
        << std::setw(2) << std::setfill('0') << timeinfo->tm_hour << ":"
        << std::setw(2) << std::setfill('0') << timeinfo->tm_min << ":"
        << std::setw(2) << std::setfill('0') << timeinfo->tm_sec;

    std::string ret = str.str();

    return ret;
}

rijobro:
Ok, I've updated such that it matches the Gadgetron way.

PR #287, opened 2019-01-09
Option do disable python/matlab

rijobro:
Although I have previously entered the matlab and python locations, I might want to disable one component or the other. At the moment, this isn't possible.

Behaviour largely the same as before. Difference: Default both to `ON`, so error will be thrown if one or the other can't be found. At that point, either point to the required location or disable.


KrisThielemans:
good idea. should we set the default according to if it's found or not?

rijobro:
Yes, but seemed a touch complicated since we'd have to check if the variable exists first so as not to overwrite it. Something like this?
```
if (enable_python doesn't exist)
   find python quiet
endif()
option(enable_python "use python" ${Python_found})
if (enable_python and not ${python_found})
    find python required
endif()
```

KrisThielemans:
thanks. seems that there's no good reason to `find` twice. what about
```
find python quiet
option(enable_python "use python" ${Python_found})
if (enable_python and not ${python_found})
    message(FATAL_ERROR "Python enabled but not found")
endif()

rijobro:
Good point, done.

PR #297, opened 2019-01-25
Use aliases in Matlab without copying folders

rijobro:
you can now do 
```
alias_name = set_up_PET([])
a = alias_name.ImageData(filename);
```
Small caveat - `alias_name` is a struct. `help(alias_name.ImageData())` will return the same help as `help(mSTIR.ImageData())`, but `help(alias_name)` will only return `aa is a variable of type struct.`.


rijobro:
We could actually now do away with the '[]' that was required in `set_up_PET()`. We would then replace `if isempty(engine)` with `if nargin == 0`. Looks a little nicer but would break backwards compatibility.

rijobro:
Lastly, functionality in `set_up_MR` and `set_up_PET` was identical except for the default engine they chose. So we created a function `set_up_engine`, which both `set_up_MR` and `set_up_PET` call.

src/common/set_up_engine.m
```diff
@@ -0,0 +1,49 @@
...
+        return;
+    end
+    
+    %% For returning as an alias
+    
+    % Get the folder containing the engine
+    filename = mfilename();
+    filepath = mfilename('fullpath');
+    l = length(filepath) - length(filename);
+    path = [filepath(1:l) '+m' engine '/'];
```
KrisThielemans: better to use
```matlab
path=fileparts(mfilename('fullpath'));
```
and use `fullfile` below

src/common/set_up_engine.m
```diff
@@ -0,0 +1,49 @@
...
+    
+    % Loop over all classes and functions and set alias to handle
+    files = dir([path '*.m']);
+    for i=1:size(files,1)
+        file = files(i).name(1:end-2); % remove .m
+        eval(['alias.' file ' = @m' engine '.' file ';']);
```
KrisThielemans: let's avoid `eval` as much as we can. Something like
```matlab
alias.(file) = ['@m' engine '.' file];
```
See https://blogs.mathworks.com/loren/2005/12/13/use-dynamic-field-references/

KrisThielemans:
> We could actually now do away with the '[]' that was required in set_up_PET(). We would then replace if isempty(engine) with if nargin == 0. Looks a little nicer but would break backwards compatibility.

could keep backwards compatibility by checking for `nargin==0` and `'[]`

src/common/set_up_engine.m
```diff
@@ -0,0 +1,46 @@
...
+    end
+    
+    if nargout == 0 
+        return;
+    end
+    
+    %% For returning as an alias
+    
+    % Get the folder containing the engine
+    path = [fileparts(mfilename('fullpath')) '/+m' engine '/'];
```
KrisThielemans: likely not portable to Windows. Use 'fullfile' or 'filesep'

src/common/set_up_engine.m
```diff
@@ -0,0 +1,46 @@
+function alias = set_up_engine(engine)
+% Creates a string, evaluating which imports PET engine.
```
KrisThielemans: this docstring doesn't make a lot of sense to me. When there's no output-arg, there's no string being created. And when there is, `alias` is a `struct`.

Also, it refers to PET.

Best to include an example in the docstring. also add
```matlab
% See also set_up_PET and set_up_MR
```
(see https://www.mathworks.com/help/matlab/matlab_prog/add-help-for-your-program.html)

similar changes required in the docstring of `set_up_PET` etc

src/common/set_up_engine.m
rijobro: ok, my new implementation moves the need of the `/` to here:
```
files = dir([path '/*.m']);
```
From the [Matlab doc](https://uk.mathworks.com/help/matlab/ref/dir.html), I think this is ok:
> For example, dir \*/*.txt lists all files with a txt extension exactly one folder under the current folder

KrisThielemans: sorry. why not
```
files = dir(fullfile(path, '*.m'));
```

rijobro: much better!

PR #301, opened 2019-02-05
Add getter for number of total voxels to NiftiImageData

johannesmayer:
This function is useful to loop over the elements.

### DIFF
--- a/src/Registration/cReg/NiftiImageData.cpp
+++ b/src/Registration/cReg/NiftiImageData.cpp
@@ -364,6 +364,12 @@ const int* NiftiImageData<dataType>::get_dimensions() const
 }
 
 template<class dataType>
+size_t NiftiImageData<dataType>::get_num_voxels() const
+{
+    return _nifti_image->nvox;
+}
+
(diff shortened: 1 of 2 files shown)


rijobro:
Merged. Just for interest, I was using `get_raw_nifti_sptr()->nvox` before.

KrisThielemans:
can I ask how you're using this?

If we need it, we should have a virtual function higher up (which could just get it from the geo_info)

rijobro:
Apologies for merging, do you want me to revert? 

Although we should be using geom_info, I don't fully trust it quite yet (e.g., varying slice distances for MR images). Should we make this an issue to solve in the future?

johannesmayer:
I decided to load the segmentation input of my simulation into a 'NiftiImageData3D' (no modality attached, just geometry and image information).

At some points I want to loop over all the pixels in the image and assign them to a specific tissue type.
I didn't see the functionality which Richard just suggested, so my code became very messy.

Don't think personally I will use it very often on top of that.

PR #310, opened 2019-02-13
Fixes #309

### LINKED ISSUE #309: Raw pointer type of sorting index in gadgetron_data_containers
The data member `int* index_` keeping track of sorting images and acquisitions is a raw pointer.
I encountered a related segfault at some point.

johannesmayer:
The sorting index was previously used as a raw pointer using `new` and `delete[]`.
This has been replaced by a vector, with size changes handled by `std::vector::resize()`.

Types of function return arguments of `int* index(...)` are left the same but the adress of  the zeroth element
of the vector is returned instead. Hence other code remains compatible.

Also I deleted the destructor previously keeping track of deleting the index.


src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
```diff
@@ -549,17 +550,17 @@ AcquisitionsFile::take_over(MRAcquisitionData& ac)
 {
 	AcquisitionsFile& af = (AcquisitionsFile&)ac;
 	acqs_info_ = ac.acquisitions_info();
-	if (index_)
-		delete[] index_;
+	if (index_.size() > 0)
+		index_.resize(0);
 	int* index = ac.index();
 	sorted_ = ac.sorted();
 	if (sorted_ && index) {
```
KrisThielemans: I guess this test on `index` will now fail. I suppose it was possible that it was a 0 pointer before, but now it will never be. I'm not so sure what would happen when calling the new `index()` (i.e. `&index_[0]`) if `index_.size()==0`.

I guess it now has to be
```C++
if (sorted_ && ac.index_.size()>0)
```
or variations on that.

However, @evgueni-ovtchinnikov is the original logic correct? It will set `index_` to empty if `ac` is non-sorted. Is that intentional?

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
```diff
@@ -549,17 +550,17 @@ AcquisitionsFile::take_over(MRAcquisitionData& ac)
 {
 	AcquisitionsFile& af = (AcquisitionsFile&)ac;
 	acqs_info_ = ac.acquisitions_info();
-	if (index_)
-		delete[] index_;
+	if (index_.size() > 0)
+		index_.resize(0);
```
KrisThielemans: we don't need these 2 lines anymore as far as I can see. The code below will always re-assign or resize. We don't need to first resize. 2 lines less :-)

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
```diff
@@ -763,10 +764,10 @@ GadgetronImageData::sort()
         std::cout << "Before sorting. Image " << i << "/" << ni <<  ", Contrast: " << t[0] << ", Repetition: " << t[1] << ", Projection: " << t[2] << "\n";
 #endif
 	}
-	if (index_)
-		delete[] index_;
-	index_ = new int[ni];
-	Multisort::sort(vt, index_);
+	if (index_.size() > 0)
```
KrisThielemans: can delete these 2 lines, as you'll resize below anyway

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
```diff
@@ -549,17 +550,17 @@ AcquisitionsFile::take_over(MRAcquisitionData& ac)
...
+		index_.resize(0);
 	int* index = ac.index();
 	sorted_ = ac.sorted();
 	if (sorted_ && index) {
 		unsigned int n = number();
-		index_ = new int[n];
-		memcpy(index_, index, n*sizeof(int));
+		index_.resize(n);
+		memcpy(&index_[0] , index, n*sizeof(int));
```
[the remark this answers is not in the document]
johannesmayer: I changed the return types of the index() function to return vectors so there can be a direct assignment

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
johannesmayer: Done, also did this in some other places where there was unnecessary resizing.

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
johannesmayer: no in this case `int* index` is not `index()`, s.t. the if asks if the `ac.index_` object is empty.
I changed the name to `other_index` to distinguish it a bit more and deleted the resize.

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
```diff
@@ -549,17 +556,16 @@ AcquisitionsFile::take_over(MRAcquisitionData& ac)
 {
 	AcquisitionsFile& af = (AcquisitionsFile&)ac;
 	acqs_info_ = ac.acquisitions_info();
-	if (index_)
-		delete[] index_;
-	int* index = ac.index();
+	
+	std::vector<int> other_index = ac.index();
```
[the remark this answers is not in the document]
KrisThielemans: replace all with `index_ = ac.index_;`

johannesmayer:
Sets sorted_ to true if the `sort_by_time()` function is called.

Also deleted the two default constructors for `MRAcquisitionData` and `ISMRMRDImageData` since all they did was initialize `sorted_` with false, and now it is set in the class instead.

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
```diff
@@ -493,11 +498,14 @@ MRAcquisitionData::sort_by_time()
...
-	index_ = new int[num_acquis];
-	Multisort::sort( vt ,index_ );
+	index_.resize(num_acquis);
+	
+	if( num_acquis <= 0 )
+		std::cerr << "WARNING: You try to sort by time an empty container of acquisition data." << std::endl;
+	else
+		Multisort::sort( vt, &index_[0] );
 
+	sorted_ = true;
```
evgueni-ovtchinnikov: Not a good idea IMHO - methods that compute coil sensitivity maps check whether the acquisitions are sorted **with respect to slices** (at present, on Python/Matlab level, but will be moved down to c++ eventually), rather call it `sorted_by_time_` and add method `sorted_by_time()` (or else add argument to `sorted()` specifying the kind of sort - actually, `sort()` could have it too, instead of your separate sorting method).

KrisThielemans:
> One cannot access protected members of a parent class type.

that doesn't make sense. It's exactly what `protected` does (vs `private`). maybe you can give the compilation error that you had?

KrisThielemans:
also worth merging `master` again, as Travis should now be green.

johannesmayer:
> > One cannot access protected members of a parent class type.
> 
> that doesn't make sense. It's exactly what `protected` does (vs `private`). maybe you can give the compilation error that you had?

I wrote a minimal program which gives the same compilation error:
Abstract base class, publically inherited child class with some method trying to access the protected member of a parent class reference.
(Which in our case translates to: `AcquisitionsFile::take_over(MRAcquisitionData& ac)`)
```
// Example program
#include <iostream>
#include <string>

class A{
    public:
        A(){ a_=0; }
        int get_a(){ return a_; };
        virtual void set_a( int const val, A& parent_reference )=0;

    protected:
        int a_;
};

class B: public A{
    public:
        void set_a( int const val, A& parent_reference ){ parent_reference.a_ = val; };
};

int main()
{
    
    B b;

    b.set_a( 10, b);
    
    std::cout << "Instance of child class has the member valued at :" << b.get_a() <<std::endl;

}
```
This gives the error:
```
 In member function 'virtual void B::set_a(int, A&)':
14:13: error: 'int A::a_' is protected
19:76: error: within this context
 ```

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
johannesmayer: Agreed, in the `index(int i)` there is not even a check any more if `sorted_== true`, just if `index.size()>0`. So a bool variable to check if it is time sorted is not necessarry atm.

KrisThielemans:
oh well. unexpected!

I'd then try to make `index()` a `protected` member...

By the way as it's protected, I'd be ok with letting it return a `const vector<>&`

johannesmayer:
Well, since it won't allow access to protected members then I don't think that calling a protected member function is possible. 
Also this cast 

https://github.com/CCPPETMR/SIRF/blob/c958113fd26b6b3760132bb68632b03658dc97bb/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp#L557-L558

did not enable this but I don't even know if that was the aim of that line.
So `index()` will have to remain public until the `take_over()` method does not need to access the index any more.

I changed the return type of `index const()`.

KrisThielemans:
So... a `B` object can access protected members of `A` in another `B` object, but it cannot access them in an `A` object (here's [someone else struggling with this](https://stackoverflow.com/questions/47116046/access-protected-members-from-a-public-derived-class)). Seems a bit weird to me, but fine.

PR #311, opened 2019-02-14
better geom info check and printing

rijobro:
method to print ISMRMRD header 
better checks when doing set_up_geom_info()


src/xGadgetron/cGadgetron/include/sirf/cGadgetron/gadgetron_data_containers.h
```diff
@@ -751,6 +751,9 @@ namespace sirf {
             return std::unique_ptr<GadgetronImagesVector>(this->clone_impl());
         }
 
+        /// Print header info
+        void print_header(const unsigned im_num=0);
```
KrisThielemans: would be more useful if we'd do
```C++
void print_header(std::ostream&, const unsigned im_num=0);
```

Also, I find it not so intuitive to have a default `im_num=0`. It'd make more sense to me if the `im_num` is required. We could then later on add a `print_header(std::ostream&)` which loops over all available images.

rijobro: ok, done

PR #314, opened 2019-02-20
get the voxel sizes in matlab and python for niftiimagedata

rijobro:


KrisThielemans:
hi. a bit weird to have this on the nifti side only. I think we had this discussion recently that this would be a temporary thing as really it should be coming from the geoinfo. Was there an issue to put it in `ImageData` later?

rijobro:
It's tricky because on the nifti side, it's an 8D array (allowing for tensors and multiple time points). If we were to put it in to `ImageData`, it would only be 3D.

Sorry I rushed this one, if you want I'll create an issue to tidy it up when I'm back.

KrisThielemans:
ah. these are the Nifti 8x1 arrays. those obviously only make sense for Nifti. I'd call the functions somewhat different then though. as I don't think it'll work for for us to squeeze ISMRMRD dimensions into the nifti conventions.  `get_nifti_dimensions` etc? For me, `get_voxel_size` is a 3D concept though.

PR #328, opened 2019-03-17
improve handling of the matrix in Python AcquisitionModelUsingRayTracingMatrix

KrisThielemans:
Previously, get_matrix() created a new RayTracingMatrix, therefore loosing any settings we had. It's now disabled as it seems we cannot create a Python object from a handle. set_num_tangential_LORs now works directly in terms of the matrix handle.

Also added get_num_tangential_LORs


evgueni-ovtchinnikov:
> Previously, get_matrix() created a new RayTracingMatrix, therefore loosing any settings we had.

wrong, `get_matrix()` returns a `RayTracingMatrix` object that is essentially a reference to the `RayTracingMatrix` that was previously set either by the constructor or `set_matrix` method of an `AcquisitionModelUsingMatrix`, so the settings are not lost, e.g. the number of tangential LORs can be obtained by `acq_model.get_matrix().get_num_tangential_LORs()`

KrisThielemans:
ok , I see now. `RayTracingMatrix` is almost like an empty shell. It's its `handle` that does the stuff. Great.

I've reverted some changes but kept a few others. What do you think?

Issue #330, opened 2019-03-18
STIR reconstruction error due to unsupported z-grid spacing


ckolbPTB:
I run the example script "reconstruct_from_listmode.py" with

"sirf@sirf-VirtualBox:~/devel/build/sources/SIRF/examples/Python/PET$ python reconstruct_from_listmode.py " without any modifications, so using the provided phantom data. I ran the script 10 times. On a few runs the script finished successfully, but the image I got in the last figure which pops up was zero everywhere. No error message though.

On all other runs I got the same basic error message saying that the image z-spacing does not fit to the ring spaces but always with different values for the image z-spacing. Here are some examples of the error messages:

RUN 1
applying attenuation (please wait, may take a while)...
setting up, please wait...

ERROR: DataSymmetriesForBins_PET_CartesianGrid can currently only support z-grid spacing equal to the ring spacing of the scanner divided by an integer. Sorry. (Image z-spacing is 0.000267 and ring spacing is 4.062500)

RUN 2
applying attenuation (please wait, may take a while)...
setting up, please wait...

ERROR: DataSymmetriesForBins_PET_CartesianGrid can currently only support z-grid spacing equal to the ring spacing of the scanner divided by an integer. Sorry. (Image z-spacing is -0.234735 and ring spacing is 4.062500)

RUN 3
applying attenuation (please wait, may take a while)...
setting up, please wait...

ERROR: DataSymmetriesForBins_PET_CartesianGrid can currently only support z-grid spacing equal to the ring spacing of the scanner divided by an integer. Sorry. (Image z-spacing is -0.000000 and ring spacing is 4.062500)

I guess this means that there is a memory problem somewhere, i.e. that the image z-spacing is reading out a random part of memory...

I am running it on a Ubuntu 18.04 virtual machine using a SIRF_SuperBuild with DDEVEL_BUILD=ON

rijobro:
@paskino, if @ckolbPTB is using the most recent VM, could you test it (might mean there's a bug in the VM). I'll try on my custom VM and we'll see if that's the problem.

paskino:
I've tested it in the newest VM and it works.
https://github.com/CCPPETMR/SIRF/issues/333#issuecomment-473886056

rijobro:
I unfortunately get the same error as Christoph (on a non-official VM). I suppose this hints at it being an inherent SIRF/STIR problem...

KrisThielemans:
can you use this `valgrind` on this? Just
[...]

Issue #333, opened 2019-03-18
Fail to run reconstruct_from_listmode


paskino:
@ckolbPTB reported by email the following issue.

I tried to run the demo "reconstruct_from_listmode" and I received the following error: 

```
File: /home/sirf/devel/build/INSTALL/python/sirf/STIR.py 
Line: 1666 
check_status found the following message sent from the engine: 
'DataSymmetriesForBins_PET_CartesianGrid can currently only support z-grid spacing equal to the ring spacing of the scanner divided by an integer. Sorry\n' exception caught at line 675 of /home/sirf/devel/build/sources/SIRF/src/xSTIR/cSTIR/cstir.cpp 
the reconstruction engine output may provide more information 
```

paskino:
On the yet-to-release Ubuntu18.04 VM the code runs without issue.

rijobro:
duplicate from #330.

paskino:
How come it's duplicate? I don't see the error. What versions are you running.

rijobro:
No, just that Christoph already created the issue, so we should be using that thread instead.

PR #335, opened 2019-03-19
Fix python recon lm example

rijobro:
Problem was a faulty C-style cast that was causing voxel sizes to be incorrect.


rijobro:
fixes #330 (and #333). Problem was due to a faulty C-style cast.

PR #374, opened 2019-04-29
Fix PET interactive scripts

### LINKED ISSUE #370: Python/PET/interactive demos fail on VM
The script `display_and_projection.py` runs ok, but subsequent running of any of the rest fail.

In Spyder (started from shell), running further scripts in the same console produces error messages suggesting that `pystir` is `None`, whereas running them in a separate dedicated console produces error message 'name pet is not defined'.

Richard checked that they also fail in IPython run in a terminal, so it is not a Spyder problem.

paskino:
closes #370
fix functionality of pet interactive examples
renamed files to clarify the order they should be run
uses iterno instead of iter
uses sliceno instead of slice


paskino:
Still standing issue: running the whole script in spyder doesn't run the whole script.
Running the script cell-by-cell works.

paskino:
any comments?

examples/Python/PET/interactive/01-display_and_projection.py
```diff
@@ -198,5 +198,5 @@ def make_cylindrical_FOV(image):
 
 #%% close all plots
 plt.close('all')
-
+print ("here")
```
KrisThielemans: delete

examples/Python/PET/interactive/03-basic_reconstruction.py
```diff
@@ -69,13 +69,14 @@
 
 #%% display of image
 reconstructed_array=reconstructed_image.as_array()
-slice=reconstructed_array.shape[0]//3;
-show_2D_array('reconstructed image after 5 sub-iterations',reconstructed_array[slice,:,:,]);
+# slice=reconstructed_array.shape[0]//3;
+sliceno = 9
```
KrisThielemans: I'd prefer to have it in terms of the `shape` somehow, but fine to leave if that's too hard

KrisThielemans:
one other thing: you didn't update the README...

examples/Python/PET/interactive/03-basic_reconstruction.py
paskino: > I have the same comments as @evgueni-ovtchinnikov
> 
> * rename `sliceno` to `slice_num` etc

OK

> * why remove `help`?

I removed `help` thinking that it may fix the problem that that was the reason spyder wouldn't run the whole script correctly. As that is not the case I'll put it back. However, the problem of spyder reaching the end without running correctly still remains.

paskino:
According to the [README](https://github.com/CCPPETMR/SIRF/blob/pet_interactive_example/examples/Python/PET/interactive/README.md) the scripts should be run cell-by-cell.

PR #384, opened 2019-05-14
print all wrapped NiftyReg methods

rijobro:


KrisThielemans:
I wasn't entirely happy with this PR.

For Python and Matlab, the built-in `help` should have done the trick. Moreover, printing to `std::cout` often doesn't work (when using a GUI) as we know (if we don't have an issue for that, we should...). For C++, this function makes little sense.

So why do we need this?

rijobro:
I think this does make sense because some of the wrapped methods are set via a string or via a parameter file; they don't directly exist as a C++ method (nor python/matlab). e.g.,
```
reg = NiftyAladinSym();
reg.set_parameter('SetPerformRigid','1');
```
Without a `print_all_wrapped_methods` method, you would have no way of knowing which parameters you could set without delving through the source code. Does that make sense? I'm happy to change the code if you're still not happy.

Also, I agree with the problem of `std::cout`, but would hopefully be solved one day with a logger (as discussed here: https://github.com/CCPPETMR/SIRF/issues/332).

KrisThielemans:
ok

Issue #397, opened 2019-06-03
code duplication in matlab and python get/set parameters


rijobro:
For both Matlab and Python, each engine has its own set of methods for getting and setting parameters:

- For Matlab, this means that each engine has files named `parameter.m` and `setParameter.m`.
- For Python this means that each file (e.g., `pSTIR.py`, `pGadgetron.py`, `pReg.py`) starts with these definitions.

These methods should all be grouped together. Naturally this reduces code duplication and makes mistakes less likely.

PR #399, opened 2019-06-07
Removed duplication of parameter get/set functions in Python (addresses #397)

evgueni-ovtchinnikov:


DANAJK:
`def int_pars` in `parameters.py` is missing an initial declaration for variable `value`?

rijobro:
Could you take care of the registration stuff, too? I followed what you did for STIR and Gadgetron, so there shouldn't be any surprises.

evgueni-ovtchinnikov:
Well spotted - thanks David!

evgueni-ovtchinnikov:
@rijobro: I notice you have `_sirf` suffix in your parameter methods in Reg.py - is it ok to drop it?

rijobro:
yes, absolutely. Thanks.

evgueni-ovtchinnikov:
@rijobro: updated Reg.py, but could not test - my SIRF build does not install it for some reason (I did check `REG_ENABLE` in CMake), hopefully Travis will do it

rijobro:
I got the following error for the python registration ctest 
```
ERROR: Failure: ModuleNotFoundError (No module named 'sirf.select_module')
```

rijobro:
In fact, I get the same error for all of SIRF's python ctests.

evgueni-ovtchinnikov:
i vaguely remember seeing this kind of message, which disappeared when i put debugging print in `CMakeLists.txt` in common - it appears this forced cmake to notice that `select_module.py` and `parameters.py` need to be installed

btw travis fails because you have `cReg_` prefix before `setParameter` and `parameter` in `cReg.h`

src/Registration/pReg/Reg.py
```diff
@@ -44,89 +49,15 @@
 ERROR_CHANNEL = 2
 ALL_CHANNELS = -1
 
-###########################################################
-############ Utilities for internal use only ##############
-
 
-def _setParameter_sirf(hs, set_, par, hv, stack=None):
-    # try_calling(pyreg.cReg_setParameter(hs, set, par, hv))
+def _set_parameter(hs, group, par, hv, stack = None):
```
rijobro: Can't this one be unified (and therefore deleted from here)?

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -41,6 +41,10 @@
 import sirf.pygadgetron as pygadgetron
 import sirf.pysirf as pysirf
 
+import sirf.select_module as select_module
+select_module.module = 'pygadgetron'
+import sirf.parameters as parms
```
rijobro: should these all be `params`?

src/Registration/pReg/Reg.py
evgueni-ovtchinnikov: this is just a quick fix to sort out the error you hit first

rijobro: which error? My initial error was just that the `select_module.py` and `parameters.py` weren't getting installed.

rijobro:
I currently get this error:
```
ImportError: cannot import name 'setParameter' from 'sirf.pyreg' (/Users/rich/Documents/Code/SIRF/Install/python/sirf/pyreg.py)
```

evgueni-ovtchinnikov:
this is because you have `cReg_` prefix in front of `setParameter` and `parameter` in `cReg.h`

src/xGadgetron/pGadgetron/Gadgetron.py
evgueni-ovtchinnikov: no objections, could even be `parameters` or `set_get_methods` etc etc

src/Registration/pReg/Reg.py
evgueni-ovtchinnikov: have you fixed this?

evgueni-ovtchinnikov:
@rijobro: to fix travis build myself, tried to build SIRF with registration (this time, succeeded in generating the project with it), got these errors:
```
3>C:\Users\wps46139\Documents\GitHub\build\SIRF-SuperBuild\sources\SIRF\src\Registration\cReg\include\sirf/Reg/ParserKey.h(228): error C2664: 'void sirf::ParserKeyBase<Z>::get_argument(const std::string &,const int,unsigned long &) const': cannot convert argument 3 from 'std::size_t' to 'const char *&'

3>C:\Users\wps46139\Documents\GitHub\build\SIRF-SuperBuild\sources\SIRF\src\Registration\cReg\NiftiImageData.cpp(1135): error C2872: 'NREG_CONV_KERNEL_TYPE': ambiguous symbol
3>  C:\Users\wps46139\Documents\GitHub\build\SIRF-SuperBuild\INSTALL\include\_reg_tools.h(27): note: could be 'NREG_CONV_KERNEL_TYPE NREG_CONV_KERNEL_TYPE'
```
fixed the first one with the explicit cast

could the second be caused by `NREG_CONV_KERNEL_TYPE` somehow defined twice?

if cannot fix this quickly, perhaps you could just remove `cReg_` prefix yourself, i.e. have in cReg.h (and cReg.cpp)
```
    void* setParameter(void* ptr, const char* obj, const char* name, const void* value);
    void* parameter(const void* ptr, const char* obj, const char* name);
```
i will then create an issue about `NREG_CONV_KERNEL_TYPE`

rijobro:
For your compilation we'll have a look when we're together tomorrow. I don't think it's ever been compiled on Windows, so there's surely a complication due to that. But it would be good to sort it out for any other Windows users in the future (including yourself).

And I'll try and fix the `cReg` business now.

rijobro:
ctests now pass for the registration components on my machine. Could you check it over and make sure my changes agree with yours (not sure I understood everything you were doing).

Two comments: 

- In `STIR.py` you use: `parms.setParameter`. However, I can't see that method defined anywhere. Should it be: `parms._set_parameter`?
- In matlab, we still have `parameter.m` and `setParameter.m`, which could be unified.

If I've correctly understood the first bullet point, then it's worrying that they didn't get picked up by the ctest.

PR #405, opened 2019-06-25
no functions in matlab scripts

rijobro:


evgueni-ovtchinnikov:
`test_mReg` throws
```
Error using sirf.Reg.ImageWeightedMean
Error: File: ImageWeightedMean.m Line: 50 Column: 23
The input character is not valid in MATLAB statements or expressions.

Error in test_mReg (line 670)
		wm1 = sirf.Reg.ImageWeightedMean();
```
because of " used in line 50 of ImageWeightedMean.m (must be ')

rijobro:
> because of " used in line 50 of ImageWeightedMean.m (must be ')

Just changed. Could you give it a go now?

evgueni-ovtchinnikov:
i did already before commenting (edited that line myself) to make sure it works now

btw why not put folder results in mReg/test in .gitignore?

rijobro:
> btw why not put folder results in mReg/test in .gitignore?

Why would we want to do that? Doing so would mean that we wouldn't track any changes to `test_mReg.m` and the `CMakeLists.txt`, which we certainly want to do.

rijobro:
ah sorry, i misread your comment. But anyway, the results of the ctest should go into your binary folder, so shouldn't be affected by the gitignore.

PR #411, opened 2019-07-02
reg: take slope and intercept into account

rijobro:


rijobro:
closing as superceded by #416.

PR #416, opened 2019-07-03
reg: take slope and intercept into account

rijobro:
upon reading an image, convert its array such that data=slope*data+intercept. Then set slope and intercept to 0 and 1 respectively.

### DIFF
--- a/src/Registration/cReg/NiftiImageData.cpp
+++ b/src/Registration/cReg/NiftiImageData.cpp
@@ -733,6 +733,15 @@ void NiftiImageData<dataType>::set_up_data(const int original_datatype)
     _nifti_image->nbyper = sizeof(float);
     this->_data = static_cast<float*>(_nifti_image->data);
 
+    // Take slope and intercept into account
+    if (std::abs(_nifti_image->scl_slope-1) > 1e-4f || std::abs(_nifti_image->scl_inter) > 1e-4f) {
+        for (unsigned i=0; i<this->get_num_voxels(); ++i)
+            _data[i] = _nifti_image->scl_slope * _data[i] + _nifti_image->scl_inter;
+        _nifti_image->scl_slope = 1.f;
+        _nifti_image->scl_inter = 0.f;
+
+    }
+
     // Lastly, initialise the geometrical info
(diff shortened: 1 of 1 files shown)


src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -222,6 +226,12 @@ NiftiImageData<dataType> NiftiImageData<dataType>::operator*(const float& val) c
     return maths(val,mul);
 }
 
+template<class dataType>
+NiftiImageData<dataType> NiftiImageData<dataType>::operator/(const float& val) const
```
KrisThielemans:
```suggestion
NiftiImageData<dataType> NiftiImageData<dataType>::operator/(const float val) const
```
generally never use a ref for `const float/int` etc. it's unclear and would just slow it down.

KrisThielemans: I see it's done with `*` as well though. Maybe accept this one, and then do a separate PR for all of those (not urgent)

src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -61,7 +61,9 @@ template<class dataType>
 NiftiImageData<dataType>::NiftiImageData(const NiftiImageData<dataType>& to_copy)
 {
     copy_nifti_image(_nifti_image,to_copy._nifti_image);
-    set_up_data(to_copy._original_datatype);
+    this->_data = static_cast<float*>(_nifti_image->data);
```
KrisThielemans: I'll take your word for it (is it always `float`?)

rijobro:
When reading an image, I convert to `float` (and convert back at the end). This means that the static cast from `void *` to `float` is safe.

Nifti images in general can be complex, but I only accept a subset of all nifti images, so this isn't a problem. Accepted data types: DT_BINARY, DT_INT8, DT_INT16, DT_INT32,  DT_FLOAT32, DT_FLOAT64, DT_UINT8, DT_UINT16, DT_UINT32, DT_INT64, DT_UINT64, DT_FLOAT128.

PR #419, opened 2019-07-16
synergistic: include stir last

rijobro:
Updated to boost 1.70 on OSX and error created upon building synergistic code (code that builds on STIR, Gadgetron and registration). This was solved by including STIR components last in synergistic code. 

Not sure why this was necessary, nor why it solved the problem...


KrisThielemans:
This is worrying. It probably means that gadgetron vs STIR containers include a different boost file or set a different define (could be due to  the annoying `std::shared_ptr` STIR workaround). Please create an issue with the errors that you observe without this fix.

Diagnosis will likely mean running `gcc -E` to output the preprocessed files and check.

PR #441, opened 2019-09-24
MR geom info

rijobro:


rijobro:
We're currently not capable of correctly placing ISMRMRD images in 3D space. Without this, we cannot convert between STIRImageData, GadgetronImageData and NiftiImageData, which is necessary for synergistic algorithms. 

I've been thinking about this for a while, but still haven't managed to figure it out. I'm unsure of how to calculate direction matrix from the unit vectors `SliceDir`, `ReadDir` and `PhaseDir`, and the offset (I think) needs to come from `Position`.

rijobro:
This relates to these issues:
https://github.com/gadgetron/gadgetron/issues/530
https://github.com/ismrmrd/ismrmrd/issues/58
https://github.com/ismrmrd/ismrmrd/issues/132

rijobro:
@DANAJK, @KrisThielemans - do you remember looking at this together about a year ago? Do you think we could use the following Gadgetron writer to DICOM to help? 

https://github.com/gadgetron/gadgetron/blob/f95e51dd1f75c8af6e1916c3ef94cf3e20e4b194/gadgets/dicom/dicom_ismrmrd_utility.cpp#L699-L706

DANAJK:
I remember we looked at this during the KCL Hackathon and got bogged down in definitions about field of view.
My rough expectation is that you can use `ReadDir` and `PhaseDir` directly (as in the Gadgetron writer) to fill the `ImageOrientationPatient` 6-element vector. The readout and phase encode directions do not change after Fourier Transform.

The `ImagePositionPatient` is a 3-element vector that points to the centre of the top left image pixel, whereas `Position` in ISMRMRD is _probably_ the coordinate of the centre of the volume in 3D. Here 'centre' means where in space the image origin is located after Fourier Transform from k-space. For example, the even-length array 1,2,3,... 64 would have its 'centre' at point 33 (in this 1-based system). To figure out `ImagePositionPatient`, you will also need the reconstructed image voxel sizes.

PR #445, opened 2019-09-26
copy metadata from acq to image

rijobro:
SIRF's wrapper around the acquisition data takes into account the XML meta data. However, until now images didn't. 
So in this PR, when doing a recon, copy the meta data from the acquisition data to the image. Make sure that images read, write and store metadata if it's present.


rijobro:
@evgueni-ovtchinnikov 

Everything works as expected, but I now get the following error:
```
ERROR: File Error in /Users/rich/Documents/Code/SIRF-SuperBuild/Build/sources/ISMRMRD/libsrc/dataset.c, line 848: Path to element not found.
ERROR: File Error in /Users/rich/Documents/Code/SIRF-SuperBuild/Build/sources/ISMRMRD/libsrc/dataset.c, line 1364: Failed to read image header.
```
Do you see this? Any ideas on how to fix it? I'm tempted to merge anyway as all works, but I'm a little scared. Could you try this on some of your data?

KrisThielemans:
seems that you need to check if something exists before trying to read it

evgueni-ovtchinnikov:
@rijobro: which script throws this error and on which data?

Issue #447, opened 2019-09-27
Inclusion of HKEM code requires recent version of STIR


rijobro:
HKEM has been implemented in SIRF (https://github.com/CCPPETMR/SIRF/pull/446) following merging PR in STIR (https://github.com/UCL/STIR/pull/186).

The master branch of SIRF now requires headers that are only available in STIR since this PR, meaning that new versions of SIRF compiling against older versions of STIR won't work. 

This could be fixed by incrementing STIR version (e.g., 3.2.0), and in SIRF CMake checking the `STIRConfigVersion.cmake`. In the meantime, I'll fix this by checking for the existence of `KOSMAPOSLReconstruction.h`. If not available, don't compile the relevant code.

PR #448, opened 2019-09-27
Only build HKEM if version of STIR is recent enough

rijobro:
Only build HKEM if version of STIR is recent enough


rijobro:
this fixes #447.

PR #467, opened 2019-10-29
set verbosity of stir reconstructions

rijobro:
requires https://github.com/UCL/STIR/pull/411


KrisThielemans:
As STIR's verbosity is a global setting, it makes more sense to expose `stir::Verbosity::set` and `get`, https://github.com/UCL/STIR/blob/e2f0bcb74af6bbf067785e07f58a8430faf3c64d/src/include/stir/Verbosity.h#L42

could you have `sirf.stir.set_verbosity(int)` and `get`

PR #490, opened 2019-12-12
zoom image from python

rijobro:


src/xSTIR/cSTIR/cstir.cpp
```diff
@@ -962,6 +962,37 @@ void* cSTIR_writeImage(void* ptr_i, const char* filename)
 	CATCH;
 }
 
+extern "C"
+void* cSTIR_ImageData_zoom_image(void* ptr_im, const size_t zooms_ptr_raw, const size_t offsets_in_mm_ptr_raw,
```
KrisThielemans: from the .h,seems that `size_t`->`PTR_FLOAT`

src/xSTIR/cSTIR/cstir.cpp
```diff
@@ -962,6 +962,37 @@ void* cSTIR_writeImage(void* ptr_i, const char* filename)
...
+                                 const size_t new_sizes_ptr_raw, const char *zoom_options)
```
KrisThielemans:
```suggestion
                                 const size_t new_sizes_ptr_raw, const char *const zoom_options)
```

src/xSTIR/cSTIR/include/sirf/STIR/stir_data_containers.h
```diff
@@ -43,6 +43,7 @@ limitations under the License.
 #include "sirf/common/PETImageData.h"
 #include "sirf/STIR/stir_types.h"
 #include "sirf/common/GeometricalInfo.h"
+#include "stir/zoom.h"
```
KrisThielemans:
```suggestion
#include "stir/ZoomOptions.h"
```

src/xSTIR/cSTIR/cstir.cpp
rijobro: think it can be anything (if seen `void *` elsewhere) as it goes through a C-style cast.

PR #493, opened 2019-12-13
linux to use apt install itk

rijobro:


KrisThielemans:
I haven't checked, but I think we should have at least one without ITK, and one where we build it ourselves.

rijobro:
Already left one that builds ITK itself, but you're right, one should be without. I'll do that now.

PR #519, opened 2020-01-29
MRAcq.Data Getter and Setter fail if not sorted

johannesmayer:
since the index is accessed in the getter and setter, if not sorted there is nothing to access.

now just checks if index_ has size larger than zero, could be changed into checking for sorted() method also but then sort_by_time() does not set any boolean yet.


johannesmayer:
Hello Evgueni,

sorry I misspoke above, because I was not aware of the real reason.
The problem is, that the `index()` method is not virtual, and hence, references to `MRAcquisitionData&` call the base class implementation which throws an exception.

I will fix this later and will update this PR.

PR #531, opened 2020-02-04
Inverse deformation

rijobro:
Inverse deformation of `NiftiImageData3DDeformation`. Displacements can be done too, by converting to deformations, getting the inverse then converting back to a displacement again.


rijobro:
@ALEXJAZZ008008 Think you were interested in the inverse transformation last week. Could you give this a go for me? Refer to the python test for the syntax.

PR #533, opened 2020-02-04
GadgetronImageData::set_meta_data to all set_up_geom_info

### LINKED ISSUE #527: get_output() does not call set_up_geom_info()
When using e.g. FullySampledReconstructor() to reconstruct and MR image and then call get_output() to get an Gadgetron.ImageData object, set_up_geom_info() is not called. It is only called when writing/reading to/from file. 

This means after calling get_output() and then doing Reg.NifitImageData3D(Gadgetron.ImageData) an error occurs.

rijobro:
fixes #527. When setting meta data, make sure to call the `set_up_geom_info`.

@ckolbPTB Could you give this a go and let me know if it works for you? I managed to convert an MR reconstructed image to a `NiftiImageData`. 

(#532 enables `NiftiImageData(sirf.ImageData)` as well as `NiftiImageData3D(sirf.ImageData)`, which we discussed in the Hackathon.)

### DIFF
--- a/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
+++ b/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
@@ -1026,6 +1026,13 @@ GadgetronImageData::set_real_data(const float* z)
 	}
 }
 
+void
+GadgetronImageData::set_meta_data(const AcquisitionsInfo &acqs_info)
+{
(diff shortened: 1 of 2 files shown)


ckolbPTB:
I successfully tested it. I could now create a NiftiImage for data reconstructed with FullySampledReconstructor() using both NiftiImageData and NiftiImageData3D.

PR #535, opened 2020-02-04
stir from nifti

rijobro:
Still in progress. 

@KrisThielemans We discussed calculating the transformation matrix and then converting each SIRF index into a STIR index, along the lines of:

https://github.com/rijobro/SIRF/blob/c8578ba65fb44eb95d324ba40a738d9651878ded/src/xSTIR/cSTIR/stir_data_containers.cpp#L360-L367

I realise that we can't access each element of the SIRFImageData individually, so this method won't work. Could we create a STIRImageData, fill it, and then reorient the image? I suppose my question is if there is reorient functionality in STIR?

### DIFF
--- a/src/Synergistic/tests/test_cSynergistic.cpp
+++ b/src/Synergistic/tests/test_cSynergistic.cpp
@@ -64,25 +64,22 @@ int main(int argc, char* argv[])
         if (argc > 2)
             mr_recon_h5_filename = argv[2];
 
-        // Test STIR -> Nifti
+        // Test Nifti -> STIR -> Nifti
         {
             std::cout << "\nPerforming STIRImageData to NiftiImageData conversion...\n";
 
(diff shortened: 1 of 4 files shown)


KrisThielemans:
@evgueni-ovtchinnikov is it guaranteed that SIRF iterators run through data in the same order as https://github.com/rijobro/SIRF/blob/c8578ba65fb44eb95d324ba40a738d9651878ded/src/xSTIR/cSTIR/stir_data_containers.cpp#L362-L364? (at least when it's a 3D volume).

we must have some correspondence between indices and order somewhere. Otherwise the geom-stuff doesn't make any sense.

evgueni-ovtchinnikov:
@KrisThielemans: SIRF's PET image iterators are wrappers for your `DiscretisedDensity<3, float>::full_iterator` (and the `const` one).

PR #537, opened 2020-02-04
GadgetronImageData from SIRFImageData

rijobro:
I've created an `ISMRMRD::ImageHeader` from the `VoxelisedGeometricalInfo3D`, and filled in as many fields as possible (but not yet debugged).

@evgueni-ovtchinnikov  I'm not sure how to convert the `ISMRMRD::ImageHeader` to an `ImageWrap`, could you advise? 

My two current ideas are to do:
1. `ISMRMRD::ImageHeader`->`ISMRMRD::Image<float>`->`ImageWrap`
1. `ISMRMRD::ImageHeader`-> directly to `ImageWrap`.

Any ideas? You can see the progress I've already made in the PR. 

@ckolbPTB tagging you so that you don't waste any time starting from scratch.


evgueni-ovtchinnikov:
`ImageWrap` is constructed from image type (stored in `ISMRMRD::ImageHeader` property `data_type`) and void pointer to `ISMRMRD::Image`, so go the first root.

evgueni-ovtchinnikov:
Forgot to add: `ImageWrap` takes the ownership of the image whose pointer it receives, so it must be image just created by `new`.

paskino:
Suggest to archive

PR #544, opened 2020-02-08
added pass through of threshold for recon

ALEXJAZZ008008:


KrisThielemans:
@ALEXJAZZ008008 could you give this a quick look such that we can merge it  in before 2.2? Should only take 5 mins to add in the doc. I don't see why this PR should be updating the `data` submodule though. As that's rather fragile, I'd vastly prefer if it doesn't.

ALEXJAZZ008008:
Yes, I'll look at this first thing tomorrow.
I agree that there shouldn't be changes to data I believe this will be because I used "got add ."

PR #553, opened 2020-02-19
wrap niftypet projector to python

rijobro:


rijobro:
I've tried wrapping to Matlab and the same C-function succeeds for Python, but not for Matlab. The function in question is `cSTIR_setupAcquisitionModel`. That is to say, the following works in python:

```
acq_model = AcquisitionModelUsingNiftyPET()
acq_model.set_cuda_verbosity(verb)
acq_model.set_up(template_acq_data, image)
```

But the following crashes in Matlab:
```
acq_model = sirf.STIR.AcquisitionModelUsingNiftyPET();
acq_model.set_cuda_verbosity(true);
acq_model.set_up(template_acq_data, image);
```
The matlab error is as follows:
```
9:     identifier: 'AcqModUsingNiftyPET:set_up:error'
9:        message: '??? std::bad_alloc exception thrown at line 463 of /home/rich/Documents/Code/SIRF-SuperBuild/Build/sources/SIRF/src/xSTIR/cSTIR/cstir.cpp, ...'
```
Strange that it works in python but not Matlab. @evgueni-ovtchinnikov any ideas?

KrisThielemans:
Seems you're running out of memory

rijobro:
change this PR to only do python. Matlab wrapper will be added at a later date. PR #554.

PR #557, opened 2020-02-21
added PETAcquisitionModel.set_image_data_processor

### LINKED ISSUE #556: expose image-based PSF modelling in STIR
STIR can do basic resolution modelling by first smoothing the image and then projecting it. It's done via the "Pre Data Processor" for forward projection, and "post" for back projection.

To expose this to SIRF, I suggest the following interface in `PETAcquisitionModel`:
```python
   set_image_data_processor(...)
```
which would take an `ImageProcessor`. We can then write some doc that the `SeparableGaussianImageProcessor` is the one to use for PSF modelling.

I think this can then be done on the C++ side by adding in [PETAcquisitionModel](https://github.com/CCPPETMR/SIRF/blob/01924a7f23aaa5a47c02c2a35154705410ae9fb4/src/xSTIR/cSTIR/include/sirf/STIR/stir_x.h#L315)
```C++
   set_image_processor(const ImageProcessor&)
```
This would then  use [`ForwardProjectorByBin::set_pre_data_processor`](https://github.com/UCL/STIR/blob/5baf1ab1506780630c364cb49827f09084189688/src/include/stir/recon_buildblock/ForwardProjectorByBin.h#L134) and similar for the backprojector. It would have to check that the projector_pair is already initialised, otherwise throw.

Strictly speaking this is only correct for those image-processors where the `forward`=`adjoint`. In STIR, we currently have only `forward` though. We could let the user add a pair if necessary. let's do that later.
Will have to be listed in CHANGES.md and UserGuide.md

KrisThielemans:
allows PSF modelling.

Fixes #556


KrisThielemans:
Warning: this needs https://github.com/UCL/STIR/pull/454 to be merged before the SeparableGaussianFilter will do anything at all.

KrisThielemans:
https://github.com/UCL/STIR/pull/454 is now merged to STIR master.

this PR works without that (but it actually wouldn't change any results)

PR #558, opened 2020-02-21
Resampling complex images in python and c++

rijobro:
if complex image is given, internally split it up, resample the two real and imaginary components and then stick it back together.


src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -183,6 +192,40 @@ std::shared_ptr<nifti_image> NiftiImageData<dataType>::create_from_geom_info(con
     return _nifti_image;
 }
 
+template<class dataType>
+void NiftiImageData<dataType>::construct_NiftiImageData_from_complex_im_real_component(std::shared_ptr<NiftiImageData> &out_sptr, const std::shared_ptr<const ImageData> in_sptr)
+{
+    // Create image from input
+    out_sptr = std::make_shared<NiftiImageData<dataType> >(*in_sptr);
```
KrisThielemans: slightly confused. Isn't this  copying the data already? It's calling the constructor... So what would it do if the input data is complex? (And if it isn't, then just calling the copy constructor would be fine)

src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -183,6 +192,40 @@ std::shared_ptr<nifti_image> NiftiImageData<dataType>::create_from_geom_info(con
...
+
+    ImageData::Iterator_const &it_in = in_sptr->begin();
+    typename NiftiImageData<dataType>::Iterator it_out = out_sptr->begin();
```
KrisThielemans: could use `auto` here I guess

src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -183,6 +192,40 @@ std::shared_ptr<nifti_image> NiftiImageData<dataType>::create_from_geom_info(con
...
+}
+
+template<class dataType>
+void NiftiImageData<dataType>::construct_NiftiImageData_from_complex_im_imag_component(std::shared_ptr<NiftiImageData> &out_sptr, const std::shared_ptr<const ImageData> in_sptr)
+{
+    if (!in_sptr->is_complex())
+        std::cout << "\nNiftiImageData<dataType>::construct_NiftiImageData_from_complex_im. Warning, input image is not complex. Complex component will be empty\n";
+
+    // Create image from input
+    out_sptr = std::make_shared<NiftiImageData<dataType> >(*in_sptr);
```
KrisThielemans: as above

src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -183,6 +192,40 @@ std::shared_ptr<nifti_image> NiftiImageData<dataType>::create_from_geom_info(con
...
+
+    ImageData::Iterator_const &it_in = in_sptr->begin();
+    typename NiftiImageData<dataType>::Iterator it_out = out_sptr->begin();
+    for (; it_in!=in_sptr->end(); ++it_in, ++it_out)
+        *it_out = (*it_in).complex_float().imag();
```
KrisThielemans: will this work if `*t_in` is real?

src/Registration/cReg/include/sirf/Reg/NiftyResample.h
```diff
@@ -42,6 +42,43 @@ class BSplineTransformation;
 
 namespace sirf {
 
+template<class dataType>
+class ComplexNiftiImageData
```
KrisThielemans: do you consider as a temporary work-around? If so, I'd put it in a `detail` `namespace`. If not, I'd put it in its own file and add a bit of doxygen (probably useful to make its intent clear anyway)

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_image_wrap.h
```diff
@@ -520,6 +538,12 @@ namespace sirf {
 			dim[2] = im.getMatrixSizeZ();
 			dim[3] = im.getNumberOfChannels();
 		}
+        template<typename T>
+		void get_data_type_(const ISMRMRD::Image<T>* ptr_im, ISMRMRD::ISMRMRD_DataTypes* data_type_ptr) const
+		{
+			const ISMRMRD::Image<T>& im = *static_cast<const ISMRMRD::Image<T>*>(ptr_im);
```
KrisThielemans: very complicated way to write the cast... In fact, it looks like there's no cast at all...

KrisThielemans:
Also, can you try to make your commit messages more complete? They might make some sense in the context of the PR, but once merged, many of them will be very puzzling.

KrisThielemans:
Travis failing with
```
Cloning into 'SIRF'...
fatal: reference is not a tree: 1c0ee36f48e0e3b7a5f250483216be01f80feebe
CMake Error at /home/travis/build/CCPPETMR/SIRF-SuperBuild/builds/SIRF/tmp/SIRF-gitclone.cmake:75 (message):
  Failed to checkout tag: '1c0ee36f48e0e3b7a5f250483216be01f80feebe'
```
I guess this is the usual "cannot run Travis for a PR from a fork". Do we have an Issue for that (couldn't find it).

src/Registration/cReg/NiftiImageData.cpp
rijobro: I create a clone to get the shape and metadata etc. Then correct the voxel values. I think it takes the constructor takes the magnitude, so then they get replaced with the real component.

src/Registration/cReg/NiftiImageData.cpp
rijobro: Yes, think the num ref class is set up to be able to return a complex number from a real one. Just the imaginary component will be `0i`.

src/Registration/cReg/include/sirf/Reg/NiftyResample.h
rijobro: I don't consider it a temporary workaround, but I only see it being used as a wrapper in the resampler, so didn't really want to expose it to the rest of SIRF. What would you prefer?

rijobro:
Sure, for the commit messages I was planning to squash to tidy them up anyway

src/Registration/cReg/NiftiImageData.cpp
KrisThielemans: ok. seems wasteful, but let's just add a TODO. (Unless we happen to have a `get_uniform_copy` already (which we do for `STIRImageData`)

src/Registration/cReg/NiftiImageData.cpp
KrisThielemans: ok

src/Registration/cReg/include/sirf/Reg/NiftyResample.h
KrisThielemans: ok. if not to be exposed, let's put it in its own namespace. See [here](https://github.com/UCL/STIR/blob/e2f0bcb74af6bbf067785e07f58a8430faf3c64d/src/include/stir/IO/read_data_1d.h#L36) for an example

rijobro: done

PR #574, opened 2020-03-11
change CListModeData to ListModeData

rijobro:
fixes https://github.com/CCPPETMR/SIRF/issues/571

### DIFF
--- a/src/xSTIR/cSTIR/CMakeLists.txt
+++ b/src/xSTIR/cSTIR/CMakeLists.txt
@@ -27,6 +27,13 @@ else()
   remove_definitions(-DUSE_HKEM)
 endif()
 
+# Check for existence of ListModeData.h
+if (EXISTS "${STIR_INCLUDE_DIRS}/stir/listmode/ListModeData.h")
+  add_definitions(-DSTIR_USE_LISTMODEDATA)
+else()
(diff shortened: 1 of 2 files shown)


KrisThielemans:
The problem with this approach is that it will break compilation with older STIR (as usual). I was going to try a simple work-around first, and do this change later.

I'll send the PR some time this  afternoon.

rijobro:
ok. if STIR version is available in C++, we could potentially do an if/else `#define`?

rijobro:
or do you mean fix it on the STIR side? I guess that would be preferable if possible.

KrisThielemans:
STIR is still on 4.0.0-alpha (sorry). We have no way to check the version therefore.

We could put in a horrible check if `ListModeData.h` exists (which I  think you've done before), but I'm now inclining to just accept this PR, but also update `version_config.cmake` for default STIR version.

rijobro:
As per your suggestion, check for existence of `ListModeData.h` and create compiler definition accordingly. Use a typedef in `stir_x.cpp` to use `CListModeData` or `ListModeData` as necessary.

Think this is better than updating `version_config.cmake` as it's a little more forgiving in terms of backwards compatibility.

Issue #593, opened 2020-03-24
Memory leak in PET_MCIR.py


rijobro:
There is/are error leak(s) in [examples/Python/PETMR/PET_MCIR.py](https://github.com/CCPPETMR/SIRF/blob/master/examples/Python/PETMR/PET_MCIR.py).

I ran it through valgrind like this:
```
valgrind --tool=memcheck --leak-check=full --show-leak-kinds=all \
    python PET_MCIR.py \
        --trans "mfs_i_nii/mf_i_g1.nii" \
        --sinos "EM_g1.hs" \
        --outp mcir_output/recon 
        --trans_type-t disp \
        --iter 1 \
        --save_interval 1 \
        --verbosity 10
        --nifti
```
The valgrind output is attached, I can't make head or tail of it.

I also tried [memory profiler](https://pypi.org/project/memory-profiler/) and the output is more concise. Their example shows memory increasing during allocation, and then decreasing when objects go out of scope/destructors are called. My output doesn't show any decrease in memory. I wonder if it means that **everything** is leaking memory (please don't be this!) or that the profiler isn't smart enough to notice objects are getting deleted on the C-level. Output also attached.

Any ideas? I suppose one thing I need to do is break the code into small chunks and verify each bit, but this will take a long time...

## Log files
[valgrind_output.txt](https://github.com/CCPPETMR/SIRF/files/4375920/valgrind_output.txt)
[memprofiler_output.txt](https://github.com/CCPPETMR/SIRF/files/4375967/memprofiler_output.txt)

rijobro:
Also, how should I interpret the end of the valgrind output, with 20GB still reachable?
```
==10926== LEAK SUMMARY:
==10926==    definitely lost: 1,608 bytes in 12 blocks
==10926==    indirectly lost: 16,460 bytes in 41 blocks
==10926==      possibly lost: 895,780 bytes in 1,615 blocks
==10926==    still reachable: 20,810,150 bytes in 24,290 blocks
==10926==                       of which reachable via heuristic:
==10926==                         stdstring          : 5,374 bytes in 107 blocks
==10926==                         multipleinheritance: 4,448 bytes in 5 blocks
==10926==         suppressed: 0 bytes in 0 blocks
```

rijobro:
Lastly, it's also possible that this is a CIL issue as PET_MCIR.py uses both SIRF and CIL. @paskino

rijobro:
[...]

PR #594, opened 2020-03-25
fix memory leak in set_verbosity

rijobro:
As discussed in https://github.com/CCPPETMR/SIRF/issues/593


rijobro:
testing the following with valgrind before and after this PR shows that it solves the memory leak noted in #593.
```
import sirf.STIR
sirf.STIR.set_verbosity(0)
```

PR #596, opened 2020-03-26
use sptr of ProjDataInfo instead of ptr

rijobro:
Fixes related part of memory leak in https://github.com/CCPPETMR/SIRF/issues/593. 

I think this PR should be independent of https://github.com/UCL/STIR/pull/470 (corresponding issue: https://github.com/UCL/STIR/issues/466), since `get_proj_data_info_sptr` already existed on the STIR side.

### DIFF
--- a/src/xSTIR/cSTIR/stir_x.cpp
+++ b/src/xSTIR/cSTIR/stir_x.cpp
@@ -320,34 +320,34 @@ ListmodeToSinograms::estimate_randoms_()
 	ProjData& proj_data = *randoms_sptr->data();
 
 	const int num_rings =
-		template_projdata_ptr->get_proj_data_info_ptr()->get_scanner_ptr()->
+		template_projdata_ptr->get_proj_data_info_sptr()->get_scanner_ptr()->
 		get_num_rings();
 	const int num_detectors_per_ring =
-		template_projdata_ptr->get_proj_data_info_ptr()->get_scanner_ptr()->
+		template_projdata_ptr->get_proj_data_info_sptr()->get_scanner_ptr()->
 		get_num_detectors_per_ring();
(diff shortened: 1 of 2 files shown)


KrisThielemans:
This doesn't fix any leaks  as far as I can see. Local use of the bare pointer is ok. We never created a new object as far as I can see. But it doesn't do any harm, so fine.

src/xSTIR/cSTIR/stir_x.cpp
```diff
@@ -320,34 +320,34 @@ ListmodeToSinograms::estimate_randoms_()
...
-			scanner_sptr(new Scanner(*proj_data_info_ptr->get_scanner_ptr()));
+			scanner_sptr(new Scanner(*proj_data_info_sptr->get_scanner_ptr()));
 		unique_ptr<ProjDataInfo> uncompressed_proj_data_info_uptr
 			(ProjDataInfo::construct_proj_data_info(scanner_sptr,
 				/*span=*/1, max_ring_diff_for_fansums,
 				/*num_views=*/num_detectors_per_ring / 2,
 				scanner_sptr->get_max_num_non_arccorrected_bins(),
 				/*arccorrection=*/false));
-		const ProjDataInfoCylindricalNoArcCorr * const uncompressed_proj_data_info_ptr =
+		const ProjDataInfoCylindricalNoArcCorr * const uncompressed_proj_data_info_sptr =
```
KrisThielemans: This is the wrong name though. It isn't a `shared_ptr`. Introducing variables with the wrong name doesn't seem a great idea.

Can we revert this? Or change it to using a reference (probably clearest).

rijobro: Sorry, probably a case of "replace all" gone wrong. Ok, i'll change to ref and put in new PR.

Issue #526, opened 2020-01-31
add tests between different SIRFImageData types


KrisThielemans:
- add `VoxelisedGeometryInfo::operator==` (use accuracy as in https://github.com/CCPPETMR/SIRF/blob/12d2a9dc36f63efc854e1ac5bb293e3ee6e15a1c/src/Registration/cReg/NiftiImageData.cpp#L837)
- write `!=`  in terms of `==`
- add `SIRFImageData::operator==` (and `!=`)
   1. compare `voxelised_geometry_info`
   2. check `norm(diff)/norm(arg1)< 1.E-3`
- add ctest in `cSynergistic`: 
   - test is: 
   ```
   im1=read filename with engine1
   im2= otherengine.ImageData(im1)
   im1==im2;
   ```
   - to do this, copy lots of lines from https://github.com/CCPPETMR/SIRF/blob/master/src/Synergistic/sirf_convert_image_type.cpp and a bit of https://github.com/CCPPETMR/SIRF/blob/master/src/Synergistic/tests/test_cSynergistic.cpp (later, use #525)
   

- expose to Matlab/Python and add tests there. Need to add test on norm of `as_array` as well.

evgueni-ovtchinnikov:
@rijobro: So far as I can see `otherengine` can only be `Reg` at present - or is there a branch that has other conversions implemented?

rijobro:
yes unfortunately we can only convert to `NiftiImageData`, not the other way round. 

PR for ImageData->Gadgetron https://github.com/CCPPETMR/SIRF/pull/537, 
PR for ImageData->STIR https://github.com/CCPPETMR/SIRF/pull/535.

KrisThielemans:
I guess the point is to have the test in master, then merge it into the PRs above, and extend it to the other cases.

PR #600, opened 2020-03-27
ImageData objects conversion tests, addresses #526

### LINKED ISSUE #584: Operator == for ImageData conflicts with the one for NiftiImageData
@rijobro: I have implemented `operator==` for `ImageData`, but run into a conflict with the one in `NiftiImageData`: executing
```
nifti_image = NiftiImageData3D(image)
print(image == nifti_image)
``` 
where `image` is an `ImageData`, raises AssertionError in `NiftiImageData.__eq__` (which is bewildering as `ImageData.__eq__` should have been called).

I added `or isinstance(other, SIRF.ImageData)` in `NiftiImageData.__eq__`, but I got
```
'NiftiImageData<dataType>::are_equal_to_given_accuracy: Image 2 not initialised.' exception caught at line 360 of /home/sirfuser/devel/buildVM/sources/SIRF/src/Registration/cReg/cReg.cpp
```

If I comment out `NiftiImageData.__eq__` (which should be ok because I have `__eq__` in the base class), then the above problem disappears, and my Python ctests run ok, but `SYN_TEST_PYTHON` fails:
```
2:       Start 10: SYN_TEST_PYTHON
2: 10/10 Test #10: SYN_TEST_PYTHON .........................***Failed    1.89 sec
2: 
2: # --------------------------------------------------------------------------------- #
2: #                             Starting STIR to Nifti test...
2: # --------------------------------------------------------------------------------- #
2: 
2: WARNING: Unable to determine patient position. Internally this will generally be handled by assuming HFS
2: 
2: Saving image to file (results/stir_to_nifti.nii)...F
2: ======================================================================
2: FAIL: test_pSynergistic.test
2: ----------------------------------------------------------------------
2: Traceback (most recent call last):
2:   File "/usr/local/lib/python2.7/dist-packages/nose/case.py", line 197, in runTest
2:     self.test(*self.arg)
2:   File "/home/sirfuser/devel/buildVM/sources/SIRF/src/Synergistic/tests/test_pSynergistic.py", line 209, in test
2:     try_stirtonifti(nifti_filename)
2:   File "/home/sirfuser/devel/buildVM/sources/SIRF/src/Synergistic/tests/test_pSynergistic.py", line 50, in try_stirtonifti
2:     raise AssertionError("Conversion from STIR to Nifti failed.")
2: AssertionError: Conversion from STIR to Nifti failed.
```

evgueni-ovtchinnikov:
Adds simple `ImageData` objects conversion tests (currently to `NiftiImageData` only, further work to be continued on master).

Also, quick-fixes #584 and #587.


src/common/csirf.cpp
```diff
@@ -25,6 +25,7 @@ limitations under the License.
 #include "sirf/iUtilities/DataHandle.h"
 #include "sirf/common/DataContainer.h"
 #include "sirf/common/ImageData.h"
+#include "../../Synergistic/cSyn/include/sirf/Syn/utilities.h"
```
KrisThielemans: This shouldn't be done as it is.

evgueni-ovtchinnikov: Had to do it this way because the build failed on Windows when I used the correct one. Magically, now builds ok! Could be something wrong with some CMakeLists actually.

rijobro: If this is a new folder (I think `Syn` didn't exist before), then you need to do `target_link_directories` in CMake.

rijobro: Apologies, I see that you've actually done that.

src/common/SIRF.py
```diff
@@ -55,7 +55,7 @@ def __del__(self):
         print("SIRF.DataContainer __del__ with handle {}.".format(self.handle))
         if self.handle is not None:
             pyiutil.deleteDataHandle(self.handle)
-    @abc.abstractmethod
+#    @abc.abstractmethod
```
rijobro: Why don't we want this to be abstract anymore?

evgueni-ovtchinnikov: `tests_syn.py` fails if this method is abstract

src/common/csirf.cpp
evgueni-ovtchinnikov: Actually, I was wrong about 'now builds ok' - ok on Windows and VM, but some Travis builds fail with
```
fatal error: sirf/Syn/utilities.h: No such file or directory
```

evgueni-ovtchinnikov:
Codacy went bonkers, complaining about trailing whitespace in a non-existent line in Gadgetron.py, so I am merging.

PR #603, opened 2020-03-27
matlab: fix libstd++ version differences on linux

rijobro:
version of libstdc++ used to compile .mex files and Matlab's version libstdc++ aren't necessarily the same, causing problems at runtime. Simple answer is to link statically against libstdc++

### DIFF
--- a/src/Registration/CMakeLists.txt
+++ b/src/Registration/CMakeLists.txt
@@ -23,6 +23,8 @@ if (BUILD_MATLAB)
 		list(APPEND Matlab_LIBRARIES 
 			"${Matlab_ROOT_DIR}/extern/bin/glnxa64/libMatlabEngine.so" 
 			"${Matlab_ROOT_DIR}/extern/bin/glnxa64/libMatlabDataArray.so")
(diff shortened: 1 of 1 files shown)


KrisThielemans:
I don't think this is a great idea. There's various incompatibilities lurking. Quoting from https://stackoverflow.com/a/14082540/3467846

> The reason it might not work if you dynamically load a library (using dlopen) is that libstdc++ symbols it depends on might not have been needed by your application when you (statically) linked it, so those symbols will not be present in your executable.

I'm not sure I understand this, but it seems to say that you could be breaking the Matlab part. Also, any "system" libraries we link with could still be using the shared library, so I believe a conflict will appear at some point.

There's also the problem that not all compilers will necessarily support that flag (gcc,clang and not too old intel C++ compilers do).

I believe that the only safe option is to compile SIRF with a Matlab supported compiler, as mentioned on https://github.com/CCPPETMR/SIRF/wiki/SIRF-and-MATLAB

I veto this PR therefore. Reopen if you have good reasons :-)

PR #604, opened 2020-03-27
HDF5 matlab fix

rijobro:
By linking against `ISMRMRD::ISMRMRD`, we also link properly against HDF5, avoiding mismatches with Matlab's HDF5 version.

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -29,8 +29,6 @@ option(DISABLE_Gadgetron "Disable building the SIRF interface to Gadgetron" OFF)
 if (DISABLE_Gadgetron)
   message(STATUS "Gadgetron support disabled.")
 else()
-  # include explicit search for HDF5 for older ISMRMRD versions.
-  find_package(HDF5 1.8 COMPONENTS C REQUIRED)
   find_package(ISMRMRD REQUIRED)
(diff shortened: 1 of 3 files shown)


src/CMakeLists.txt
```diff
@@ -29,8 +29,6 @@ option(DISABLE_Gadgetron "Disable building the SIRF interface to Gadgetron" OFF)
 if (DISABLE_Gadgetron)
   message(STATUS "Gadgetron support disabled.")
 else()
-  # include explicit search for HDF5 for older ISMRMRD versions.
-  find_package(HDF5 1.8 COMPONENTS C REQUIRED)
   find_package(ISMRMRD REQUIRED)
```
KrisThielemans:
```suggestion
  find_package(ISMRMRD 1.4.1 REQUIRED)
```

KrisThielemans:
Excellent! we were a bit behind on ISMRMRD developments. This target has been available for a while now (https://github.com/ismrmrd/ismrmrd/pull/109 merged in Dec 2018) and we use ISMRMRD 1.4.1 in the SIRF-SuperBuild.

KrisThielemans:
I guess this wouldn't necessarily avoid run-time problems due to shared DLL/SO stuff, but definitely avoids conflicts with include and linking.

PR #606, opened 2020-03-30
in place axpby

rijobro:
As a continuation of #599, this does in-place `axpby` (which also means subtract and addition).

However, many ctests fail. This is because I fundamentally changed the behaviour of `axpby`.

Beforehand:

1. Create an empty copy
2. Append with result of `axpby`

Now:

1. Copy first image
2. Replace values with result of `axpby`.

This breaks the `MRAcquisitionData` implementation, for example, which uses the `append_acquisition` method. I thought about inserting a `clear_acquisitions` method at the beginning of `MRAcquisitionData::axpby`, but this seems tricky for `AcquisitionsFile`. Thoughts @evgueni-ovtchinnikov ?


rijobro:
Ideally, I would have all `ImageData` use the same method, and simply iterate across the whole dataset, something along the lines of
```
ImageData::axpby()
{
    auto &iter = this->begin();
    auto &x_iter = x.begin();
    auto &y_iter = y.begin();
    for ( ; iter != this->end(); ++iter, ++x_iter, ++y_iter)
        *iter = a*(*x_iter) + b*(*y_iter);
}
```

KrisThielemans:
The suggested implementation should indeed be the default. Should even live on the `DataContainer` level. Ideally it would have a check first that meta-data is the same/compatible first. Then specific classes can overload it with faster stuff.

Obviously, now we're starting to enter BLAS territory, but I guess we have to rely on classes having faster implementations.

So, what's stopping us from using this generic strategy?

rijobro:
Can't unfortunately live on the `DataContainer` level, as they currently don't have iterators.

rijobro:
@evgueni-ovtchinnikov I've tried a few combinations, but I can't get it to work. Do you think you could give it a go? Currently failing for Gadgetron images and acquisition data.

evgueni-ovtchinnikov:
@rijobro: OK I will have a look

rijobro:
Thanks!

KrisThielemans:
will be superseded soon

PR #607, opened 2020-03-30
where ProjData is const, need to copy ProjDataInfo

rijobro:


KrisThielemans:
This change would probably cope with https://github.com/UCL/STIR/pull/470 (I guess that's why you did it), but I'm left wondering if there's something quick we can do on the STIR front instead, see https://github.com/UCL/STIR/issues/479

KrisThielemans:
I have the impression that this PR would work with both old (i.e. pre https://github.com/UCL/STIR/pull/470) and new STIR. Is that correct? If so, it's probably a good choice. It might be a bit wasteful in `proj_data_info` clones, but that's going to be a tiny thing anyway.

It also seems smaller than #609 

Would it work with https://github.com/UCL/STIR/pull/480?

rijobro:
Confirmed, this works with both old and new versions of STIR. I'll merge now.

PR #608, opened 2020-03-31
cmake_minimum_required should be first line

paskino:
At a CMake course at Kitware they told that it's good practice to declare the `cmake_minimum_required` version at the top of `CMakeLists.txt`

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -20,6 +20,12 @@
 
 # cmake file for building SIRF. See the SIRF User's Guide and http://www.cmake.org.
 
+cmake_minimum_required(VERSION 3.9.0)
+# require 2.8.3 to get FOLDER properties support
+# require 3.3 for descent FindMatlab.cmake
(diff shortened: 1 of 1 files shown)


paskino:
@KrisThielemans STIR builds all failed on Travis. 
This change however is harmless and I will merge?

KrisThielemans:
This is due to https://github.com/CCPPETMR/SIRF/blob/7ebe03a35a534501cd443278a648a9dbedf32cd9/.travis.yml#L62
forcing STIR to master (an old work-around, see https://github.com/CCPPETMR/SIRF/issues/366 which says "this will create trouble at some point"...).

The failure is because @rijobro's latest STIR PR https://github.com/UCL/STIR/pull/470 was merged, but it breaks SIRF. Seems we forgot to check.

So, we're in the usual dependency mess. Discussion to be continued on #366

Nothing of this has anything to do with this PR, so let's merge this indeed.

PR #609, opened 2020-04-01
projdata const

rijobro:
Use `shared_ptr<const ProjDataInfo>` where possible (as opposed to `shared_ptr<ProjDataInfo>`). Complements https://github.com/UCL/STIR/pull/480.


rijobro:
closing as #607 was preferrable.

PR #617, opened 2020-04-06
Add adjoint tests 

AnderBiguri:
Tests for linear operator's "adjointness"

Tested for STIR (CPU and GPU(WIP))

@paskino, you mentioned mathematicians like to call this differently? I haven't been able to find it with some other name, can you let me know what is the preferred name for this function? 

Requires #628 and #626 

TODO: 

- [x] Test Reg (multiple Registration engines? How many different ones? figure that out)
- [x] Test Gadgetron
- [x] Create ctest if STIR is installed
- [x] Create ctest if Gadgetron is installed
- [x] Create ctest if Registration is installed (yes, this is more than one)
- [x] Create ctest if NiftyPet is installed (?? Not yet ??)


src/common/Utilities.py
```diff
@@ -463,3 +463,29 @@ def str_to_int_list(str_list):
             int_item = list(range(strt, stop + 1))
         int_list = int_list + int_item
     return int_list
+
+def dot_product_test(operator, num_tests = 20, max_err = 10e-5):
+    '''
+    Test if a given operator is adjoint.
+    The operator needs to have been already set_up() with valid AcquisitionModel and ImageData objects.
```
KrisThielemans, line 470:
```suggestion
    The operator needs to have been already set_up() with valid objects.
```
I generalised this a bit, as the text wouldn't have applied to registration etc

src/common/Utilities.py
```diff
@@ -463,3 +463,29 @@ def str_to_int_list(str_list):
             int_item = list(range(strt, stop + 1))
         int_list = int_list + int_item
     return int_list
+
+def dot_product_test(operator, num_tests = 20, max_err = 10e-5):
```
KrisThielemans, line 467: It sounds like it's testing the dot-product... What about `test_operator_adjoint` or `test_adjoint_operation`? Also, needs some doc on `max_err`

src/common/Utilities.py
```diff
@@ -463,3 +463,29 @@ def str_to_int_list(str_list):
...
+    '''
+    Test if a given operator is adjoint.
+    The operator needs to have been already set_up() with valid AcquisitionModel and ImageData objects.
+    The operator needs to have methods direct() and adjoint() implemented
+    '''
+    # Get image and data to know the sizes of each operation.
+    image=operator.domain_geometry()
+    image.fill(1.0)
+    data=operator.direct(image) # Is there a better way than this? Similar to .domain_geomerty() ?
```
KrisThielemans, line 476: currently we don't have one in SIRF I guess. @paskino? You could do this inside the loop to avoid having to do one extra `direct`

src/common/Utilities.py
```diff
@@ -463,3 +463,29 @@ def str_to_int_list(str_list):
...
+    for __ in range(num_tests):
+        # generate random data
+        x0 = numpy.random.rand(*image.shape) * 10**numpy.random.randint(-2, 3)
+        y0 = numpy.random.rand(*data.shape ) * 10**numpy.random.randint(-2, 3)
```
KrisThielemans, line 480: I guess will have to find out if it needs to be complex or not. Can you get the data-type from `image` and `data`

src/common/Utilities.py
```diff
@@ -463,3 +463,29 @@ def str_to_int_list(str_list):
...
+        # fill image and data
+        image.fill(x0)
```
KrisThielemans, line 482: probably better to call this `x`  as you don't know that it'll be an image. same for `y` below

src/common/Utilities.py
AnderBiguri: Sure, this was a temporary name, as @paskino  said he chatted with some mathematicians and they had some opinions on the correct name of the test, something like "linearity test" but I don't 100% recall.

KrisThielemans: certainly not `linearity_test`. An operator can be linear, while we have a lousy `adjoint` implementation (or even none at all)

AnderBiguri: That is literally what I said in that meeting, but I am waiting for Edo to confirm.

src/common/Utilities.py
AnderBiguri: Ah pity, its much cleaner if it is out of the loop.

src/common/Utilities.py
AnderBiguri: I suspect I can check against `numpy` types here, for complex? You don't have your own implementation for complex, right? I also do not see any method/attribute that tells me if its complex or not.

src/common/Utilities.py
KrisThielemans: sure. possibly we have `range_geometry` I sort of thought CIL would need it.

By the way, if the objects are allocated upfront, you might be able to save some mememory by using the `out` parameter

src/common/Utilities.py
```diff
@@ -464,28 +464,41 @@ def str_to_int_list(str_list):
...
+        y0 = numpy.random.rand(*y_hat.shape) * 10**numpy.random.randint(-2, 3)
+        # add a complex part if necessary
+        if numpy.any(numpy.iscomplex(y_hat.as_array())):
+            y0=y0+1j*numpy.random.rand(*y_hat.shape) * 10**numpy.random.randint(-2, 3)
+        y=y_hat.copy()
+        y.fill(y0)
+        x_hat = operator.adjoint(y)
         # Check dot product identity
-        norm_err = abs(y_hat.dot(data) - x_hat.dot(image))/(y_hat.dot(data)*0.5 + x_hat.dot(image)*0.5)
+        norm_err = abs(y_hat.dot(y) - x_hat.dot(x))/(y_hat.dot(y)*0.5 + x_hat.dot(x)*0.5)
```
KrisThielemans, line 501: would need complex conjugate when just using `dot` I guess. See e.g. https://mathworld.wolfram.com/HermitianInnerProduct.html

Also, in the denominator need `abs(y_hat.dot(data))*0.5 + ...`

src/common/Utilities.py
```diff
@@ -464,28 +464,41 @@ def str_to_int_list(str_list):
...
-        # fill image and data
-        image.fill(x0)
-        data.fill(y0)
-        # compute direct and adjoint
-        y_hat = operator.direct(image)
-        x_hat = operator.adjoint(data)
+        ## generate random data and direct()
+        x0 = numpy.random.rand(*x.shape) * 10**numpy.random.randint(-2, 3)
+        # add a complex part if necessary
+        if numpy.any(numpy.iscomplex(x.as_array())):
```
KrisThielemans, line 488: this can presumably never be true, as all its elements are real by construction. I believe you need `numpy.iscomplexobj`

src/common/Utilities.py
AnderBiguri: @KrisThielemans I believe this should be already handled by `AdquisitionData.dot()` and `ImageData.dot()`. I will make sure its like that.

KrisThielemans: might be. If it does, it would probably make more sense to adhere to `numpy` conventions where `dot` is matrix multiplication, but `vdot` uses complex conjugate. https://docs.scipy.org/doc/numpy/reference/generated/numpy.vdot.html#numpy.vdot. Once you know what happens, please create an issue accordingly.

paskino:
@AnderBiguri sorry for late reply, my mathematicians call it `dot_test`.

KrisThielemans:
> my mathematicians call it dot_test.

tell them I disagree then :-) It doesn't test the `dot` operation (i.e. the metric), it tests the operator and its adjoint. I could possibly agree if `dot_test` would be a member of `operator` but it seems rather weird to pollute objects with their test functions.

paskino:
The [equivalent method](https://github.com/vais-ral/CCPi-Framework/blob/master/Wrappers/Python/ccpi/optimisation/operators/Operator.py#L167) in CIL is very similar.

The 2 biggest differences are

1. we hide the random allocation inside the `allocate` method of the `AcquisitionData ImageData` templates (we call them `AcquisitionGeometry ImageGeometry`).
1. we do only one test

The templates can be accessed from the `AcquisitionModel` (`Operator` in CIL) by calling `range_geometry` and `domain_geometry` both in SIRF and CIL. So it may be better to expand the appropriate `allocate` methods for `Gadgetron`, [here ](https://github.com/SyneRBI/SIRF/blob/master/src/xGadgetron/pGadgetron/Gadgetron.py#L621)and [here](https://github.com/SyneRBI/SIRF/blob/master/src/xGadgetron/pGadgetron/Gadgetron.py#L1018), STIR [here ](https://github.com/SyneRBI/SIRF/blob/master/src/xSTIR/pSTIR/STIR.py.in#L376)and [here ](https://github.com/SyneRBI/SIRF/blob/master/src/xSTIR/pSTIR/STIR.py.in#L807)and [Nifty](https://github.com/SyneRBI/SIRF/blob/master/src/Registration/pReg/Reg.py.in#L293).

paskino:
> > my mathematicians call it dot_test.
> 
> tell them I disagree then :-) It doesn't test the `dot` operation (i.e. the metric), it tests the operator and its adjoint. I could possibly agree if `dot_test` would be a member of `operator` but it seems rather weird to pollute objects with their test functions.

As a matter of facts it is a static method of the[ `LinearOperator`](https://github.com/vais-ral/CCPi-Framework/blob/master/Wrappers/Python/ccpi/optimisation/operators/Operator.py#L167) class, so it makes somewhat more sense. Here you are adding a "simple" function which will perform the test without being linked to any specific class.

Regardless of the name, I think it is better placed in CIL.

AnderBiguri:
> As I commented [#617 (comment)](https://github.com/SyneRBI/SIRF/pull/617#issuecomment-615100801) I think there's not much difference with the equivalent CIL method. I believe we should strive to keep the code as similar as possible.

I agree!

> In the comment I propose to move the random allocation to the appropriate `DataContainer` template. as it adds in readability, it uses already available methods and makes the code more similar to CIL.

Right. I think I may need a bit of guidance to do this. This makes sense, to be able to ask for a randomly initialized data container (I am assuming `AcquisitionData` and `ImageData` are data containers). It looks like just `imageData.allocate(value = 'random')` should do the job. I will test this.


> Also in the following comments it comes out how in CIL this test is linked to the `LinearOperator` class and the name is `LinearOperator.dot_test` while here it is a function testing an operator (e.g. `AcquisitionModel`) and `dot_test` is not appropriate.

Yes, so it makes sense for it to be the `dot_test` of a linear operator, or in general the test to know if a operator is adjoint, depends how explicit it is. How should I proceed with this then? Do I leave it as a function? Or you prefer for me to change it to something like CIL?

src/common/Utilities.py
AnderBiguri: @KrisThielemans hum, so according to those conventions, if this works, all of them should be `vdot` in SIRF, right?

KrisThielemans:
> Do I leave it as a function? Or you prefer for me to change it to something like CIL?

we don't have a `LinearOperator` in SIRF I guess. In any case, I feel it belongs in a test module, not in the main class.

AnderBiguri:
> I feel it belongs in a test module, not in the main class.
OK, this makes sense. So where do I put it? Its easy to throw it in a function in `pSIRF/tests/something.py`, but then I would need to repeat it in each of the `repo/tests/something.py`. Or is there a test module for all toolboxes somewhere? I could only find `class pTest(object)` in utilities, but we discussed that it does not really fit there.

src/common/Utilities.py
KrisThielemans: guess so. Please just raise a SIRF issue for the creation of `vdot`, but just use `dot` here (and `numpy` conjugate).

KrisThielemans: @paskino this would also apply to CIL `dot_test`

AnderBiguri: Sure. No need of numpy conjugate I believe. According to @paskino gadgetron.dot() will do it.

KrisThielemans:
> here do I put it? Its easy to throw it in a function in pSIRF/tests/something.py, but then I would need to repeat it in each of the repo/tests/something.py. Or is there a test module for all toolboxes somewhere? I could only find class pTest(object) in Utilities.py, but we discussed that it does not really fit there.

@evgueni-ovtchinnikov @rijobro can you please advise where we should put a Python test function applicable to multiple engines? `common/test` seems to be the best place to me. (It's not a synergistic test, as it uses only a single engine)

evgueni-ovtchinnikov:
agree about `common/test`

src/common/Utilities.py
KrisThielemans: then `gadgetron.dot` has the wrong name...

AnderBiguri: @KrisThielemans  right, all of them should be `vdot`, as if its for numpy convention, all of them are wrong named, i.e. STIR does return matrices, not vectors, yet the `dot` does not compute matrix dot products, but vector dot products, right?

evgueni-ovtchinnikov: but MATLAB uses `dot`, not `vdot`, and we are trying to make our Python and MATLAB scripts look as close as possible (ideally, ; at the end being the only difference).

AnderBiguri: @evgueni-ovtchinnikov hum, but MATLAB's `dot` does not behave like neither  `np.dot` or `np.vdot`. The difference being that for matrices, MATLAB takes vector by vector dot product (vectorized approach to multiple `vdot`s) while python's `dot` does matrix multiplication.

KrisThielemans: @evgueni-ovtchinnikov that's true, but in this case unfortunately we might have to diverge. We have to balance this with the "native" feeling of the wrapper. (another example being the index order).

In any case, MATLAB's `dot`  doesn't do complex conjugate either. https://uk.mathworks.com/help/matlab/ref/dot.html

AnderBiguri: @KrisThielemans that is not correct. 

```
A = [1+i 1-i -1+i -1-i];
B = [3-4i 6-2i 1+2i 4+3i];

sum(conj(A).*B)
ans =1.0000 - 5.0000i

sum(A.*B)
ans = 7.0000 -17.0000i

dot(A,B)
ans =1.0000 - 5.0000i
```

KrisThielemans: sorry for giving the wrong info. trying to be too fast!

so, we're stuck in naming now anyway. Seems that MATLAB's `dot` is neither like Python's `vdot` (it will give an n-1 dim array), nor `dot` (it does complex conjugate). sigh.

We might just have to give up and invent our own `scalar_product` function.

Let's see what @paskino thinks.

AnderBiguri: My 2 cents: The behavior is closer to `vdot` than `dot` if we want to be closer to numpy. As this is python, I think it makes more sense to be closer to python implementations than MATLAB. In fact, what SIRF is doing in all cases is the same behavior as `vdot`: unroll, and complex conjugate if needed.

src/common/Utilities.py
```diff
@@ -463,3 +463,38 @@ def str_to_int_list(str_list):
...
+            x_hat = operator.adjoint(y)
+            # Check dot product identity
+            norm_err = abs(y_hat.dot(y) - x_hat.dot(x))/(abs(y_hat.dot(y))*0.5 + abs(x_hat.dot(x))*0.5)            
+            if norm_err > max_err:
+                if verbose:
+                    print(type(operator).__name__ + " is not adjoint, with normalized error of " + str(norm_err))
+                return False
+            elif verbose:
+                print("Testing " + type(operator).__name__ + ": Iteration " + str(iter1*num_tests+iter2+1) + "/" + str(num_tests**2))
+    return True
```
rijobro, line 500: should just be `return norm_err > max_err` and remove `return False` and `return True`.

rijobro: Also the verbose output for successful is a little misleading. To me, `Testing blah blah` implies that it's about to test that thing. More useful would be something along the lines of `Successfully completed iteration 1/25, with a norm of...`.

AnderBiguri: But then I need to remove the `elif verbose`. Or you meant replace the False and the True by that condition?

rijobro: > But then I need to remove the elif verbose. Or you meant replace the False and the True by that condition?

Ah, sorry I understand now. You want to return False as soon as there's an error, else you keep going all the way to the end and then return `True`. Looks fine as is!

AnderBiguri: @rijobro point taken with the verbose! I will work on that to print more useful info. This is still all WIP, so please keep bringing those suggestions ;)

AnderBiguri:
@KrisThielemans @rijobro @evgueni-ovtchinnikov SPM does not have `direct` or `adjoint` stuff. Seems to be about registration. Should it have it, thus, should it be implemented and tested, or not?

rijobro:
I've only wrapped the registration component of SPM, not the resampler. So no need to test :)

src/common/Utilities.py
AnderBiguri: I am marking this conversation as resolved, as it has drifted to `vdot` vs `dot`. Better to open a new issue on the topic.

AnderBiguri:
My tests show that Gadgetron is not adjoint. Is this expected?

KrisThielemans:
> My tests show that Gadgetron is not adjoint. Is this expected?

no, certainly not. I believe @evgueni-ovtchinnikov  has a (basic) test on adjoint as well somewhere (probably with a single data point).

AnderBiguri:
Hum I see, its on `pGadgetron/test2.py` but does some `recon` and `sort` operations on the data before making the tests, which my function does not. Not sure if this is because there is something non-uniform in SIRF, or I am missing/misunderstanding something.

evgueni-ovtchinnikov:
@AnderBiguri: `test2.py` does reconstruction just to create an image (our MR data in `data` contains acquisition data only, no MR images there), and acquisitions must be sorted to compute coil sensitivity maps needed for forward/backprojections

evgueni-ovtchinnikov:
Following BLAS, SIRF's `dot` conjugates the first argument, and in your test numpy conjugates the second.

AnderBiguri:
@evgueni-ovtchinnikov that was not wrong, and the complex dot product requires the second input conjugate. But regarthless, the dot equality of the adjoint tests requires an extra complex conjugate to be actually equal.  Fixing now.

AnderBiguri:
@paskino  @evgueni-ovtchinnikov @KrisThielemans 

All test running and working. What is missing is:

1.- Review final changes before I squash.
2.- I still don't get where you want the test. You want me to make a new module in `common/test`?

KrisThielemans:
> 1.- Review final changes before I squash.

will need to be rebased/squashed after #628 is merged, as this is in here as well.

> 2.- I still don't get where you want the test. You want me to make a new module in common/test?

In the interest of moving on, let's leave your new `is_operator_adjoint` in `utilities.py` now. The rest is where it should be I guess.

Given #629 should this code not be using `get_inner_product`?

AnderBiguri:
@KrisThielemans The current SIRF `dot` does work well with complex data.

Before, when I complained about Gadgetron not being adjoint problem, SIRF was OK, my test was not (the `dot` equality is not strictly valid for complex operators, you need the complex conjugate of one of the results of the dot products). The existing test was only checking the real part of the result, thus passing. Fair but not complete. 

I only found `get_inner_product` because it is what @rijobro was using to test NiftyReg, as `dot` did not work for that data (#627) but I fixed that in #628. I just oppened #629 to make you aware of this, but I have avoided using it, as I do not know how extended over SIRF that function is.  


PD: just to clarify, when I say `dot` in all these cases, I mean proper Hermitian product/inner product for vectors. The correct one for complex data.

PR #621, opened 2020-04-16
make compatible with cmake 3.10

AnderBiguri:
`add_compile_definitions` is 3.12. This is the same functionality with 3.10 syntax.


AnderBiguri:
This could (should?) also be made with `target_compile_definitions(library_name PRIVATE SOME_BOOL_VARIABLE)` but neither STIR, xSTIR or SIRF seem to work as a substitute to `library_name`. I am happy to try suggestions.

PR #626, opened 2020-04-19
resolve #625, swap domain/range geometry niftyreg

### LINKED ISSUE #625: NiftyResample uses domain/range_geometry swapped
For STIR/NiftyPET:

```
    # domain gives you what you need as input for direct() 
    x = acq_model.domain_geometry()
    x.allocate(value = 'random')
    y_hat = acq_model.direct(x)

    # range gives you what you need for adjoint()
    y = acq_model.range_geometry()
    y.allocate( value = 'random')
    x_hat = acq_model.adjoint(y)   # This fails 
```
But for NiftyResample:

```
    # RANGE!!!! gives you what you need as input for direct() 
    x = nr.range_geometry()
    x.allocate(value = 'random')
    y_hat = nr.direct(x)

    # DOMAIN!!! gives you what you need for adjoint()
    y = nr.domain_geometry()
    y.allocate( value = 'random')
    x_hat = nr.adjoint(y)   # This fails 
```

AnderBiguri:


rijobro:
I have no idea which way round these two should be. Leaving it up to you and @paskino (think you implemented it, right?)

AnderBiguri:
@rijobro well, the other way around errors, so I assume this is the correct way.

PR #630, opened 2020-04-21
matlab STIR save to nifti and test (python too)

rijobro:


rijobro:
I've removed the testing, since this caused failure when STIR wasn't built with ITK.

PR #635, opened 2020-04-24
Proper in-place algebra

evgueni-ovtchinnikov:
In-place data containers algebra (`+=`, `-=`, `*=` and `/=`) was temporarily implemented in SIRF via numpy in-place algebra and `as_array`/`fill`.

This PR replaces the temproary implementation with a proper one that eliminates unnecessary copying of data between SIRF data containers and numpy arrays.

### DIFF
--- a/examples/Python/PET/acquisition_data.py
+++ b/examples/Python/PET/acquisition_data.py
@@ -106,7 +106,10 @@ def main():
     acq_factor = acq_data.get_uniform_copy(0.1)
     new_acq_data = acq_data / acq_factor
     print('norm of acq_data*10: %f' % new_acq_data.norm())
-    acq_copy = acq_data.get_uniform_copy()
+    acq_copy = acq_data.get_uniform_copy(1.0)
+    acq_copy *= acq_data
+    diff = acq_copy - acq_data
+    print('norm of acq_copy - acq_data: %f' % diff.norm())
     acq_copy.fill(acq_data)
     diff = acq_copy - acq_data
     print('norm of acq_copy - acq_data: %f' % diff.norm())
@@ -130,6 +133,9 @@ def main():
     diff -= image
     print('norm of image.clone() - image: %f' % diff.norm())
     image_copy = image.get_uniform_copy()
+    image_copy *= image
+    diff = image_copy - image
+    print('norm of image_copy - image: %f' % diff.norm())
     image_copy.fill(image)
     diff = image_copy - image
     print('norm of image_copy - image: %f' % diff.norm())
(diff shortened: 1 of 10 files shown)


examples/Python/PET/acquisition_data.py
```diff
@@ -100,12 +100,16 @@ def main():
     print('norm of acq_data.as_array(): %f' % numpy.linalg.norm(acq_array))
     print('acq_data.norm(): %f' % s)
     print('sqrt(acq_data.dot(acq_data)): %f' % math.sqrt(t))
-    diff = new_acq_data - acq_data
+    diff = new_acq_data
+    diff -= acq_data
```
KrisThielemans, line 104: this update looks a non-intuitive way to write the code. why would we do this?

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
```diff
@@ -378,11 +378,13 @@ complex_float_t a, complex_float_t b)
 	int n = y.number();
 	ISMRMRD::Acquisition ax;
 	ISMRMRD::Acquisition ay;
+	ISMRMRD::Acquisition acq;
```
KrisThielemans, line 381: updates here don't seem have anything to do with the title of this PR. Are these intentional?

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -222,6 +222,7 @@ namespace sirf {
 		// abstract methods
 
 		virtual void empty() = 0;
+		virtual void take_over(MRAcquisitionData&) = 0;
```
KrisThielemans, line 225: not related to PR?

examples/Python/PET/acquisition_data.py
evgueni-ovtchinnikov: just was checking `-=`

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
evgueni-ovtchinnikov: all updates are necessary for in-place algebra the way I implemented it: e.g. now that we want to overwrite acquisition data, we need to check first whether the acquisition to be overwritten is not to be ignored

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
evgueni-ovtchinnikov: directly related - actually, the main difference from the previous approach is that now new data is created before the old one is dismissed, after which the object takes the ownership of the new data and discards the old one

Issue #555, opened 2020-02-20
matlab catch-up


KrisThielemans:
make a list of things that need to be done

- [ ] `SeparableGaussianImageProcessor`
- [ ] `set_image_data_processor` for STIR (#556)

rijobro:
- [ ]  matlab niftypet projector

KrisThielemans:
- [ ] min/max threshold (#643)

KrisThielemans:
- [ ] doc for `GeometricalInfo` currently says https://github.com/SyneRBI/SIRF/blob/75364cb9367da94dabca8f79cbb976fcd5eb1d3a/src/common/%2Bsirf/%2BSIRF/GeometricalInfo.m#L2

paskino:
In Python STIR.py and in `AcquisitionModel` we updated the default signature of `forward` and `backward` https://github.com/SyneRBI/SIRF/pull/821#issuecomment-756064065

 - [ ] add instance member `num_subsets` and `subset_num` to `AcquisitionModel` in MATLAB
 - [ ] use these as default values for `backward` and `forward`

KrisThielemans:
we've stopped supporting MATLAB.

PR #646, opened 2020-04-29
Coildata Cleanup

johannesmayer:
The aim of this PR is to get rid of all the excess classes involved in computing coil sensitivity maps (CSMs). Most of these classes were some form of data containers for CSM data and the images on whose basis the CSMs are calculated. However, since CSMs are nothing but MR images and carry the exact same header information this code is not necessary any more since we have MR image classes readily available and can store the CSMs in them.

New classes
-  `CoilSensitivitiesVector` inherited from `GadgetronImagesVector`
    - it carries two public functions `calculate(MRAcquisitionData&)` and `calculate(GadgetronImagesVector)` that allow to either compute the CSMs directly from rawdata or use given image data to comput them. This code existed before but has now been centralised to this class.
    - furhter methods necessary to guarantee backwards compatibility such as `get_dim` and `get_data`.
    - the interface to refactor the coil combination and coil application in the forward and backward model has already been initialised as `forward()` and `backward` but is neither `public` nor implemented yet.

Deleted classes
- `CoilData`
    - The forward and backward methods of the `MRAcquisitionModel` employed this class, this has been replaced by `CFImage`=typedef for `ISMRMRD::Image<complex_float_t>` since this has the same operator() implementation and hence did not require any change in the `fwd_` and `bwd_` methods. 
- `CoilDataAsCFImage`
- `CoilDataContainer`
- `CoilDataVector`
- `CoilImagesContainer`
- `CoilImagesVector`
- `CoilSensitivitiesContainer`
- `CoilSensitivitiesAsImages`

Further  changes
- `ImageWrap` received two new methods `get_complex_data(float* re, float*im)` and `set_complex_data(float* re, float*im)` to be able to pass data using the current interfaces.
- `mrtest_auxiliary_funs.h/cpp` is compiled now and can be used to define some auxiliary functions in the MR `C++` tests.
Unfortuantely there was a white space vs tab issue so that some parts are marked as changed which don't actually carry any new code. God, I never felt so alive!

Important note: the Python tests all run with this new code, however nothing has been done to the Matlab side.


KrisThielemans:
thanks!

If it's ok with you, let's postpone merging this until SIRF 2.2 is out. We can do a code review of course.

johannesmayer:
Definitely

rijobro:
@evgueni-ovtchinnikov have you had a chance to look at this yet? Would be good to get it merged as soon as possible after the SIRF release.

src/xGadgetron/cGadgetron/tests/mrtest_auxiliary_funs.h
```diff
@@ -0,0 +1,20 @@
+// author: Johannes Mayer
```
rijobro, line 1: should have proper copyright

rijobro: check for all files

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -127,62 +127,62 @@ def mr_data_path():
...
-        for m in range(nm):
-            pygadgetron.cGT_getCoilData \
-                (self.handle, m, re.ctypes.data, im.ctypes.data)
-            j = i + nz
-            array[:, i : j, :, :] = re + 1j * im
-            i = j
-        return array
-
-DataContainer.register(CoilImageData)
+#class CoilImageData(DataContainer):
```
rijobro, line 130: i would delete rather than comment (we have version control for that!)

evgueni-ovtchinnikov:
I believe we could do without extra `ImageWrap::get_complex_data()`. We only used it for `CoilData`, which you are removing, are you not? Handling `re` and `im` separately leads to unnecessary copying.

same for `set` of course

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -127,62 +127,62 @@ def mr_data_path():
...
+#            pygadgetron.cGT_getCoilData \
+#                (self.handle, m, re.ctypes.data, im.ctypes.data)
+#            j = i + nz
+#            array[:, i : j, :, :] = re + 1j * im
+#            i = j
+#        return array
+#
+#DataContainer.register(CoilImageData)
 
 class CoilSensitivityData(DataContainer):
```
evgueni-ovtchinnikov, line 187: inherit from `ImageData`

evgueni-ovtchinnikov: have done 3/4 dimensions switch, hope will not break anything

evgueni-ovtchinnikov: actually, a bit more work is needed on re-arranging data a la 176-183 above, I leave it to you

johannesmayer:
> I believe we could do without extra `ImageWrap::get_complex_data()`. We only used it for `CoilData`, which you are removing, are you not? Handling `re` and `im` separately leads to unnecessary copying.
> 
> same for `set` of course

I have implemented this for the `cGT_getCoilData(...)` and `cGT_appendCSM(...)`.
These pass two separate pointers for real and imaginary and I could not find a way to merge two `float*` into one `complex_float_t*`. 
Is there a way around this?

johannesmayer:
> > I believe we could do without extra `ImageWrap::get_complex_data()`. We only used it for `CoilData`, which you are removing, are you not? Handling `re` and `im` separately leads to unnecessary copying.
> > same for `set` of course
> 
> I have implemented this for the `cGT_getCoilData(...)` and `cGT_appendCSM(...)`.
> These pass two separate pointers for real and imaginary and I could not find a way to merge two `float*` into one `complex_float_t*`.
> Is there a way around this?

Ah sorry, now I see that the `getCoilData()` is not necessary if inherited from `ImageData`. Is there a way to append images to `ImageData` s.a. `appendCSM` would do?

rijobro:
I'm a little lost, is this a question for @evgueni-ovtchinnikov?

evgueni-ovtchinnikov:
@johannesmayer sorry a bit too busy today, hope to have time tomorrow

BTW, have you tried my suggestion on the implementation of `as_array()` sent to you by e-mail (May 8)?

johannesmayer:
> @johannesmayer sorry a bit too busy today, hope to have time tomorrow
> 
> BTW, have you tried my suggestion on the implementation of `as_array()` sent to you by e-mail (May 8)?

Yes it is in there now, Coil-Stuff is inherited from ImageData and I tested it for multislice-2D and 3D -multi-contrast coilmaps and both work. [Of course then (coil-channels nc, contrasts ns, z, y, x) are 5 dimensions in a [nc, ns*nz, ny, nx] sized 4D array ]

src/xGadgetron/pGadgetron/Gadgetron.py
johannesmayer: I think the Matlab side should be fine as far as I can tell.
If the files containing the `mGT_*` are auto-created during the build process with Matlab switched on the ones that are called in the Matlab class for CoilSensitivities should be available...

src/xGadgetron/pGadgetron/Gadgetron.py
johannesmayer: done

johannesmayer:
@evgueni-ovtchinnikov Hey evgueni, I addressed all the changes you proposed, the tests on travis passed. Could you please check again and say if you are happy? 

To not break backwards compatibility I had to implement the `ImageWrap::set/get_complex_data(float*, float*)` methods for the methods `cGT_getCoilData` and `cGT_appendCSM`. 
These were previously interfaced with `CoilDataAsCFImage` which does not exist any more.

KrisThielemans:
Is this one ready? Possibly best to merge master in here to be sure.

Can you add a line in CHANGES.md? (if not done yet)

KrisThielemans:
How backwards compatible is this?

evgueni-ovtchinnikov:
@johannesmayer 
>I have implemented this for the cGT_getCoilData(...) and cGT_appendCSM(...).
>These pass two separate pointers for real and imaginary and I could not find a way to merge two float* into one complex_float_t*.
>Is there a way around this?

Now that the coil data container `CoilSensitivitiesVector` inherits from `ImageData`, you can use 
`cGT_getImageDataAsCmplxArray` instead of `cGT_getCoilData`, and `cGT_appendCSM` can use `ImageData.append` (and should actually be renamed `cGT_appendImageData`).

evgueni-ovtchinnikov:
@KrisThielemans 
> How backwards compatible is this?

C++ not compatible, Python and Matlab can be made compatible I believe.

johannesmayer:
> How backwards compatible is this?

I made the Python mostly backwards compatible, you just cannot access the `CoilImageData` class any more, that is now simpy an image.  But that was only an intermediate step anyway where @evgueni-ovtchinnikov already had supplied the interface to go from acquisition data to coilmaps directly.

@evgueni-ovtchinnikov if I will delete the methods like `append_csm()` in python then this will additionally be unavailable with the plus of being able to delete the two methods I added to the `ImageWrap`. You have to tell me what to do :)

evgueni-ovtchinnikov:
@johannesmayer 
> Is there a way to append images to ImageData s.a. appendCSM would do?

`ISMRMRDImageData` has `append` (see also my previous comment).

evgueni-ovtchinnikov:
@johannesmayer 
> if I will delete the methods like append_csm() in python then this will additionally be unavailable

you could keep it as an alias for `ImageData.append` for backward compatibility

btw if we really care about backward compatibility, then we should derive `CoilSensitivityData` from `CoilImageData` derived from `GadgetronImagesVector` in C++ and ImageData in Python/MATLAB

johannesmayer:
> @johannesmayer
> 
> > if I will delete the methods like append_csm() in python then this will additionally be unavailable
> 
> you could keep it as an alias for `ImageData.append` for backward compatibility
> 
> btw if we really care about backward compatibility, then we should derive `CoilSensitivityData` from `CoilImageData` derived from `GadgetronImagesVector` in C++ and `ImageData` in Python/MATLAB

Hell no...  I'll keep the alias for appending and address the rest on how to access the data without having to add extra `ImageWrap` stuff

johannesmayer:
Ok I deleted the interface to the 'ImageWrap` again and used the `fill` method instead in the `append` method so now it is backwards compatible. I think this should have addressed everything from above.

johannesmayer:
Ok I discovered that there are two methods for the Matlab method `csm_as_array()` that might not work now.

However, I don't know how that is to be done. As discussed I now deleted the interface to `CoilSensitivityData` that is used there so someone would have to change the Matlab code to access the complex data, I cannot do that unfortunately.

johannesmayer:
@evgueni-ovtchinnikov I found your example script where you want to compare against inati. 

My next PR will involve sending the data to Gadgetron to the `CoilComputationGadget` that I wrote. 
There only the Inati method will be used so this becomes obsolete, SRSS will not exist any more.

I will delete this part in the example script now since I would delete it in the next one anyway and it is not compatible with this coil computation setup. And we will save another extra inheritance layer nobody will probably ever use.

evgueni-ovtchinnikov:
Working with complex data in Matlab: you just need to treat a complex array of total size (product of all dimensions) `n` as 2-by-`n` dimensional array of floats  - see e.g `as_array` and `fill` methods of `AcquisitionData`.

evgueni-ovtchinnikov:
Inati and SRSS: I do not like the idea of removing SRSS - at the very least, we need to see how the two compare first (you could do it by running coil sensitivities script on master).

johannesmayer:
> Inati and SRSS: I do not like the idea of removing SRSS - at the very least, we need to see how the two compare first (you could do it by running coil sensitivities script on master).

In my mind's eye, SIRF is supposed to be lightweight wrapping functionality from the engines.  SRSS is about equally as fast as the Inati method but requires a huge number of code lines to for results that are always worse than Inati. I would strongly opt for dropping SRSS.

KrisThielemans:
Clearly, this is a decision for the MR experts, including @johannesmayer @ckolbPTB @DANAJK

ckolbPTB:
I think we can rely on Gadgetron (Inati) to do a good job in calculating the coil maps and do not need our own implementation (SRSS).

johannesmayer:
I want to stress again that THIS PR does not even delete any SRSS functionality nor does it add any Inati functionality via Gadgetron. It is just restructuring shifting code around and cleaning up.  
It is just that the comparison with Inati in the example script cannot be performed anymore because I deleted the `CoilImageData` class in the process.

johannesmayer:
@evgueni-ovtchinnikov 
ok so now I allowed getting images out of the C++ level to compute the Inati output.
Now the example will run and allow Inati even in this PR...

However, setting the parameter via the call `CSMs.calculate(processed_data, method='SRSS(niter = %d)' % nit)` did not work with this parameter setter and I couldn't figure out why so I set it from `CSMs.smoothness = nit`.

> Working with complex data in Matlab: you just need to treat a complex array of total size (product of all dimensions) `n` as 2-by-`n` dimensional array of floats - see e.g `as_array` and `fill` methods of `AcquisitionData`.

I cannot do this since I don't have SIRF in Matlab, could you please do this for me?

KrisThielemans:
Thanks @johannesmayer 

Codacy complains about the CSM.`calculate` method being "too complex", presumably saying that you need to split it into functions. I'm assuming this will be nearly @evgueni-ovtchinnikov original implementation, and it's not your business in cleaning our code up, so let's ignore this.

All Linux jobs on Travis are green! OSX jobs still running, but it's looking good (better than it was anyway 😄 ).

We can easily defer to MATLAB job till later by adding it to #555

DANAJK:
Just to add to the mix, on the Gadgetron course last week, they were giving a demo of the Gadgetron external interface. They sent data from Gadgetron to MATLAB, from where a BART routine was called to calculate coil sensitivities!

The Gadgetron external interface looks really cool. I'm wondering if it is helpful to us in potentially reducing the amount of wrappers we have to write. I would recommend watching the talk of Kristoffer Knudsen (Day 3, Part 1) when it gets onto their YouTube playlist. He has a very clear way of thinking and explaining:
[https://www.youtube.com/playlist?list=PL3dITfhovhLW-ySr_KPBS8KKI9tdLjfdR](https://www.youtube.com/playlist?list=PL3dITfhovhLW-ySr_KPBS8KKI9tdLjfdR) 

The second talk was the one that showed the link to BART (if I recall correctly).

KrisThielemans:
Travis is green.

@evgueni-ovtchinnikov ok to merge? If so, please do so and update the MATLAB issue.

evgueni-ovtchinnikov:
@johannesmayer: I noticed you are using `CoilSensitivitiesVector` class for objects storing either coil sensitivity maps or coil images, and introduced a method `calculate_csm(void)` that throws if a `CoilSensitivitiesVector` object that stores coil sensitivity maps calls this method.

I am afraid this is bound to confuse users.

I cannot understand why you are so dead set against having `CoilImagesVector`. Perhaps this is because I erroneously suggested that `CoilSensitivitiesVector` should be derived from it? Actually, they should be both derived from `GadgetronImagesVector` (as it essentially currently is in Python and, though in a much more complicated setting, in C++).

So how about this:
```
	class CoilImagesVector : public GadgetronImagesVector {
	public:
		void compute(MRAcquisitionData& ac);
		// whatever else might be useful
	};

	class CoilSensitivitiesVector : public GadgetronImagesVector {
	public:
		void compute(CoilImagesVector& cis);
		void compute(MRAcquisitionData& ac)
		{
			CoilImagesVector cis;
			cis.compute(ac);
			compute(cis);
		}
		void set_csm_smoothness(int s)
		{
			csm_smoothness_ = s;
		}
		// whatever else might be useful
    };
```

johannesmayer:
@evgueni-ovtchinnikov ok so now i papered under the crack, i don't know why I missed this before I got confused debugging.

I will not put in any other layers. If you want more layers then you have to put them in.

There is this weird swapping still that I don't understand why inati is switching dynamics around, 
```
        def __calc_from_acquisitions(self, data, method_name):
        ...

            cis_array = self.as_array()
            csm, _ = coils.calculate_csm_inati_iter(cis_array)
            
            nc, nz, ny, nx = self.dimensions()
            ns = self.number() # number of total dynamics (slices, contrasts, etc.)
            nz = nz//ns        # z-dimension of a slice

            csm = numpy.reshape(csm, (nc, ns, nz, ny, nx))
            csm = numpy.swapaxes(csm,0,1)
            csm = numpy.reshape(csm, (nc, ns*nz, ny, nx))
            
            self.append(csm.astype(numpy.complex64))
```
Like this it works. 

If i apply the swapping axis to the `cis_array` before giving it to `calulcate_csm_inati_iter` then it is still in the wrong order, so does the ISMRMRD tool interchange coils and dynamics? I don't know...

evgueni-ovtchinnikov:
@johannesmayer: ok I will add `CoilImagesVector` after merge, but I would reiterate that I do not suggest extra layer (that was a mistake), I suggest separate class for coil images - having a flag in a class defining the instance type is so non-OO.

I am afraid you have lost me with your csm re-shapings - I cannot make head or tale of them. So far as I can see `calculate_csm_inati_iter` does not reshape its first argument except for adding singleton `z` dimension to a 2D image.

johannesmayer:
> @johannesmayer: ok I will add `CoilImagesVector` after merge, but I would reiterate that I do not suggest extra layer (that was a mistake), I suggest separate class for coil images - having a flag in a class defining the instance type is so non-OO.

Ok thank you. I agree, the flags were just old stuff to help avoid doing something wrong. I had forgotten to delete them.


As for the reshaping, one part of the PR is now the update of `as_array()` to get channels out of the `Images` class where it now says:
``` 
   if len(dims) != 4:
            return array

        nc, nz, ny, nx = dims
        ns = self.number() # number of total dynamics (slices, contrasts, etc.)
        nz = nz//ns        # z-dimension of a slice

        # hope numpy is clever enough to do all this in-place:
        array = numpy.reshape(array, (ns, nc, nz, ny, nx))
        array = numpy.swapaxes(array,0,1)
        array = numpy.reshape(array, (nc, ns*nz, ny, nx))

        return array
```
So since we call `as_array()` in `__calc_from_acquisitions(self, data, method_name)` , when you call `as_array` to get the CSM out of the object, then the `reshape` and `swapaxes` have been applied twice so I have to do the inverse after computing the coilmaps with the ismrmrd tool. That's somewhat of an ugly workaround but I don't know any good solution here.

PR #657, opened 2020-05-13
use VTK for inverse displacement field

rijobro:
found discontinuities in inverse displacement fields generated with NiftyReg. Slicer seemed to avoid this and uses VTK under the hood ([link](https://discourse.slicer.org/t/using-the-invert-displacement-field-function/4768/2)), so giving it a go here.


src/Registration/cReg/tests/test_cReg.cpp
```diff
@@ -1343,6 +1343,35 @@ int main(int argc, char* argv[])
...
+        const auto nr_disp_fwd = dynamic_cast<const NiftiImageData3DDisplacement<float>&>(*NA.get_displacement_field_forward_sptr());
+        const auto nr_def_fwd  = NiftiImageData3DDeformation<float>(nr_disp_fwd);
+        const auto nr_def_inv  = nr_def_fwd.get_inverse();
+        const auto nr_disp_inv = NiftiImageData3DDisplacement<float>(*nr_def_inv);
+
+        // Get inverse using VTK
+        auto vtk_disp_inv = nr_disp_fwd.get_inverse_vtk();
+
+        nr_disp_inv.write("/Users/rich/Desktop/nr_disp_inv");
+        vtk_disp_inv.write("/Users/rich/Desktop/vtk_disp_inv");
```
KrisThielemans, line 1362: obviously will have to change this. Not sure what we're using as a strategy for  temporary files

rijobro: yes, i'll just delete. only needed for debugging.

rijobro:
According to the link in my first post, VTK does a better job of getting inverse fields than ITK, which is why Slicer uses it for that purpose. 

This is still WIP, see this thread on the vtk discourse website for some help I've been getting: https://discourse.vtk.org/t/nifti-to-vtkimagedata-in-memory/3302/5

rijobro:
Think I'm done with this one. Attaching images of inverse displacement fields calculated with NiftyReg and VTK. In the itkSnap image, NiftyReg is on the left and VTK is on the right. 

With VTK there are no discontinuities. Arguably, the output is too smooth, but I think it shows that the functionality is working.

image

image

rijobro:
@KrisThielemans , do you still want to wait for after 2.1 on this one? It seems stable to me and I have no plans to add to it.

Issue #659, opened 2020-05-19
building SIRF picks-up old installed files


KrisThielemans:
Order of include directories is currently problematic. Checking with `make VERBOSE=ON` shows that `cstir`  is compiled with an order like `various-sirf-dirs .../install/include .../hdf5/... sirf-reg-dirs STIR-install-dir/include ...`

This means that if there's an old STIR version in `.../install/include`, it gets picked up before the actual `STIR-install-dir/include` that I pointed to.

I guess (but am not sure) this is because `cgadgetron` uses `ISMRMRD_INCLUDE_DIR`, https://github.com/SyneRBI/SIRF/blob/dcf7a4d37153f3d6334386255767e7d777f0aee6/src/xGadgetron/cGadgetron/CMakeLists.txt#L47

Possible ways to resolve this:
- why does `cstir` depend on `cgadgetron`?
- not use `ismrmrd` headers directly in cgadgetron include files, but this seems a lot of work
- make sure that `cstir` uses STIR include paths first
- don't use `target_include_directories` but just depend on `ISMRMRD:ISMRMRD` although that will likely give the same dependencies.

rijobro:
For your last point, we've already made the change such that we don't use `target_include_directories` and only use the target `ISMRMRD::ISMRMRD`:

https://github.com/SyneRBI/SIRF/blob/dcf7a4d37153f3d6334386255767e7d777f0aee6/src/xGadgetron/cGadgetron/CMakeLists.txt#L75

rijobro:
The problem is that the SB installs everything into the same location. We could modify the SB installation method for better compartmentalisation? The `env_ccppetmr.sh` would have to be slightly more complicated, but everything should be tidier. E.g,:

```
Install
    STIR
        include
        lib
        share
    Gadgetron
        include
        lib
        share
etc
```

rijobro:
Lastly, could this be the problematic bit of code? Put in by me 17 months ago apparently. Not sure why it uses the install interface as we don't currently install our headers.

https://github.com/SyneRBI/SIRF/blob/dcf7a4d37153f3d6334386255767e7d777f0aee6/src/xSTIR/cSTIR/CMakeLists.txt#L39-L42

KrisThielemans:
> For your last point, we've already made the change such that we don't use target_include_directories and only use the target ISMRMRD::ISMRMRD:

[...]

PR #663, opened 2020-05-20
ISMRMRD target_include_dir only if no ISMRMRD::ISMRMRD target

rijobro:
Fixes https://github.com/SyneRBI/SIRF/issues/659 (maybe)


KrisThielemans:
feels a bit weird actually to specify a link target to tell CMake what to do with include files. probably it'd be better to have
```cmake
target_include_directories(cgadgetron PUBLIC ISMRMRD::ISMRMRD)
```
but the CMake doc doesn't make it clear if this can be done.

Seems I'm pushing my CMake knowledge with this one...

rijobro:
My understanding was that an imported library would be created during the `*Targets.cmake`, which would itself use the `target_include_directories`. Then you could use `target_link_libraries` in your project and it would already know which headers are needed. 

E.g. in `ISMRMRDTargets.cmake`, I have:

```
add_library(ISMRMRD::ISMRMRD SHARED IMPORTED)

set_target_properties(ISMRMRD::ISMRMRD PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include/ismrmrd"
  INTERFACE_LINK_LIBRARIES "/usr/local/Cellar/hdf5/1.12.0/lib/libhdf5.dylib;/usr/local/opt/szip/lib/libsz.dylib;/usr/lib/libz.dylib;/usr/lib/libdl.dylib;/usr/lib/libm.dylib"
)
```

KrisThielemans:
superseded by #840

Issue #654, opened 2020-05-12
KOSMAPOSLReconstructor.set_up() failure


evgueni-ovtchinnikov:
In demo `hkem_reconstruction.py`,  `KOSMAPOSLReconstructor.set_up()` fails with error message 
```
ERROR: Error opening file 
```
apparently thrown by `stir::IterativeReconstruction::post_processing()`

evgueni-ovtchinnikov:
must be something to do with VM, works ok on Windows

KrisThielemans:
that would be weird. Are you sure your VM STIR was updated to 4.0.1?

KrisThielemans:
`update_VM.sh -t origin/master` would do the whole thing possibly, although with probably with the problem of the STIR tag, so `rm -rf ~/devel/buildVM/sources/STIR` first.

KrisThielemans:
confirmed on my updated VM.

KrisThielemans:
@evgueni-ovtchinnikov can you please clarify the calling sequence here? I can't immediately find where `post_processing` is called. Do we do that because STIR's `KOSMAPOSLReconstruction::set_up` is incomplete? It does look that way.

Do you know if it is thrown in `stir::IterativeReconstruction::post_processing()` or further in the chain? (e.g. `stir::KOSMAPOSLReconstruction::post_processing()`)?

evgueni-ovtchinnikov:
`KOSMAPOSLReconstruction::post_processing` is supposed to read anatomical images, but their filenames are empty.

KrisThielemans:
ok. if you'd be able to point to the line, it would help @danieldeidda 

Would resolving https://github.com/UCL/STIR/issues/537 solve the SIRF problem? (i.e. we'd no longer call `post_processing`)

evgueni-ovtchinnikov:
@danieldeidda line 224 in `KOSMAPOSLReconstruction.cxx` fails to read from file with empty name.

danieldeidda:
> 
> 
> @danieldeidda line 224 in `KOSMAPOSLReconstruction.cxx` fails to read from file with empty name.

Could it be because that member is now a vector? in the parfile we need to put the name inside {}

KrisThielemans:
@evgueni-ovtchinnikov could you try with https://github.com/UCL/STIR/pull/540?

KrisThielemans:
There's also the SIRF-Exercises I guess

evgueni-ovtchinnikov:
[...]

PR #666, opened 2020-05-22
No longer call post_processing in recon.set_up()

KrisThielemans:
This was necessary for previous versions of STIR but now breaks things.

See #654


KrisThielemans:
ok. let's abandon this one then. only cosmetically different from master now.

Issue #680, opened 2020-05-27
f3d permuted output


AnderBiguri:
When running the f3d registration, I noticed that my outputs/displacement/deformations are X-Y permuted. For a given identity transformation I get correct Displacement fields, they are just non-zero, they describe a X-Y permutation. You can see the output image being permuted too. 

You can test this with:

```
import sirf
import sirf.STIR as stir
import sirf.Reg as reg
import matplotlib.pyplot as plt
import numpy

# load your favorite Nifti Image. 
reg.ImageData(file)
reg_alg = reg.NiftyF3dSym()
reg_alg.set_initial_affine_transformation(reg.AffineTransformation(numpy.eye(4, dtype=numpy.float32))) # not necessary, but making sure.

reg_alg.set_reference_image(image)
reg_alg.add_floating_image(image)
# Process
reg_alg.process()
# Store results. 
## Change for your folders
reg_alg.get_output(0).write( "res/out_"))
reg_alg.get_deformation_field_forward(0).write(os.path.join(path, "res/DefVF_") )
reg_alg.get_deformation_field_inverse(0).write(os.path.join(path, "res/iDefVF_") )
reg_alg.get_displacement_field_forward(0).write(os.path.join(path, "res/DispVF_"))
reg_alg.get_displacement_field_inverse(0).write(os.path.join(path, "res/iDispVF_"))

```
I can reproduce the same results with different images (i.e. I have not only tried an identity transform). I can also reproduce it by calling `sirf_registration --algo f3d --ref ....` via the command line.

I can **not** reproduce it if I call niftyreg directly:

`reg_f3d ....` 

No parameters have been set in any of the cases, whichever are default have been used. 

In my particular test, Displacement fields look like this:

![Screenshot from 2020-05-27 17-19-18](https://user-images.githubusercontent.com/11854388/83049758-725a6080-a043-11ea-9722-c50747b1c997.png)

KrisThielemans:
weird and worrying.

Can you confirm that this image is the displacement field output (I guess so, as z=0)? Is your registered output image ok? What happens if you use the output disp field to resample the "floating"?

Are the niftyreg disp fields zero?

AnderBiguri:
Answers:

1) Yes, definitely displacement (and not deformation), but I can see the permute in the deformation too. 
2) Yes, the image looks OK, but permuted. It is a clean permute, as the displacement fields suggest.
[...]

PR #681, opened 2020-05-27
Option to use symmetric f3d. default to off

rijobro:
Fixes https://github.com/SyneRBI/SIRF/issues/680.

Added a test that says "if ref==flo, I expect registered image to ~= ref". This test fails with old (symmetric) implementation, but passes for the new fix.


rijobro:
@AnderBiguri could you give this a go for me and confirm? Since use_symmetric defaults to off, python and matlab should both use the non-symmetric version by default now. I haven't bothered exposing that option (since the functionality is broken).

Suppose I'll add something to the doxygen and Changes.md

AnderBiguri:
@rijobro Confirmed! This works without permuting the output.

Issue #686, opened 2020-05-28
virtual methods not working correctly when passing between wrapped (python) and C++


rijobro:
I have the following code:
```
import sirf.STIR as pet
pet.AcquisitionData.set_storage_scheme('memory')
out = pet.AcquisitionData("sino.hs")
out.axpby(1,2,out)
```
The `axpby` goes to the method in `DataContainer`, in `SIRF.py`. That method then calls `cSIRF_axpby` in `csirf.cpp`. `DataHandles` are cast to `DataContainer`, and then the correct virtual implementation of `axpby` is chosen (e.g., `PETAcquisitionData`).

I know that `out` is `ProjDataInMemory`, and so should go to the virtual function there (`ProjDataInMemory::axpby`). However, it ends up in `ProjData::axpby`.

Whilst debugging, I do some casting:

```
template<class Type>
static
void
print_quick(const Type * obj, const std::string &var_name, const std::string &class_name)
{
    std::cout << var_name << " is " << class_name << "? " << (stir::is_null_ptr(obj) ? "No" : "Yes") << "\n";
}

void
PETAcquisitionData::axpby(
const void* ptr_a, const DataContainer& a_x,
const void* ptr_b, const DataContainer& a_y
)
{
    float a = *(float*)ptr_a;
    float b = *(float*)ptr_b;

    // Are they PETAcquisitionData?
    const PETAcquisitionData *x_temp = dynamic_cast<const PETAcquisitionData*>(&a_x);
    const PETAcquisitionData *y_temp = dynamic_cast<const PETAcquisitionData*>(&a_y);
    print_quick(x_temp, "x_temp", "PETAcquisitionData");
    print_quick(y_temp, "y_temp", "PETAcquisitionData");

    // Are they ProjData?
    const ProjData* x = x_temp->data().get();
    const ProjData* y = y_temp->data().get();
    print_quick(x, "x", "ProjData");
    print_quick(y, "y", "ProjData");

    // Are they ProjDataInMemory
    ProjDataInMemory* this_pdim = dynamic_cast<ProjDataInMemory*>(this->data().get());
    const ProjDataInMemory* x_pdim = dynamic_cast<const ProjDataInMemory*>(x);
    const ProjDataInMemory* y_pdim = dynamic_cast<const ProjDataInMemory*>(y);

    print_quick(this_pdim, "this_pdim", "ProjDataInMemory");
    print_quick(x_pdim, "x_pdim", "ProjDataInMemory");
[...]

PR #688, opened 2020-05-29
if memory, convert PDFromStream to PDInMemory

rijobro:
It seems that if `storage_scheme` is set to memory, sinograms are still read as `ProjDataFromStream`. Here, we convert to `ProjDataInMemory` if desired. Fixes problem noted in #686 (which has a misleading title).


KrisThielemans:
tricky possibly. `fill` etc would normally change content in file/stream. Probably no longer with this PR (I didn't check).

could work for read-only streams.

KrisThielemans:
Having `AcquisitionDataInMemory` is fine. I'm still worried about having all data in memory (but I didn't check)

rijobro:
If the data were too big that they couldn't all fit into memory, then shouldn't the user be using  `storage_scheme==file` anyway?

KrisThielemans:
I meant that if `storage_scheme=memory`, if we're not careful we could have the following
```python
acq=pet.AcquisitionData("somefile")
acq.fill(1)
```
will no longer change the data in file. I believe that this would be a major change. Also, the behaviour has to be independent of `storage_scheme`

KrisThielemans:
we do open `PETAcquisitionData` read-write I believe https://github.com/SyneRBI/SIRF/blob/c9846869b536527280e1c2cd2f98c618c59d9541/src/xSTIR/cSTIR/include/sirf/STIR/stir_data_containers.h#L85

so I guess the proposed change is not safe. sorry

rijobro:
You're right that this is quite a fundamental change. However, if I ran
```
pet.AcquisitionData.set_storage_scheme('memory')
acq=pet.AcquisitionData("somefile")
acq.fill(1)
```
I wouldn't expect the underlying file to have been modified.

As an aside, I think files are opened as read only, so you would need to clone the data to be able to use the `fill` method.

KrisThielemans:
> I wouldn't expect the underlying file to have been modified.

Possibly it is undesirable/surprising, but I think currently it does modify the file. No idea what happens on the MR side.

We could open read-only. This would break backwards compatibility in a fairly fundamental way.

Alternatively, we could open a temp file for every file that we open, such that the `fill` would be possible, but would not modify the original. That would still break backwards compatibility.

Would have to be for SIRF 3.0 therefore.

rijobro:
```
In [1]: import sirf.STIR as pet                                                                                                                               

In [2]: pet.AcquisitionData.set_storage_scheme("memory")                                                                                                      

In [3]: a = pet.AcquisitionData('EM_g1.hs')                                                                                                                   

INFO: Setting energy window keys as in STIR3.0

In [4]: a.fill(1)                                                                                                                                             
---------------------------------------------------------------------------
error                                     Traceback (most recent call last)
<ipython-input-4-ded0ac23c507> in <module>
----> 1 a.fill(1)

~/Documents/Code/SIRF-SuperBuild/Install/python/sirf/STIR.py in fill(self, value)
    735         assert self.handle is not None
    736         if self.read_only:
--> 737             raise error('Cannot fill read-only object, consider filling a clone')
    738         if isinstance(value, numpy.ndarray):
    739             if value.dtype is numpy.dtype('float32'):

error: ??? 'Cannot fill read-only object, consider filling a clone'
```

KrisThielemans:
old or new version? And with `storage_scheme=file`?

why the read/write in the above line?

rijobro:
New commit means that if storage scheme is `file` you'll need to clone (as before). However, if `memory`, you won't need to. This is different from previous functionality, but more correct IMO.

rijobro:
> old or new version? And with storage_scheme=file?
> why the read/write in the above line?

Write protection was happening on Python side: 

https://github.com/SyneRBI/SIRF/blob/4e90d8ee9bbc93953afcd789b895330abce6c2e5/src/xSTIR/pSTIR/STIR.py.in#L614-L617

KrisThielemans:
needs checking what happens on MR. would need line in CHANGES as backwards compatibility break. Cannot be merged until after 2.2

rijobro:
Superseded by https://github.com/SyneRBI/SIRF/pull/694.

PR #689, opened 2020-05-29
Pdim axpby

rijobro:
Requires https://github.com/SyneRBI/SIRF/pull/688 and https://github.com/UCL/STIR/pull/557.

### DIFF
--- a/src/xSTIR/cSTIR/stir_data_containers.cpp
+++ b/src/xSTIR/cSTIR/stir_data_containers.cpp
@@ -94,45 +94,19 @@ const void* ptr_a, const DataContainer& a_x,
 const void* ptr_b, const DataContainer& a_y
 )
 {
-	float a = *(float*)ptr_a;
-	float b = *(float*)ptr_b;
-	DYNAMIC_CAST(const PETAcquisitionData, x, a_x);
-	DYNAMIC_CAST(const PETAcquisitionData, y, a_y);
-	//PETAcquisitionData& x = (PETAcquisitionData&)a_x;
-	//PETAcquisitionData& y = (PETAcquisitionData&)a_y;
-	int n = get_max_segment_num();
-	int nx = x.get_max_segment_num();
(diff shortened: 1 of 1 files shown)


rijobro:
Running the following code:
```
for i in range(10):
    out.axpby(1,2,out)
    out.axpby(1,2,out,out)
```
Sees a dramatic increase in speed when comparing new code to old - 2.85 vs 46.6 s!

rijobro:
```
In [1]: import sirf.STIR as pet                                                                                                                               

In [2]: pet.AcquisitionData.set_storage_scheme("memory")                                                                                                      

In [3]: a = pet.AcquisitionData('EM_g1.hs')                                                                                                                   

INFO: Setting energy window keys as in STIR3.0

In [4]: a.fill(1)                                                                                                                                             
---------------------------------------------------------------------------
error                                     Traceback (most recent call last)
<ipython-input-4-ded0ac23c507> in <module>
----> 1 a.fill(1)

~/Documents/Code/SIRF-SuperBuild/Install/python/sirf/STIR.py in fill(self, value)
    735         assert self.handle is not None
    736         if self.read_only:
--> 737             raise error('Cannot fill read-only object, consider filling a clone')
    738         if isinstance(value, numpy.ndarray):
    739             if value.dtype is numpy.dtype('float32'):

error: ??? 'Cannot fill read-only object, consider filling a clone'
```

rijobro:
@KrisThielemans - These PRs will fail for `DEVEL_BUILD=OFF` since the required features are only available on STIR `master` (well, `release_4`). Thoughts?

KrisThielemans:
>  These PRs will fail for `DEVEL_BUILD=OFF` since the required features are only available on STIR `master` (well, `release_4`). Thoughts?

I'd recommend updating the SIRF-SuperBuild `STIR_TAG` to the tag on `release_4` . Obviously, can only do that after 2.2 is out. So... a PR on the SuperBuild.

PR #691, opened 2020-05-29
Pdim asarray fill

rijobro:
as_array, fill, norm, multiply, divide, etc. all sped up for `AcquisitionDataInMemory`. Requires https://github.com/UCL/STIR/pull/558.

Seems to be factor ~10 faster.

Diff will simplify as other PRs are merged.


KrisThielemans:
actually we should put `fill` and some other overloads in STIR I guess?

rijobro:
I thought about that already, but `fill` (and similar) are templated in STIR, which means they can't be overloaded. I looked into using `boost::any`, but it got a little convoluted. Happy to hear if you have suggestions.

rijobro:
À la: https://stackoverflow.com/a/7968213/12411867

PR #693, opened 2020-06-01
Added doc about Displacement vs Deformation fields

### LINKED ISSUE #673: Deformation vs Displacement field docs
I see that you can get both, and with a chat with @KrisThielemans seems that Deformation fields will have absolute locations of sampling points while Displacement fields will have relative ones.

I have not seen this documented anywhere in SIRF, maybe we should add it somewhere? Either the classes themselves or at least in the samples?

AnderBiguri:
Resolves #673 

Added it in MATLAB, C, python classes and MATLAB and python examples.

### DIFF
--- a/examples/Matlab/Registration/sirf_registration.m
+++ b/examples/Matlab/Registration/sirf_registration.m
@@ -17,9 +17,9 @@ function sirf_registration(varargin)
 %   -------------------------------------------------------------
 %
 %   --warped_prefix <fname>      warped image filename
-%   --disp_fwd_prefix <fname>    forward displacement field image
+%   --disp_fwd_prefix <fname>    forward displacement field image, Displacement fields encode relative change
 %   --disp_inv_prefix <fname>    inverse displacement field image
-%   --def_fwd_prefix <fname>     forward deformation field image
+%   --def_fwd_prefix <fname>     forward deformation field image,  Deformation fields encode absolute locations
 %   --def_inv_prefix <fname>     inverse deformation field image
 %
 %   Optional arguments for rigid/affine algorithms (aladin/spm)
--- a/examples/Python/Registration/sirf_registration.py
+++ b/examples/Python/Registration/sirf_registration.py
@@ -13,8 +13,8 @@ Options:
   --algo <algo>                registration algorithm (aladin,f3d,spm) [default: aladin]
 
   --warped_prefix <file>       warped image filename prefix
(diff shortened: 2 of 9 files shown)


rijobro:
Thanks! Since you're putting it into executables, there's also `sirf_resample`, which would benefit from it. And there's C++ implementations of both `sirf_resample` and `sirf_registration`

rijobro:
also, if you're only updating documentation, you can use `[ci skip]` in your commits to not trigger Travis builds (means other builds won't have to wait in the queue).

AnderBiguri:
@rijobro  Where do you want it in the resampler examples? There is not space in the initial docs (I would need multiple lines) and it does not exist in the code as you access them dynamically. Any suggestion?

Also, I have added them to the examples in SIRF/examples. Where are the C++ examples?

rijobro:
C++ (live in `src/Registration/cReg/`):
```
➜  ~ sirf_resample -h

*** sirf_resample usage ***

  Required flags:
    -algo:		Resampling algorithm (currently only niftyreg)
    -ref:		reference image
    -flo:		floating image

  Optional flags:
    -eng_ref:		engine to open reference image (Nifti, STIR, Gadgetron)
    -eng_flo:		engine to open floating image (Nifti, STIR, Gadgetron)
    -output:		output image filename
    -interp:		interpolation (0=NN, 1=linear, 3=cubic, 4=spline)
    -add_affine:	add affine transformation
    -add_def:		add deformation transformation
    -add_disp:		add displacement transformation
    -adj:		adjoint transformation. Give ref and flo as you would in the forward case.
```
Could be added after `add_def` and `add_disp`.

You're right, no space for it in the python or matlab resample examples.

AnderBiguri:
@rijobro I have no idea where that functionality sits within the files, could you point me to whichever file is actually printing that? there is no file `sirf_resample.cxx` and I rather not open and read all of them to find the code hehe.

AnderBiguri:
hum. I broke CI, I have no idea how.

rijobro:
Sorry! It actually lives in synergistic code (since it requires ISMRMRD and STIR to be able to open those files): https://github.com/SyneRBI/SIRF/blob/master/src/Synergistic

rijobro:
> hum. I broke CI, I have no idea how.

all the more reason to `[ci skip]` 😂

AnderBiguri:
@rijobro point taken  :D I added the tiny docs to the c++ code.

PR #694, opened 2020-06-02
ProjDataInMemory changes

rijobro:
It seems that if `storage_scheme` is set to `memory`, sinograms are still read as `ProjDataFromStream`. Here, we convert to `ProjDataInMemory` if desired. Fixes problem noted in #686 (which has a misleading title).


rijobro:
merging as only Travis error was unrelated.

Issue #715, opened 2020-06-12
Changing storage scheme of AcquisitionData


rijobro:
If I do something along the lines of:
```
pet.AcquisitionData.set_storage_scheme("file")
a = pet.AcquisitionData(fname1)
pet.AcquisitionData.set_storage_scheme("memory")
b = pet.AcquisitionData(fname2)
print(a.get_storage_scheme())
```
Despite the fact that `a` uses `file` as its storage scheme, the printed output will be `memory`. This is because the storage scheme is a static variable. 

It seems to me that it would be safer to set the storage scheme for each `AcquisitionData` individually. It can default to `file`, since this is the current default. Could look like this:

```
a = pet.AcquisitionData(fname, scheme="file")
```

If we do not like this, then at the very least, we need to change `get_storage_scheme` so that it is no longer static and returns its own storage scheme. This could be done by dynamically casting to `ProjDataInMemory`, and returning the value of `is_null_ptr`.

evgueni-ovtchinnikov:
one would then have to specify somehow the storage scheme for forward projection, acquisitions processor and the results of most algebraic operations

rijobro:
I don't think that's necessarily true. A forward projector will be set up with an `AcquisitionData`. And that `AcquisitionData` knows which storage scheme it uses.

PR #716, opened 2020-06-12
_storage_scheme -> _default_storage_scheme

rijobro:
- fixes https://github.com/SyneRBI/SIRF/issues/715.

@evgueni-ovtchinnikov how about this. Change `_storage_scheme` to `_default_storage_scheme`. If using the static method, return the default. However, if the object exists, return the corresponding type.

I've added the test, which fails with the old code, but passes with these changes:
```
    AcquisitionData.set_storage_scheme("memory")
    ad = AcquisitionData(raw_data_file)
    AcquisitionData.set_storage_scheme("file")
    test.check_if_equal("memory", ad.get_storage_scheme())
```

### DIFF
--- a/src/xSTIR/mSTIR/tests/test1.m
+++ b/src/xSTIR/mSTIR/tests/test1.m
@@ -106,6 +106,14 @@ delta = max(image_array(:))*eps;
 test.check(s)
 test.check(v, delta)
 
+% Check that the storage scheme for a pre-initialised object
+% doesn't change just because the default has changed.
+temp = PET.AcquisitionData();
+temp.set_storage_scheme('memory');
+ad = PET.AcquisitionData(fullfile(pathname, filename));
+temp.set_storage_scheme('file');
+test.check_if_equal('memory', ad.get_storage_scheme());
+
 failed = test.failed;
 ntests = test.ntest;
 
--- a/src/common/+sirf/+Utilities/mTest.m
+++ b/src/common/+sirf/+Utilities/mTest.m
@@ -81,5 +81,20 @@ classdef mTest < handle
                 end
             end
         end
+        function check_if_equal(self, expected, value)
+            if value ~= expected
+                self.failed = self.failed + 1;
+                if self.verbose
+                    fprintf...
+                        ('+++ test %d failed: expected %e, got %e\n', ...
+                        self.ntest, expected, value);
+                end
+            else
+                if self.verbose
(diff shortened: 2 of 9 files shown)


KrisThielemans:
strategy seems ok to me, but is it necessary to change the C++ name (you don't on the python side)? One more thing that breaks backwards compatibility, for not so good reasons?

rijobro:
I can leave the naming as it is if you'd like, but it seems confusing to me.

Keeping the old name makes it unclear as to whether you're returning the default scheme or the scheme for the current object.

I didn't change the naming on the python side because it handles both cases - if object is `None`, return default scheme, else return scheme for specific object.

Thinking about it, I should probably update the matlab side...

src/xSTIR/mSTIR/tests/test1.m
```diff
@@ -106,6 +106,14 @@
 test.check(s)
 test.check(v, delta)
 
+% Check that the storage scheme for a pre-initialised object
+% doesn't change just because the default has changed.
+temp = PET.AcquisitionData();
+temp.set_storage_scheme('memory');
```
KrisThielemans, line 112: I'd be amazed if this changes the default. In C++,Python you need `PET.AcquisitionData.set_storage_scheme`

KrisThielemans:
> Keeping the old name makes it unclear as to whether you're returning the default scheme or the scheme for the current object.
> 
> I didn't change the naming on the python side because it handles both cases - if object is `None`, return default scheme, else return scheme for specific object.

I don't see the difference between C++ and Python in that respect. `PETAcquisitionData::storage_scheme()` has no object, so will return the default. `acq_data.storage_scheme()` will return the value for the object. That's what you do in Python as well: https://github.com/SyneRBI/SIRF/blob/7cde815e584354b75c7c44e0876eba97e2a0053f/src/xSTIR/pSTIR/tests/tests_one.py#L45

I agree that the `default_storage_scheme()` name might be better, but it does need some more changes in demos and UserGuide (ok, the latter will have to be updated in any case!).

In any case, I want to have the same names in Python/MATLAB/C++.

KrisThielemans:
@evgueni-ovtchinnikov is this PR still valid (I'm sure it'd need changes). If so, please assign to 3.1

evgueni-ovtchinnikov:
I am not convinced we need this (happily lived without it for 4+ years by now), so closing - but feel free to reopen if disagree, just try to make a stronger case for it.

Issue #662, opened 2020-05-19
Running SIRF examples on Travis?


KrisThielemans:
would it be possible to run the non-interactive Python demos on Travis. They tend to create figures, which probably will fail on travis, but maybe there's a way to just let this pass? It would allow us to do a minimal automatic test on them.

evgueni-ovtchinnikov:
tried with `fully_sampled_recon_single_chain.py` but only succeeded after sizeable amount of editing (there seem to be some docopt conflict with @casperdcl runner - had to hardcode options)

KrisThielemans:
Is there a travis job you could point us to? or some other info.

evgueni-ovtchinnikov:
job 2427 (passed)

evgueni-ovtchinnikov:
the new test is `test4.py` (in `pGadgetron/tests`)

KrisThielemans:
that test looks useful. thanks.

However, i guess I wasn't quite clear what I meant (sorry).

I was wondering if we can the *actual* demos to Travis. That way, we (somewhat) prevent the demo being out-of-date, and only discover problems when someone runs the demo by hand. Obviously, the demos are not intended for testing, and therefore most don't do any checking of output. However, just the fact that they run or not would be useful to know.

Problems that I'd anticipate are 
- matplotlib failing (as no X environment on Travis). Might be solvable by using `Agg` (via `.matplotlibrc`, see https://stackoverflow.com/questions/4931376/generating-matplotlib-graphs-without-a-running-x-server, or apparently `xvfb`, see https://stackoverflow.com/questions/35403127/testing-matplotlib-based-plots-in-travis-ci
- demos asking for input of the user, which happens with `show` I believe. Might be solvable by just taking input from `/dev/null` (choosing defaults).

This can be tested on the VM by doing `DISPLAY="" python demo.py </dev/null` I guess.

I certainly don't want to make the demos more complicated to cope with this. Maybe we do it only for a few where it doesn't create trouble.

KrisThielemans:
great. thank!

PR #723, opened 2020-06-19
Run ctest on SIRF demo scripts.

evgueni-ovtchinnikov:
- Trying again to resolve issue #662.

Travis OSX builds fail if PET demos are run by ctest, apparently because of the Travis 10 min limit for running tests - without PET demos all builds succeed, but every OSX build spends more than 500 sec on SIRF tests.

### DIFF
--- a/examples/Python/PET/run_all.py
+++ b/examples/Python/PET/run_all.py
@@ -24,6 +24,8 @@ import sys
 
 for i in sorted(glob.glob('*.py')):
     narg = len(sys.argv)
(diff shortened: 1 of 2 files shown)


paskino:
I guess we could uncomment these lines and see what happens? The issue was closed and now a `--noninteractive` argument can be passed (skips matplotlib) but the demos never ran as part of the testing https://github.com/SyneRBI/SIRF/blob/e0f53f517bb4ef8f04550b6bdf24f7ead3977a9d/src/xSTIR/pSTIR/tests/CMakeLists.txt#L24-L26

PR #725, opened 2020-06-22
add poisson noise to petAcqData

rijobro:
@KrisThielemans @evgueni-ovtchinnikov 
I don't know how you feel about this. Useful functionality, but slightly pollutes the class with simulation functionality. Relatively small though.

### DIFF
--- a/examples/Python/PET/sirf_sino_add_noise.py
+++ b/examples/Python/PET/sirf_sino_add_noise.py
@@ -17,8 +17,8 @@
 
 from sirf.Utilities import error
 import sirf.STIR as pet
-import numpy as np
 import argparse
+
 
 class Range(object):
     def __init__(self, start, end):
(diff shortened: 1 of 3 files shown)


KrisThielemans:
would work, but I believe better to use https://github.com/UCL/STIR/blob/e2f0bcb74af6bbf067785e07f58a8430faf3c64d/src/include/stir/GeneralisedPoissonNoiseGenerator.h
which means it'd be for C++ as well.

name `generate_poisson_realisation` would be better (Poisson noise isn't additive).

I don't think we'd want it in the class. Seems we then would want need it for MR data as well (with a weighted Gaussian) in the class, but there it will need weight info from somewhere (not sure where).

@johannesmayer, how do you do this?

johannesmayer:
Hey guys, I wrote a small class for this and I interfaced to stir for that on my simulation branch:

https://github.com/johannesmayer/SIRF/blob/d11cd493536f14bbece1d777b87df5e97a8df5dd/src/xDynamicSimulation/cDynamicSimulation/include/sirf/cDynamicSimulation/dynsim_noisegenerator.h#L45

KrisThielemans:
ok. I suggest to drop this PR then.

PR #731, opened 2020-06-24
improve string security

rijobro:


rijobro:
I was trying to improve our codacy score from B to A. However, I only did the first page. There are many more if you want to have a go @evgueni-ovtchinnikov !

KrisThielemans:
@evgueni-ovtchinnikov this seems a good thing. we should merge this soon.

PR #736, opened 2020-06-26
More useful error messages from argument validity checks in MATLAB, fixes #466

### LINKED ISSUE #466: Matlab bad class check
Accidentally called an `ImageDataProcessor` with a 3-D array as opposed to a `sirf.STIR.ImageData`.

However, the error check in the underlying code is:
```
assert(strcmp(input.class_name(), 'ImageData'))
```

A 3-D array doesn't have any dot indexing, (i.e., `3D_arr.class_name` is impossible), leading to an unrelated error:
```
Dot indexing is not supported for variables of this type.
```
I think something like:
```
assert(strcmp(class(gpu_mlem_obj), class(sirf.STIR.ImageData)))
```
would be better. 

Thoughts?

P.S. I haven't fixed this particular bug, because it could be in other places. Probably needs a search and replace.

evgueni-ovtchinnikov:


KrisThielemans:
2 Travis jobs failed but seems to have  been network problems. In any case, these files are not tested here, so merging.

PR #743, opened 2020-06-30
CPG -> DVF converter

rijobro:
Still WIP.

### DIFF
--- a/src/Registration/cReg/NiftiImageData3DDeformation.cpp
+++ b/src/Registration/cReg/NiftiImageData3DDeformation.cpp
@@ -71,10 +71,14 @@ void NiftiImageData3DDeformation<dataType>::create_cpp(NiftiImageData<dataType> 
 {
     // Need to copy because of const-ness
     NiftiImageData<dataType> ref_copy = ref;
-    nifti_image *cpp_ptr = cpg.get_raw_nifti_sptr().get();
-    reg_createControlPointGrid<dataType>(&cpp_ptr,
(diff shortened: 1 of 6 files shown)


rijobro:
Has `forward` and `backward` methods. Output can be created or returned (to save time).

rijobro:
I haven't tested yet, or exported to python. once in python, we can test adjoint-ness.

Also, this only uses NiftyReg functionality, not Supremo.

src/Registration/cReg/NiftiImageData3DDeformation.cpp
```diff
@@ -54,18 +54,29 @@ void NiftiImageData3DDeformation<dataType>::create_from_3D_image(const NiftiImag
...
                                    );
 }
 
+template<class dataType>
+void NiftiImageData3DDeformation<dataType>::create_cpp(NiftiImageData<dataType> &cpg, const NiftiImageData<dataType> &ref, float *spacingMillimeter) const
+{
+    // Need to copy because of const-ness
+    NiftiImageData<dataType> ref_copy = ref;
+    nifti_image *cpp_ptr = cpg.get_raw_nifti_sptr().get();
+    reg_createControlPointGrid<dataType>(&cpp_ptr,
```
KrisThielemans, line 75: Source is here
https://github.com/KCL-BMEIS/niftyreg/blob/f673b7837c0824f55dedb1534b32b55bf68a2823/reg-lib/cpu/_reg_localTrans.cpp#L19
This creates a NiftiImageData with correct spacing etc, but doesn't fill in any data I believe

PR #745, opened 2020-07-01
allow 1D&2D images to be used with NiftiImageData3D

rijobro:
means they can be used with registration.


rijobro:
@ckolbPTB hopefully this should do what you want.

PR #747, opened 2020-07-02
B spline2

rijobro:
Supersedes https://github.com/SyneRBI/SIRF/pull/743.

### DIFF
--- a/src/Registration/cReg/NiftiImageData3DBSpline.cpp
+++ b/src/Registration/cReg/NiftiImageData3DBSpline.cpp
@@ -44,7 +44,9 @@ NiftiImageData3DBSpline<dataType>::NiftiImageData3DBSpline(const NiftiImageData3
     bspline.GetDVFGradientWRTTransformationParameters(def.clone()->get_raw_nifti_sptr().get(), ref_ptr);
     // Get output
     nifti_image *cpg_ptr = bspline.GetTransformationAsImage();
+    cpg_ptr->intent_p1 = SPLINE_VEL_GRID;
     *this = NiftiImageData3DBSpline<dataType>(*cpg_ptr);
+    this->check_dimensions(NiftiImageData<dataType>::_3DBSpl);
 }
 
 template<class dataType>
--- a/src/Registration/cReg/NiftiImageData3DBSpline.cpp
+++ b/src/Registration/cReg/NiftiImageData3DBSpline.cpp
@@ -1,0 +1,92 @@
+/*
+SyneRBI Synergistic Image Reconstruction Framework (SIRF)
+Copyright 2017 - 2020 University College London
+
+This is software developed for the Collaborative Computational
+Project in Synergistic Reconstruction for Biomedical Imaging (formerly CCP PETMR)
+(http://www.ccpsynerbi.ac.uk/).
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+/*!
+\file
+\ingroup Registration
+\brief Class for deformation field transformations.
+
+\author Richard Brown
+\author SyneRBI
+*/
+
+#include "sirf/Reg/NiftiImageData3DBSpline.h"
+#include "sirf/Reg/NiftiImageData3DDeformation.h"
+#include "sirf/NiftyMoMo/BSplineTransformation.h"
+
+using namespace sirf;
+
+template<class dataType>
+NiftiImageData3DBSpline<dataType>::NiftiImageData3DBSpline(const NiftiImageData3DDeformation<dataType> &def, float spacing[])
+{
+    // Get any of the tensor components as a 3d image
+    nifti_image *ref_ptr = def.get_tensor_component(0)->get_raw_nifti_sptr().get();
+    // Create the NiftyMoMo bspline transformation class
+    NiftyMoMo::BSplineTransformation bspline(ref_ptr, 1, spacing);
+    // Convert DVF to CPG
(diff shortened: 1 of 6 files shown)


rijobro:
@KrisThielemans Not adjoint unforunately:

```
ControlPointGridToDeformationConverter is not adjoint, with normalized error of 1.2098765463674779 (max: 0.0001)
```
I don't know if we expected it to be or what we can do if it isn't.

src/Registration/cReg/ControlPointGridToDeformationConverter.cpp
```diff
@@ -0,0 +1,95 @@
...
+    return cpg.get_as_deformation_field(*_template_ref_sptr);
+}
+
+template<class dataType>
+NiftiImageData3DBSpline<dataType>
+ControlPointGridToDeformationConverter<dataType>::
+backward(const NiftiImageData3DDeformation<dataType> &dvf)
+{
+    check_is_set_up();
+    return NiftiImageData3DBSpline<dataType>(dvf, _spacing);
```
KrisThielemans, line 76: it doesn't make sense. The `backward` operation has to depend on the location of the CPG, but here it only depends on the spacing. You cannot know that the CPG is in the same location as the `dvf`. I suppose it usually is of course, but would at least have to check.

src/Registration/cReg/NiftiImageData3DBSpline.cpp
```diff
@@ -0,0 +1,97 @@
...
+*/
+
+#include "sirf/Reg/NiftiImageData3DBSpline.h"
+#include "sirf/Reg/NiftiImageData3DDeformation.h"
+#include "sirf/NiftyMoMo/BSplineTransformation.h"
+
+using namespace sirf;
+
+template<class dataType>
+NiftiImageData3DBSpline<dataType>::NiftiImageData3DBSpline(const NiftiImageData3DDeformation<dataType> &def, const float spacing[])
```
KrisThielemans, line 37: this is supposed to be the implementation of the `ControlPointGridToDeformationConverter::backward`. I wouldn't put it in the this constructor, but keep it there. Most people would expect this to do something smart with finding the best CPG that fits the `def`, but that's not what `backward` is supposed to do I believe.

Think about this is way: there's some matrix `B` corresponding to `dvf = converter.forward(cpg) = B cpg`. The backward operation is supposed to be `cpgnew = converter.backward(dvf) = B' cpg`. So, if we'd do them in sequence `cpgnew = B' B cpg`. There is no reason to believe that `B'B` is an identity matrix.

src/Registration/cReg/NiftiImageData3DBSpline.cpp
```diff
@@ -0,0 +1,97 @@
...
+{
+    // not marked const, so copy
+    float spacing_nonconst[3] = {spacing[0], spacing[1], spacing[2]};
+    // Get any of the tensor components as a 3d image
+    auto ref_sptr = def.get_tensor_component(0);
```
KrisThielemans, line 42: when moving this to the converter, this will have to be the ref of the CPG I believe

src/Registration/cReg/tests/test_cReg.cpp
```diff
@@ -1144,6 +1146,48 @@ int main(int argc, char* argv[])
...
+        // Do the same, using the converter
+        ControlPointGridToDeformationConverter<float> cpg_2_dvf_converter;
+        cpg_2_dvf_converter.set_cpg_spacing(spacing);
+        cpg_2_dvf_converter.set_reference_image(*dvf_sptr->get_tensor_component(0));
+        auto dvf_to_cpg_w_converter = cpg_2_dvf_converter.backward(*dvf_sptr);
+        auto dvf_to_cpg_to_dvf_w_converter = cpg_2_dvf_converter.forward(dvf_to_cpg_w_converter);
+
+        // Compare
+        if (dvf_to_cpg_to_dvf != dvf_to_cpg_to_dvf_w_converter)
+            throw std::runtime_error("ControlPointGridToDeformationConverter DVF->CPG->DVF failed.");
```
KrisThielemans, line 1185: Does this test work? It can't see why it should.In notation above, this is `BB' dvf`. Again, `BB'` shouldn't be identity.

KrisThielemans:
I didn't see an error really. sigh. Might be worth checking if NiftyMomo moved on.

I am confused by the hierarchy. For me, `NiftiImageData3DBSpline` should be a data object, and then something else (the "converter") that implements the `Model` (which should be derived from `Transformation`), but it's more important that it works of course.

KrisThielemans:
what about the following test to visualise this a bit more: construct the forward and backward matrices by hand, e.g.
```
mat=zeroes(cpg.num_elems(), dvf.num_elems())
cpg.fill(0)
for i in all_indices_in_cpg:
   cpg(i)=1
   dvf=converter.forward(cpg)
   for j in all_indices_in_dvf:
      mat(i,j)=dvf(j)
```
and similar for `backward`. Then we compare it with the transpose in `numpy`. Not sure how easy this is to do.

src/Registration/cReg/tests/test_cReg.cpp
rijobro: ah, bad error message. just checks that the result is the same whether the converter class is used or the methods inside of the `Deformation` and `BSpline` classes are called directly.

A couple of lines higher in the code, though, I do check that DVF->CPG->DVF does == DVF (checks `BB'` is identity). I agree that this doesn't have to be the case, but I suspected it would be very close to identity for a simple transformation, and where the CPG only has voxel sizes that are x2 the original DVF. It seemed like a good salt test to check the returned images weren't all zeroes, for example.

KrisThielemans: i think that if the `normalise_kernel=true`, you'd be getting close to an inverse (but still not  100%).

anywa,y the important thing is the `adjoint` test.

src/Registration/cReg/NiftiImageData3DBSpline.cpp
rijobro: constructor removed (and placed in `backward`).

src/Registration/cReg/NiftiImageData3DBSpline.cpp
rijobro: Now I use `_template_ref_sptr` for both the `forward` and `backward` methods.

Issue #641, opened 2020-04-27
occasional Bad file descriptor in cGadgetron


KrisThielemans:
This job https://travis-ci.org/github/SyneRBI/SIRF-SuperBuild/jobs/679959452#L16334
from https://github.com/SyneRBI/SIRF-SuperBuild/pull/377 (which is a `DEVEL` build) fails, while others are fine. The error is in the MR test
```
ERROR: test3.test_main
...
error: ??? "'write: Bad file descriptor' exception caught at line 545 of /Users/travis/build/SyneRBI/SIRF-SuperBuild/sources/SIRF/src/xGadgetron/cGadgetron/cgadgetron.cpp; the reconstruction engine output may provide more information"
-------------------- >> begin captured stdout << ---------------------
File: /Users/travis/build/SyneRBI/SIRF-SuperBuild/INSTALL/python/sirf/Gadgetron.py
Line: 1384
check_status found the following message sent from the engine:
```
I'll rerun the job, as I guess this won't happen again, but it is worrying nevertheless.

@evgueni-ovtchinnikov any ideas?

evgueni-ovtchinnikov:
I have seen this error message in Travis logs many times, but the reported error never ever happened locally, so impossible to investigate, I am afraid.

johannesmayer:
I get this one too from time to time

KrisThielemans:
hmmm. this is going to be tough then. Any ideas for writing some debugging checks and doing a special test-run with 1000 tests and see when it fails?

rijobro:
@evgueni-ovtchinnikov I haven't looked through the source code, but if this pertains to file writing, could you put it in a for loop? Similar to what you already do for trying to connect to the gadgetron server)?

```
bool success = false;
unsigned num_attempts = 5;
for (unsigned i=0; i<num_attempts; ++i) {
    try {
         success = do_the_thing_that_causes_the_error();
    }
    catch {}
    if (success) break;
}
if (!success)
    throw std::runtime_error("bad file descriptor");
```

evgueni-ovtchinnikov:
@johannesmayer: if you get this error when running your `mrtest.cpp`, then one possible culprit is your `MRAcquisitionData::read`, where you create `ISMRMRD::Dataset` and call its methods `readHeader`, `getNumberOfAcquisitions` and `readAcquisition` without `Mutex` locking/unlocking.

[...]

PR #757, opened 2020-07-05
[TRAVIS] add brew upgrade for OSX

### LINKED ISSUE #753: Travis OSX new failures (BLAS?)
since a few days we get failures on some OSX jobs, e.g. 2 most recent `master` jobs https://travis-ci.org/github/SyneRBI/SIRF/builds/704742645 and https://travis-ci.org/github/SyneRBI/SIRF/builds/704664740

Error seems to be
https://travis-ci.org/github/SyneRBI/SIRF/jobs/704742651#L16831
```
dlopen(/usr/local/lib/python3.7/site-packages/numpy/core/_multiarray_umath.cpython-37m-darwin.so, 2): Library not loaded: /usr/local/opt/gcc/lib/gcc/9/libgfortran.5.dylib
  Referenced from: /usr/local/opt/openblas/lib/libopenblas.0.dylib
  Reason: image not found
```
Some blas thing?

KrisThielemans:
Travis brew packages are sometimes out of date, causing problems with dependencies. Attempt therefore to call `brew upgrade`. I'm not sure if this should be done before installing anything else or after (currently trying the latter).

Hopefully fixes #753

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -181,6 +181,9 @@ before_install:
        #tar xzf cmake-*.tar.gz
        #mv cmake-*/CMake.app/Contents/* cmake
        #export PATH="$PWD/cmake/bin:$PATH"
+
+       #now upgrade the whole thing to avoid problems with outdated packages
+       brew upgrade
      )
    elif [[ "$TRAVIS_OS_NAME" == "linux" ]]; then
      PY_EXE=python$PYMVER


rijobro:
My impression from the [first comment](https://github.com/numpy/numpy/issues/14165#issuecomment-516563707) of the numpy issue you posed in #753 is that it's caused by a bug in a recent version of openblas that's not creating the correct symlink (`ln -s libopenblas.dylib libopenblas.0.dylib` would solve it). If that's correct, we'd have to look at rolling back the openblas version.

Anyway, hopefully this works and we won't have to think about it anymore 🤞

rijobro:
the failed OSX travis build looked like it was because of https://github.com/SyneRBI/SIRF/issues/641. I've restart builds, hopefully master will go all green.

PR #770, opened 2020-07-13
gmi not to use env variable

rijobro:
SIRF_PATH is not initially set (or outdated), so pass as variable


rijobro:
skipping CI, since matlab isn't built on travis anyway

PR #785, opened 2020-07-22
control open mp threads

rijobro:
@paskino merge this for OpenMP thread control. check the test for usage.


paskino:
I'll check the error in CI

PR #787, opened 2020-07-23
kernel convolution in python

rijobro:


rijobro:
worrying but unrelated travis problems, merging anyway.

Issue #793, opened 2020-08-16
STIR.get_linear_acquisition_model ignores sensitivities etc


KrisThielemans:
This function creates a new `AcquisitionModel` using the default constructor of the `type` I believe
https://github.com/SyneRBI/SIRF/blob/b63149f3a9163b9e82b76e6e649e00774cde45e7/src/xSTIR/pSTIR/STIR.py#L1472
This has multiple problems:
- It ignores any setting of the acquisition_sensitivity_model.
- It probably also ignores any other parameters of the acq_model that were set (such as the number of rays to use).
- More than likely it creates another instance of a STIR projection matrix, which can have severe memory consequences.

KrisThielemans:
It won't be so easy to solve the 2nd problem, as we don't have a way to copy the matrix nor GPU projectors). For the former, there's a similar bug in `get_matrix` preventing us to use that (#794))

KrisThielemans:
As this was undetected for so long, it transpires we have no test for this case

evgueni-ovtchinnikov:
I understand `get_linear_acquisition_model()` is needed in CIL only and so is `direct()`.

If so, why not have `direct()` forward without constant terms (the way I suggested) rather than have it as just an alias to `forward()`? We would not need this ugly `get_linear_acquisition_model()` then, would we?

KrisThielemans:
we had a lengthy discussion on this in  #273. I can't recall it all unfortunately. Here's an example comment summarising it a bit
https://github.com/SyneRBI/SIRF/pull/237#issuecomment-443464885

My main feeling is that it'll get CIL in trouble once we want to support (truly) non-linear models see #738. I believe CIL will still want to use `direct` for the forward operation. I haven't had a lot of input on #738 though...

PR #795, opened 2020-08-16
fix STIR.*.get_linear_acquisition_model 

KrisThielemans:
currently only partially addresses #793 (taking care of the acq_sens_model)


KrisThielemans:
need to add a condition to check if `asm` is set, as otherwise the `handle` statement breaks

KrisThielemans:
@evgueni-ovtchinnikov would you be able to have a look at this and resolve the 2 deficiencies (above comment on `asm`, and the TODO flagged in the code)

evgueni-ovtchinnikov:
Fixed by PR #809.

PR #800, opened 2020-08-26
Replaces throwing error in binary operations with returning NotImplemented

evgueni-ovtchinnikov:


KrisThielemans:
ok, so I guess the `try` will get into `catch` because `other.real` and `other.imag` don't exist for `CIL.BlockDataContainer`. looks good to me.

KrisThielemans:
great. no errors! @evgueni-ovtchinnikov can you do this for the other binary operators as well? Feel free to merge then once Travis is happy.

paskino:
I think in case of not a SIRF DataContainer (and derivatives) it will throw an `AssertionError` at the beginning.

KrisThielemans:
> Unfortunately python will accept anything as `other` and type checking needs to happen inside the method itself.
> I'm suggesting to use something along these lines:
> 
> ```python
> if not issubclass(other, sirf.DataContainer):
>     raise NotImplemented
> ```
(that would have to be `return  NotImplemented` of course.)

This sounds in principle ok to me, but the code inside the `try/catch` handles scalars, and possibly even numpy arrays (not sure).

We need some kind of check on what data is supported. Not sure what that has to be...

paskino:
Well `issubclass` will accept a `tuple` of classinfo's so we could do something like this?

```python 
from numbers import Number

if not issubclass(other, (sirf.DataContainer, Number, np.ndarray)):
    return NotImplemented
```

KrisThielemans:
looks good to me!

KrisThielemans:
excellent. this looks good to me. Travis currently happy as well, but still some jobs running.

I believe we need to do the same for `__add__` etc. Then good to go as far as I can see.

PR #802, opened 2020-09-03
fix to algebra methods

### LINKED ISSUE #799: Division of a BlockDataContainer of ImageData by a non-zero scalar gives NaN pixels
I believe this is related to Issue #798 but would like to flag it independently in case it's not.

If I define a `CIL.BlockDataContainer` made of `sirf.ImageData` blocks and divide it by a non-zero scalar, some (but not all) pixels are put to NaN.

```python
ImageData = am.domain_geometry()
ImageDatax3 = BlockGeometry(*[ImageData for _ in range(3) ] )
imagex3 = ImageDatax3.allocate('random')
sigma = 1.

print('Image has shape {}, that is {} pixels'.format(imagex3[0].as_array().shape, imagex3[0].as_array().size))

for image in imagex3 / sigma:

    print('New image has shape {}, that is {} pixels'.format(image.as_array().shape, image.as_array().size))
    print('Out of these, {} pixels are NaN'.format(np.isnan(image.as_array()).sum()))
```

outputs

```python
Image has shape (127, 127, 127), that is 2048383 pixels
New image has shape (127, 127, 127), that is 2048383 pixels
Out of these, 2722 pixels are NaN
New image has shape (127, 127, 127), that is 2048383 pixels
Out of these, 2722 pixels are NaN
New image has shape (127, 127, 127), that is 2048383 pixels
Out of these, 2722 pixels are NaN
```

paskino:
closes #799

### DIFF
--- a/src/Registration/pReg/Reg.py
+++ b/src/Registration/pReg/Reg.py
@@ -520,6 +520,13 @@ class NiftiImageData(SIRF.ImageData):
         pyiutil.deleteDataHandle(h)
         return value
 
+    @property
+    def shape(self):
+        return self.dimensions()
+    @property
+    def dtype(self):
+        return numpy.float32
+
 
 class NiftiImageData3D(NiftiImageData):
     """Class for 3D nifti image data."""
--- a/src/xGadgetron/pGadgetron/Gadgetron.py
+++ b/src/xGadgetron/pGadgetron/Gadgetron.py
@@ -474,6 +474,8 @@ class ImageData(SIRF.ImageData):
         try_calling(pygadgetron.cGT_print_header(self.handle, im_num))
     @property
     def dtype(self):
+        if self.is_real():
+            return numpy.float32
         return numpy.complex64
     @property
     def shape(self):
--- a/src/xGadgetron/pGadgetron/Gadgetron.py
+++ b/src/xGadgetron/pGadgetron/Gadgetron.py
@@ -472,7 +472,10 @@ class ImageData(SIRF.ImageData):
     def print_header(self, im_num):
         """Print the header of one of the images. zero based."""
         try_calling(pygadgetron.cGT_print_header(self.handle, im_num))
-
+    @property
+    def dtype(self):
+        return numpy.float32
+        
 SIRF.ImageData.register(ImageData)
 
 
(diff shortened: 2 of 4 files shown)


src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -472,7 +472,10 @@ def allocate(self, value=0, **kwargs):
     def print_header(self, im_num):
         """Print the header of one of the images. zero based."""
         try_calling(pygadgetron.cGT_print_header(self.handle, im_num))
-
+    @property
+    def dtype(self):
+        return numpy.float32
```
evgueni-ovtchinnikov, line 477: need to check `is_real()`, and if `False`, return `numpy.complex64`

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -969,6 +972,12 @@ def allocate(self, value=0, **kwargs):
             tmp = value * numpy.ones(out.as_array().shape)
             out.fill(tmp)
         return out
+    @property
+    def shape(self):
+        return self.dimensions()
+    @property
+    def dtype(self):
+        return numpy.float32
```
evgueni-ovtchinnikov, line 980: actually, `return numpy.complex64`

src/Registration/pReg/Reg.py
```diff
@@ -520,6 +520,13 @@ def are_equal_to_given_accuracy(im1, im2, accuracy):
         pyiutil.deleteDataHandle(h)
         return value
 
+    @property
+    def shape(self):
+        return self.dimensions()
+    @property
+    def dtype(self):
+        return numpy.float32
+
```
evgueni-ovtchinnikov, line 529: Would it not be enough to have this up in `DataContainer`, `dtype` to be overrided by MR `dtype`?

Same for all PET containers of course.

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -472,7 +472,13 @@ def allocate(self, value=0, **kwargs):
     def print_header(self, im_num):
         """Print the header of one of the images. zero based."""
         try_calling(pygadgetron.cGT_print_header(self.handle, im_num))
-
+    @property
+    def dtype(self):
+        return numpy.complex64
```
evgueni-ovtchinnikov, line 477:
```
if self.is_real():
    return numpy.float32
else:
    return numpy.complex64
```

Issue #798, opened 2020-08-25
Compatibility between SIRF's and CIL's DataContainer


ClaireDelplancke:
I would like to multiply an object of type `sirf.STIR.ImageData` by an object of type `ccpi.framework.BlockDataContainer.BlockDataContainer` where each DataContainer inside BlockDataContainer are again of type `sirf.STIR.ImageData`. This should give me a BlockDataContainer. Instead, I've got the following error
```python
AssertionError                            Traceback (most recent call last)
<ipython-input-6-215be633e2e3> in <module>()
----> 1 image.multiply(imagex3)

/home/sirfuser/devel/install/python/sirf/SIRF.pyc in multiply(self, other, out)
    132             other = self.copy()
    133             other.fill(tmp)
--> 134         assert_validities(self, other)
    135         if out is None:
    136             out = self.same_object()

/home/sirfuser/devel/install/python/sirf/Utilities.pyc in assert_validities(x, y)
    394     except AssertionError as ae:
    395         raise AssertionError('Expecting same type input, got {} and {}'.format(type(x), 
--> 396                                                                                type(y)))
    397     if x.handle is None:
    398         raise AssertionError('handle for first parameter is None')

AssertionError: Expecting same type input, got <class 'sirf.STIR.ImageData'> and <class 'ccpi.framework.BlockDataContainer.BlockDataContainer'>
```

As per @paskino this comes from the fact that [DataContainer in CIL](https://github.com/vais-ral/CCPi-Framework/blob/master/Wrappers/Python/ccpi/framework/framework.py#L467) has got a `__container_priority__` member but [DataContainer in SIRF](https://github.com/SyneRBI/SIRF/blob/master/src/common/SIRF.py#L49) doesn't.

evgueni-ovtchinnikov:
the error message says it all: you cannot multiply objects of different type

if `ccpi.framework.BlockDataContainer.BlockDataContainer` contains `sirf.STIR.ImageData` objects, then you need to multiply each of them by your `sirf.STIR.ImageData` object in a loop

ClaireDelplancke:
Sure, but I would like to be able to do it! 
For example I need it in order to use [MixedL21Norm](https://github.com/vais-ral/CCPi-Framework/blob/master/Wrappers/Python/ccpi/optimisation/functions/MixedL21Norm.py#L83) implemented in CIL.

evgueni-ovtchinnikov:
[...]

PR #803, opened 2020-09-08
Add max python

### LINKED ISSUE #797: ImageData has no method max
CIL's  [MixedL21Norm convex conjugate](https://github.com/vais-ral/CCPi-Framework/blob/master/Wrappers/Python/ccpi/optimisation/functions/MixedL21Norm.py#L77) expects `x.max()` to give the element-wise maximum of `x`, but if `x` is of ImageData type, it has no method max.

paskino:
closes #797 
addresses #798 for [addition ](github.com/SyneRBI/SIRF/blob/add_max_python/src/common/SIRF.py#L174)and [subtraction](https://github.com/SyneRBI/SIRF/blob/add_max_python/src/common/SIRF.py#L274) and [multiplication](https://github.com/SyneRBI/SIRF/blob/add_max_python/src/common/SIRF.py#L302).

Unfortunately the implementation is different in each method and should be harmonised.

Unittest are currently available as part of [CIL](https://github.com/vais-ral/CCPi-Framework/blob/test_SIRF/Wrappers/Python/test/test_SIRF.py#L150L201) but they really should be part of SIRF.

### DIFF
--- a/src/common/SIRF.py
+++ b/src/common/SIRF.py
@@ -64,7 +64,7 @@ class DataContainer(ABC):
             (mn_one.ctypes.data, self.handle, zero.ctypes.data, self.handle)
         check_status(z.handle)
         return z
-#    @abc.abstractmethod
+
     def same_object(self):
         '''
         Returns an object of the same type as self.
@@ -171,6 +171,8 @@ class DataContainer(ABC):
         other: DataContainer
         out:   DataContainer to store the result to.
         '''
+        if not isinstance (other, ( DataContainer , Number, int, float, numpy.float32 )):
+            return NotImplemented
         if isinstance(other , ( Number, int, float, numpy.float32 )):
             tmp = other + numpy.zeros(self.shape, self.dtype)
             other = self.copy()
@@ -267,7 +269,11 @@ class DataContainer(ABC):
         data viewed as vectors.
         other: DataContainer
         '''
-        return self.subtract(other)
+        assert self.handle is not None
+
+        if isinstance(other, (Number, int, float, numpy.float32) ):
+            return self.subtract(other)
+        return NotImplemented
 
     def __mul__(self, other):
         '''
--- a/src/common/SIRF.py
+++ b/src/common/SIRF.py
@@ -271,7 +271,7 @@ class DataContainer(ABC):
         '''
         assert self.handle is not None
 
-        if isinstance(other, (Number, int, float, numpy.float32) ):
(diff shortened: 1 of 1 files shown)


KrisThielemans:
that seems to have gone wrong:
```
  File "/home/travis/build/SyneRBI/SIRF/src/xSTIR/pSTIR/tests/tests_four.py", line 40, in test_main
    diff = new_acq_data - acq_data
TypeError: unsupported operand type(s) for -: 'AcquisitionData' and 'AcquisitionData'
```

src/common/SIRF.py
```diff
@@ -171,6 +171,8 @@ def add(self, other, out=None):
         other: DataContainer
         out:   DataContainer to store the result to.
         '''
+        if not isinstance (other, ( DataContainer , Number, int, float, numpy.float32 )):
```
evgueni-ovtchinnikov, line 174: int, float and numpy.float32 are derived from Number, are they not?

src/common/SIRF.py
```diff
@@ -267,7 +269,11 @@ def __sub__(self, other):
         data viewed as vectors.
         other: DataContainer
         '''
-        return self.subtract(other)
+        assert self.handle is not None
+
+        if isinstance(other, (Number, int, float, numpy.float32) ):
```
evgueni-ovtchinnikov, line 274: DataContainer is missing

evgueni-ovtchinnikov (changes requested):

src/common/SIRF.py
paskino: I suppose `int` and `float` may be, but I wouldn't be sure about `numpy.float32`.

evgueni-ovtchinnikov: >>> import numpy
>>> a = numpy.ones((3,), dtype=numpy.float32)
>>> from numbers import Number
>>> print(isinstance(a[0], Number))
True

paskino: very well then

paskino:
@evgueni-ovtchinnikov once travis is happy I will merge, OK?

evgueni-ovtchinnikov:
ok

paskino:
@KrisThielemans @evgueni-ovtchinnikov the build was successful, only the last item in the travis list didn't succeed. It didn't finish the tests, however it built fine. 

Shall I merge anyhow?

evgueni-ovtchinnikov:
I vote you merge

KrisThielemans:
sure

Issue #772, opened 2020-07-13
SIRF error


paskino:
SIRF Python: I am allocating an `ImageData` using the `allocate`. In CIL we can allocate just the memory, without setting it. I have done the same with a SIRF `DataContainer` but it fails. What I see is 

```
python ../../../../buildVM/sources/SIRF/examples/Python/PETMR/PET_MCIR.py -T "fwd_tm_ms_*.txt" -t tm -S "raw_FDG_ms*hs"  -r "FGP_TV" -i 100 -s 100 --visualisations -v 0 -o spdhg --algorithm="spdhg"
Setting up acquisition models...
Setting up reconstructor...
Will output results to: spdhg.log
SPDHG setting up
wrong fill value. Should be ImageData, numpy.ndarray, float or int
```

Notice that the program ends without raising an exception but just printing to screen. So I don't have a clue where it all went wrong. There's a clue: it is a `fill` method, either [`ImageData.fill`](https://github.com/SyneRBI/SIRF/blob/master/src/common/SIRF.py#L567) or [`AcquisitionData`](https://github.com/SyneRBI/SIRF/blob/master/src/xSTIR/pSTIR/STIR.py.in#L729). Turns out to be the [second](https://github.com/SyneRBI/SIRF/blob/master/src/xSTIR/pSTIR/STIR.py.in#L759).

In this particular case, we could change the `raise error("wrong fill value")` call with a 
```python

raise TypeError("Wrong fill value. Expecting ImageData, numpy.ndarray, float or int, got {}"\
.format(type(value))
```

I'd suggest that we use standard Python error types in our calls and create a new exception for errors that come from the engine: `EngineError`

rijobro:
Is it also true that we shouldn't put our `main` in a try and catch, as this will hide the error stack:

```
try:
    main()
    print('done')
    exit(0)
except error as err:
    # display error information
    print('%s' % err.value)
    exit(1)               # <------ most of our scripts don't have this either
```

From what I understand, it should just be this:

```
if __name__ == "__main__":
    main()
```

evgueni-ovtchinnikov:
@paskino: it looks like you wanted to call `ImageData.allocate()` with a string `value` (e.g. 'random') but made some spelling mistake - I can see no other reason for such output. Can I see your script that produced it?

paskino:
It's buried inside the CIL where we `allocate(None)` to just allocate the memory. 

For the time I've changed `allocate` to handle `value=None` as `value=0`. 
[...]

PR #807, opened 2020-09-29
Added is_empty method to PETAcquisitionData to detect templates

evgueni-ovtchinnikov:
Related to #772


KrisThielemans:
I'm not sure about this. I have no idea what `items` is supposed to return (it doesn't seem to be documented, and the name isn't very clear).

Issue #816, opened 2020-11-04
sirf.Reg.compose_single_deformation check on reference is too restrictive


KrisThielemans:
The check https://github.com/SyneRBI/SIRF/blob/0721ed9332b295e11529d37aed7ce7795815157b/src/Registration/pReg/Reg.py#L720
fails if an image is read with `sirf.Reg.NiftiImageData` as opposed to `sirf.Reg.NiftiImageData3D`. 

What about putting a step in between attempting to convert it first
```python
ref=sirf.Reg.NiftiImageData3D(ref)
```
(could catch exception if the conversion fails, but not sure if that'd help).

However, I've tried to construct a `sirf.Reg.NiftiImageData3D` from a `sirf.Reg.NiftiImageData3DDisplacement`. Surprisingly, this works without any trouble, although then the number of dimensions is 4
```python
sirf.Reg.NiftiImageData3D(Disp_MRAC_to_ref).as_array().shape
  (208, 208, 128, 1, 3)
```
Should the `NiftiImageData3D` check the dimensions?

rijobro:
The first solution looks good to me -- if not `NiftiImageData3D`, try to convert and else raise an error.

As for the second, that seems entirely weird: [here](https://github.com/SyneRBI/SIRF/blob/master/src/Registration/cReg/NiftiImageData.cpp#L549) you can see the checks done on the  3D, one of the ones being that `ndim==3`, which isn't the case here... Can you do `sirf.Reg.NiftiImageData3D(Disp_MRAC_to_ref).print_header()` and see what that gives?

KrisThielemans:
I can't do `print_header` at the moment as it's running from Visual Studio Code, so hides stdout. But I can do the following
```
disp.get_dimensions()
   array([  5, 344, 344, 127,   1,   3,   1,   1], dtype=int32)
tmp=sirf.Reg.NiftiImageData3D(disp)
tmp.dimensions()
    (344, 344, 127, 1, 3)
tmp.get_dimensions()
    array([  5, 344, 344, 127,   1,   3,   1,   1], dtype=int32)
tmp.as_array().shape
  (344, 344, 127, 1, 3)
```
I also wrote both to file (`tmp.write('tmp.nii')` etc) and then print headers. They seem identical (which isn't unexpected)
```
$ sirf_print_nifti_info tmp*nii

Printing info for 2 nifti image(s):
        analyze_75_orient: 0                  0                  
        analyze75_orient:  0                  0                  
[...]

PR #817, opened 2020-11-06
if NiftiImageData3D, run check

rijobro:
Fixes https://github.com/SyneRBI/SIRF/issues/816.

@KrisThielemans -- Haven't compiled/tested but hopefully the logic should be clear here. The problem was that the python constructors for `NiftiImageData` and `NiftiImageData3D` but used the C++ `NiftiImageData` constructor, so there were no checks taking place.

### DIFF
--- a/src/Registration/cReg/cReg.cpp
+++ b/src/Registration/cReg/cReg.cpp
@@ -450,13 +450,16 @@ void* cReg_NiftiImageData_get_inner_product(const void* im1_ptr, const void* im2
     CATCH;
 }
 extern "C"
-void* cReg_NiftiImageData_from_SIRFImageData(void* ptr)
+void* cReg_NiftiImageData_from_SIRFImageData(void* ptr, const int is_3D)
 {
 	try {
         ImageData& sirf_im = objectFromHandle<ImageData>(ptr);
-        std::shared_ptr<NiftiImageData<float> >
-            sptr(new NiftiImageData<float>(sirf_im));
-        return newObjectHandle(sptr);
+
+    	// Use an int as if bool to optionally run 3D checks
+        if (is_3D == 1)
+        	return newObjectHandle(sptr(new NiftiImageData3D<float>(sirf_im)));
+        else
+        	return newObjectHandle(sptr(new NiftiImageData<float>(sirf_im)));
     }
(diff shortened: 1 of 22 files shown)


src/Registration/cReg/cReg.cpp
```diff
@@ -447,13 +447,16 @@ void* cReg_NiftiImageData_get_inner_product(const void* im1_ptr, const void* im2
...
+void* cReg_NiftiImageData_from_SIRFImageData(void* ptr, const int is_3D)
 {
-	try {
+    try {
         ImageData& sirf_im = objectFromHandle<ImageData>(ptr);
-        std::shared_ptr<NiftiImageData<float> >
-            sptr(new NiftiImageData<float>(sirf_im));
-        return newObjectHandle(sptr);
+
+        // Use an int as if bool to optionally run 3D checks
```
KrisThielemans, line 455: cryptic!

KrisThielemans: I maybe I understand, maybe we can't wrap `bool`? fine. I'd put this in the doxygen documentation  of the wrapper then

KrisThielemans (changes requested):

src/Registration/cReg/cReg.cpp
```diff
@@ -447,13 +447,16 @@ void* cReg_NiftiImageData_get_inner_product(const void* im1_ptr, const void* im2
...
+        if (is_3D == 1)
```
KrisThielemans, line 456:
```suggestion
        if (is_3D)
```

KrisThielemans:
looks good. I'll try this later I guess and merge then. thx!

src/Registration/cReg/cReg.cpp
rijobro: Yeah, as far as I know we can pass `int` without any wrapping, but I don't think the same is true for booleans.

src/Registration/cReg/cReg.cpp
rijobro: Can be, I put it in to be explicit as I was using `int` instead of `bool`. Feel free to commit the suggestion.

KrisThielemans (approved):

PR #821, opened 2020-11-16
Pet AcquisitionModel subset

paskino:
Adds a way to specify which subset a PET `AcquisitionModel` should forward/backward project.


src/xSTIR/pSTIR/STIR.py
```diff
@@ -507,8 +512,9 @@ def allocate(self, value=0, **kwargs):
                 out.fill(numpy.random.random_sample(shape))
             elif value == 'random_int':
                 max_value = kwargs.get('max_value', 100)
-                out.fill(numpy.random.randint(
-                    max_value, size=shape))
+                out.fill(numpy.random.randint(max_value,size=shape))
+        elif value is None:
+            out = self.get_uniform_copy(0)
```
evgueni-ovtchinnikov, line 517: if you just want to allocate memory, and `self` is not an empty template (`self.is_empty()` is `False`), then you can do `out = self.clone()`.

If your branch `DataContainer` does not have `is_empty()`, merge master.

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1505,42 +1516,34 @@ def get_linear_acquisition_model(self):
...
-    def adjoint(self, ad, subset_num=0, num_subsets=1, out=None):
-        """Back-project acquisition data into image space.
-
-        Only if the AcquisitionModel is linear
-
-        Added for CCPi CIL compatibility
-        https://github.com/CCPPETMR/SIRF/pull/237#issuecomment-439894266
-        """
-        if not self.is_linear():
+    def direct(self, image, out = None):
```
evgueni-ovtchinnikov, line 1519: as I have learnt belatedly, it is considered good Python style not to have spaces around `=` when setting the argument default value (I know we have those spaces in many places, but let us at least not add more)

evgueni-ovtchinnikov (changes requested):

src/common/SIRF.py
```diff
@@ -52,6 +52,20 @@ class DataContainer(ABC):
     '''
     def __init__(self):
         self.handle = None
+        self._mask = None
```
paskino, line 55: references to `mask` should be removed

src/xSTIR/pSTIR/STIR.py
paskino: if self is a template would `get_uniform_copy` return a `DataContainer`?

src/common/SIRF.py
```diff
@@ -205,13 +206,21 @@ def axpby(self, a, b, y, out=None, **kwargs):
 
         Returns the sum of the container data with another container 
         data viewed as vectors.
+        a: multiplyer to self, can be a number or a DataContainer
+        b: multiplyer to y, can be a number or a DataContainer
```
KrisThielemans, line 210:
```suggestion
        b: multiplier to y, can be a number or a DataContainer 
```

KrisThielemans (changes requested):
Overall looks fine, but the `subset` `num_subsets `members are ignored for `forward` , `backward` which will I think will be quite confusing for the user. What about doing
```python
forward(self, image, subset_num=self.subset_num, num_subsets=self.num_subsets, out=None)
```
if Python allows that?

This also needs some lines in `CHANGES.md` and more than likely also an adaptation of the UserGuide.

paskino:
We discussed with @evgueni-ovtchinnikov @gfardell @epapoutsellis about the proposed new signature. We agreed to use

```python

def forward(self, image, subset_num=None, num_subsets=None, out=None)
```
This will be clearer for the user as it will tell them that there is a default and they don't have to worry about it. I added docstrings to document this change.

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1581,15 +1607,29 @@ def domain_geometry(self):
 
     @property
     def subset_num(self):
+        '''selected subset number used for forward and back projection
+        
+        default value is 0 and corresponds to forward/backward projecting
+        the whole dataset.
```
KrisThielemans, line 1638: this is incorrect

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1476,12 +1476,25 @@ def set_acquisition_sensitivity(self, asm):
...
         """Return the forward projection of image.
 
-        image   :  an ImageData object.
+        image      :  an ImageData object.
+        subset_num : optional (int) subset number to forward project. 
+                     Default None. If None subset_num is the instance 
+                     property subset_num.
+        num_subsets: optional (int) number of subsets the data is divided 
+                     in for the projection. Default None. 
+                     If None num_subsets is the instance property num_subsets.
```
KrisThielemans, line 1513: can we formulate this in a less Pythonic-way? Most of our audience will not understand "instance property"

CHANGES.md
```diff
@@ -1,6 +1,7 @@
 # ChangeLog
 
 ## vX.X.X
+* STIR AcquisitionModel forward and backward signature has changed: default value of `num_subset` and `subset_num` is `None` and the instance properties `num_subset` and `subset_num` will be used in such case. By default these are 0 and 1 respectively, so if the user doesn't change them the behaviour is unchanged.
```
KrisThielemans, line 4:
```suggestion
* STIR AcquisitionModel `forward`, `direct`, `backward` and `adjoint` signatures have changed in Python. Subset information should now be set via `num_subsets` and `subset_num` members. `The `forward` and `backward` members can still be called with the previous syntax but this will be removed in a later version.
Note that default values of `num_subsets` and `subset_num` are 0 and 1 respectively, such that default behaviour is default behaviour (i.e. process all data) is unchanged.
```

KrisThielemans (changes requested):
Strategy is good.

Please add a corresponding update for MATLAB at #555. We will also need to have this in C++.

3 last commits are attributed to `unsystemizer`. Please correct your settings and ideally change the commit author (will need a force push).

src/xSTIR/pSTIR/STIR.py
paskino: I understand this request but I believe it's better to be clear and right on the programming point of view. I could add "member" in parenthesis.

src/xSTIR/pSTIR/STIR.py
paskino: what is not correct?

paskino:
regarding the author on the commits, I have tried to update it according to [this thread](https://stackoverflow.com/questions/3042437/how-to-change-the-commit-author-for-one-specific-commit) but I didn't manage.

At any rate I'll do a squash merge which should leave myself as author, I hope that's OK.

src/xSTIR/pSTIR/STIR.py
evgueni-ovtchinnikov: this is only true if `num_subsets` is 1, otherwise subsets other than the first are not projected

paskino: I still don't really understand what is wrong, but I tried to rephrase it.

src/xSTIR/pSTIR/STIR.py
evgueni-ovtchinnikov: you could write e.g. "the number of subsets is set to `self.num_subsets`"

src/xSTIR/pSTIR/STIR.py
evgueni-ovtchinnikov: if `num_subsets` is 2, and `subset_num` is 0, then subset 0 is projected and subset 1 is not

just drop what you wrote after "default value is 0"

Issue #831, opened 2020-11-29
bug when concatenating affine transformations?


rijobro:
We are getting bad results with `O = R^-1 * T * R`, where `R` is a transformation matrix that maps the MR to PET gantry shift, `T` is a deformation in MR coordinates, and `O` is a deformation in PET coordinates.

In SIRF, we convert all transformations to deformations and then concatenate them. 

This was giving bad results. I checked with NiftyReg and that seemed to give expected results. The difference, here, seems to be that a reference image is only used when converting transformation matrices to deformations. It is **not** used when concatenating deformations or displacements. This makes sense to me, I'm not sure why I modified the previous behaviour in this commit: https://github.com/SyneRBI/SIRF/commit/48ed3a60cf123e5fe432589fa2c2f1ded72b77cb.

PR incoming to reverse previous commit.

rijobro:
Edit: We should use the reference when composing displacements/deformations before using them with resampling, but shouldn't use them otherwise.

KrisThielemans:
> We should use the reference when composing displacements/deformations before using them with resampling, but shouldn't use them otherwise.

can't say I understand that sentence... I never understood why you'd need to use the reference image at all when composing transformations, or even defining/using them.
- A transformation matrix is independent of the "reference" sampling
- A def/disp field is defined in a certain space which I guess is the "reference" anyway.

I can see that resampling with a transf matrix needs a "reference" (i.e. the template image what it needs to resample to). Resampling with a def/disp field has a natural "reference" (but if you'd use another one, you can interpolate it accordingly).

rijobro:
Ideally I would have it such that if all transforms to be concatenated were matrices, you would just do matrix multiplication. 

If there was a combination of matrices and disps/defs, then I guess you should just be able to update the `sto_ijk` and `qto_ijk` matrices. But I that's not how the [reg_transform](https://github.com/KCL-BMEIS/niftyreg/blob/master/reg-apps/reg_transform.cpp#L577) executable works, which is what I based it on.

rijobro:
From the hyperlink in the previous comment, you should see that only the affine transformations use the reference image, whereas in the [resample executable](https://github.com/KCL-BMEIS/niftyreg/blob/master/reg-apps/reg_resample.cpp#L323) all transformations gets modified to match the reference image.

[...]

PR #832, opened 2020-11-29
do not modify disps/defs when converting them to defs

rijobro:
fixes https://github.com/SyneRBI/SIRF/issues/831


rijobro:
old implementation (left) versus corrected new implementation (right):
image

KrisThielemans:
@rijobro it seems we forgot about this one. Any reason why we shouldn't merge this? (I seem to remember I thought there was a problem somewhere, but cannot find anything, so maybe I was dreaming!)

rijobro:
@KrisThielemans I think this comment was your reservation: https://github.com/SyneRBI/SIRF/issues/831#issuecomment-735444474.

I'm ok for this to be merged.

PR #840, opened 2020-12-26
[CMake] Use ISMRMRD::ISMRMRD target also on Windows

KrisThielemans:
We were still using the old lines, but they break compilation when both Debug and Release versions of HDF5 libraries are installed.
In any case, the new code is much simpler!

### DIFF
--- a/src/Registration/cReg/SPMRegistration.cpp
+++ b/src/Registration/cReg/SPMRegistration.cpp
@@ -21,7 +21,7 @@ limitations under the License.
 /*!
 \file
 \ingroup Registration
-\brief NiftyReg's aladin class for rigid and affine registrations.
+\brief Implementation for sirf::SPMRegistration
 
 \author Richard Brown
 \author SyneRBI
--- a/src/xGadgetron/cGadgetron/CMakeLists.txt
+++ b/src/xGadgetron/cGadgetron/CMakeLists.txt
@@ -60,21 +60,7 @@ else()
   target_link_libraries(cgadgetron PUBLIC Boost::system Boost::filesystem Boost::thread Boost::date_time Boost::chrono)
 endif()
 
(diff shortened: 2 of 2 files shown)


KrisThielemans:
@evgueni-ovtchinnikov I'm merging this as we need it now... I hope it works for you

evgueni-ovtchinnikov:
All MATLAB tests and demos run ok, Python and C++ ones crash, apparently on reading ISMRMRD acquisitions Dataset header.

KrisThielemans:
@evgueni-ovtchinnikov said
> C++ test of Johannes crashes at line 99 of gadgetron_data_containers.cpp, Python demos (e.g. acquisition_data.py crash at line 724.

KrisThielemans:
@evgueni-ovtchinnikov did you attempt to build both MATLAB and Python support? That will fail because of HDF5 problems I believe. You need to build a Python-only version (`DISABLE_MATLAB=ON`).

If this isn't it, please give more info on your set-up and how you tested this

evgueni-ovtchinnikov:
Rebuilt SIRF with MATLAB disabled - still the same crashes.

I remember your saying this was because of the Python's HDF5 conflicting with installed separately HDF5.

KrisThielemans:
yes, conflicts between libraries at compilation time and DLLs at run-time are the most likely explanation.

I'm afraid I'd need more detail for on the simple statement "rebuild SIRF with MATLAB disabled".  Are you running CMake for SIRF or the SIRF-SuperBuild? Did you build in a new directory (i.e. no CMake cache)? Maybe something to discuss on Thursday?

evgueni-ovtchinnikov:
I have just built SIRF with SIRF-SuperBuild afresh in a new folder with MATLAB build disabled - MR scripts still crash.

PR #844, opened 2020-12-28
PET tests tolerance fixes

### LINKED ISSUE #843: PET tests can fail due to accuracy in recorded values
`pTest` normally compares with recorded values as opposed to known true values. Those recordings can be a bit wrong due to numerical precision.  This makes the tests with another compiler/processor more prone to failure.

I experience this on Windows with Visual Studio 2019.

KrisThielemans:
Fixes #843


evgueni-ovtchinnikov (changes requested):
looks fine to me, just one small stylistic suggestion: I would use `within_tolerance` rather than `with_tolerance`

KrisThielemans:
@evgueni-ovtchinnikov  Great idea! want to make that change yourself? You should be able to push to this branch (have to add https://github.com/KrisThielemans/SIRF as remote).  Or I can do it later. let me know.

evgueni-ovtchinnikov:
@KrisThielemans: I am fixing #842 at the moment, which will affect your fixes, so will deal with this PR a bit later.

PR #846, opened 2021-01-04
Extra throws in pTest removed

### LINKED ISSUE #842: pTest problem with __del__ and exceptions

evgueni-ovtchinnikov:
aims to fix #842

### DIFF
--- a/src/common/Utilities.py
+++ b/src/common/Utilities.py
@@ -231,14 +231,16 @@ def show_3D_array\
     return 0
 
 
-def check_tolerance(expected, actual, abstol=0, reltol=1e-4):
+def check_tolerance(expected, actual, abstol=0, reltol=2e-3):
     '''
-    Check if 2 floats are equal up to a tolerance
-    Throws an error if abs(expected - actual) > abstol + reltol*abs(expected)
+    Check if 2 floats are equal within the specified tolerance, i.e.
+    abs(expected - actual) <= abstol + reltol*abs(expected).
+    Returns an error string if they are not and None otherwise.
     '''
-    if abs(expected - actual) > abstol + reltol*abs(expected):
-        raise ValueError("|%.4g - %.4g| > %.3g" %
-                         (expected, actual, abstol + reltol*abs(expected)))
+    tol = abstol + reltol*abs(expected)
+    if abs(expected - actual) > tol:
+        return "expected %.4g, got %.4g (tolerance %.3g)" \
+               % (expected, actual, tol)
 
 
 class pTest(object):
@@ -263,9 +265,6 @@ class pTest(object):
         if self.failed:
             if self.record:
                 self.file.write(msg + '\n')
-            if self.throw:
-                raise ValueError(msg)
-            print(msg)
         if self.record:
             self.file.close()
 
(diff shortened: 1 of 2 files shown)


src/common/Utilities.py
```diff
@@ -231,14 +231,16 @@ def show_3D_array\
     return 0
 
 
-def check_tolerance(expected, actual, abstol=0, reltol=1e-4):
+def check_tolerance(expected, actual, abstol=0, reltol=2e-3):
```
KrisThielemans, line 234: intentional to decrease default tolerance in this PR?

KrisThielemans (commented):
Ideally would have a test for `pTest` but maybe not to easy (intentionally have a test that fails, and if so, the pTesttest succeeds!)

src/common/Utilities.py
evgueni-ovtchinnikov: taken from #844

KrisThielemans: ok. confusing to mix the 2 PRs (might generate a conflict) but fine for me.

evgueni-ovtchinnikov:
@KrisThielemans re testing `pTest`: done manually by introducing errors in some test scripts and making sure they are reported by both `ctest` and `run_all.py`, with all good tests run unaffected

KrisThielemans:
> @KrisThielemans re testing pTest: done manually by introducing errors in some test scripts and making sure they are reported by both ctest and run_all.py, with all good tests run unaffected

Not ideal, but certainly good enough for me, given our priority list. Feel free to merge this whenever Travis is happy (enough).

PR #851, opened 2021-01-14
Activated use of AcquisitionsVector

johannesmayer:
The `AcquisitionsVector` class was initialised as a `AcquisitionsFile`.
This is now changed s.t. the storage scheme can be used in memory to forward-project into that container, and use the `set_acquisition()` method.

### DIFF
--- a/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
+++ b/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
@@ -192,12 +192,7 @@ namespace sirf {
 
 		static std::string storage_scheme()
 		{
-			static bool initialized = false;
-			if (!initialized) {
-				_storage_scheme = "file";
-				initialized = true;
-			}
-			return _storage_scheme;
+            return _storage_scheme;
 		}
 
 		// ISMRMRD acquisitions algebra: acquisitions viewed as vectors of 
@@ -313,7 +308,7 @@ namespace sirf {
         std::vector<KSpaceSorting> sorting_;
 		AcquisitionsInfo acqs_info_;
 
-		static std::string _storage_scheme;
+        static std::string _storage_scheme;
 		// new MRAcquisitionData objects will be created from this template
 		// using same_acquisitions_container()
 		static gadgetron::shared_ptr<MRAcquisitionData> acqs_templ_;
@@ -432,14 +427,21 @@ namespace sirf {
 		{
 			acqs_info_ = info;
 		}
-		static void init() 
-		{ 
-			AcquisitionsFile::init(); 
-		}
+        static void init()
+        {
+            static bool initialized = false;
+            if (!initialized) {
+                acqs_templ_.reset(new AcquisitionsVector());
+                _storage_scheme = "memory";
+                MRAcquisitionData::storage_scheme();
+                initialized = true;
+            }
+
+        }
 		static void set_as_template()
 		{
-			init();
-			acqs_templ_.reset(new AcquisitionsVector);
+            init();
+            acqs_templ_.reset(new AcquisitionsVector);
 			_storage_scheme = "memory";
 		}
 		virtual void empty();


evgueni-ovtchinnikov (commented):
Sorry Johannes, I cannot possibly approve this PR as it contradicts our storage scheme management approach.
Let me try to explain it on a simple example. Suppose we have two `MRAcquisitionData` objects `x` and `y`, and we need to compute `x + y`. For that we need to create a new `MRAcquisitionData`  object `z` to store the result. In our adopted storage scheme management approach the type of `z` is determined by the storage scheme currently in force ('file' or 'memory'). The function `cSIRF_axpby` that is to do the job has no idea what objects it works with, and it creates the new object using  `x.new_data_container_handle()`, which in the case at hand must create either `AcquisitionsFile` or `AcquisitionsVector` depending on the current storage scheme - _**not**_ by the type of `x`, which would be the case if you PR would be merged.

johannesmayer:
Brrr okay thanks for letting me know so quickly... I will try to work around it! 💯

evgueni-ovtchinnikov:
@johannesmayer: they say do not fix it if it is not broken - what are you trying to fix?

johannesmayer:
I am trying to use the `set_acquisition()` method.
But it is not implemented for `AcquisitionsFile` and I programmed the `MRAcquisitionModel::forward` method to use it.

KrisThielemans:
@johannesmayer did you then change the following on a different branch?
https://github.com/SyneRBI/SIRF/blob/933630ece93ded99dffbdc60e4eef993051621b1/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h#L382-L386
It might help to see that to know how the `init()` fails.

I certainly don't understand the intricacies here, but a few comments on the current code status on `master` (not in this PR):
- `storage_scheme()` implements a singleton pattern. If the code only ever uses `storage_scheme()`, it is guaranteed that there's only every one `storage_scheme` and it is always initialised, which is a good thing. (This prevents problems with unpredictable order of initialisation/construction of static global variables: they are initialised before the rest of the code runs, but you don't know in what order). You can read up on singletons [on SO here](https://stackoverflow.com/a/1008289/15030207) (note that that implementationisn't thread-safe apparently, see comments)
- although this sounds great, it seems to me that is broken by the fact that we set `_storage_scheme` "by hand" somewhere else.  This then relies on knowing in what order to call things (and note that there's no access to the `static` variable `initialized` from anywhere else), easily resulting in bugs.

Example: given the current implementation in `MRAcquisitionData`
https://github.com/SyneRBI/SIRF/blob/933630ece93ded99dffbdc60e4eef993051621b1/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h#L193-L201
 I think that the following would currently happen
```c++
_storage_scheme = "memory";
auto whatisit = storage_scheme();
```
resulting in `whatisit == "file"`. (reason: it's the first time that `storage_scheme()` is called, so it will override the value.

I think therefore that the following code happens to be ok
`AcquisitionsFile`
https://github.com/SyneRBI/SIRF/blob/933630ece93ded99dffbdc60e4eef993051621b1/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h#L347-L355
but only because both of them set `_storage_scheme` to `"file". It could be made safer by reording the statements, putting the call to `MRAcquisitionData::storage_scheme()` first, but it is certainly surprising to see a call to a funcion ignoring its return value, but relying on side-effects.

Similarly, its `set_as_template` 
https://github.com/SyneRBI/SIRF/blob/933630ece93ded99dffbdc60e4eef993051621b1/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h#L357-L362
"happens" to work ok, but only because `init()` sets the (2 different) `initialized` variables to `true`

So, I don't immediatelly see a bug in the current code on `master`, but it seems to me to be surprising and error prone. I can therefore imagine that a modification runs into trouble.

([similar implementations here for instance](https://github.com/SyneRBI/SIRF/blob/933630ece93ded99dffbdc60e4eef993051621b1/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h#L435-L444) for `AcquisitionsVector`)

johannesmayer:
I still don't understand how this does return anything else but `"file"`

https://github.com/SyneRBI/SIRF/blob/8a018ce4b031d68da32b3a4ec4f91979cf8cb77b/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h#L190-L202

however, the good thing is that I don't have to right now ...
So we can throw this guy in the trash then!

KrisThielemans:
> I still don't understand how this does return anything else but "file"

it's part of the weakness. It relies on somebody else setting `_storage_scheme` outside of this function. The first time, `storage_scheme()` would override that value (which is a bug arguably), but the second time it won't.

@evgueni-ovtchinnikov can you create an issue with my concerns about the current code please? Not urgent to address it as it's a bit of a corner case, but if we don't have the issue, we'll forget.

@johannesmayer good that you managed to work around your original problem

evgueni-ovtchinnikov:
@johannesmayer: `storage_scheme()` returns the default value "file" *only* if none of the derived classes' `set_as_template()` was called, otherwise it will return the scheme set by the latest call to `set_as_template()`.

@KrisThielemans: the value of `_storage_scheme` can only be changed by `set_as_template()`, which calls (via `init()`) `storage_scheme()` (thus triggering the relieve of the latter of its duty to ensure `_storage_scheme` is always defined) *before* setting the temlate.

PR #853, opened 2021-01-21
Add python algebra test

paskino:
adds unittests for algebra between `DataContainer`s and between `DataContainer` and `Number`s


src/common/tests/test_algebra.py
```diff
@@ -0,0 +1,192 @@
+import numpy
```
KrisThielemans, line 1: insert copyright header here

src/common/tests/CMakeLists.txt
```diff
@@ -0,0 +1,23 @@
+#========================================================================
+# Copyright 2017 - 2021 University College London
+# Copyright 2017 - 2021 Science Technology Facilities Council
```
KrisThielemans, line 3:
```suggestion
# Copyright 2021 Science Technology Facilities Council
```

src/common/CMakeLists.txt
```diff
@@ -63,7 +63,8 @@ if (BUILD_PYTHON)
   INSTALL(FILES "${CMAKE_CURRENT_BINARY_DIR}/pysirf.py" DESTINATION "${PYTHON_DEST}/sirf")
   file(GLOB PythonFiles "${CMAKE_CURRENT_LIST_DIR}/*.py")
   INSTALL(FILES ${PythonFiles} DESTINATION "${PYTHON_DEST}/sirf")
-
+  # add tests directory
+  add_subdirectory("tests")
```
KrisThielemans, line 67: is it really correct to have this in `common`? Seems unlikely as it needs STIR stuff.

KrisThielemans (changes requested):

src/common/CMakeLists.txt
paskino: Because I test with a PET image?

You are right. Is it possible to create a generic `DataContainer`?

I suppose I need to add the same tests for MR images.

src/common/tests/test_algebra.py
```diff
@@ -0,0 +1,192 @@
...
+import os
+import shutil
+import unittest
+import sirf.STIR as pet
+from sirf.Utilities import  examples_data_path
+
+
+class TestSIRFDataContainerAlgebra(unittest.TestCase):
```
paskino, line 9: I suppose I will create an equal class for `AcquisitionData`

src/common/CMakeLists.txt
paskino: and for NiftyImageData?

paskino:
added Algebra tests for MR and PET `DataContainer`

KrisThielemans:
I guess we need tests for every node container type. However, ideally this is done such that the tests are common, but only the creation of the relevant object  is not. (e.g. we put [is_operator_adjoint](https://github.com/SyneRBI/SIRF/blob/242303238f82580a78e8115c06643e2be9725b0f/src/common/Utilities.py#L511) and [pTest](https://github.com/SyneRBI/SIRF/blob/242303238f82580a78e8115c06643e2be9725b0f/src/common/Utilities.py#L246) in `common`, but they're called from elsewhere).

paskino:
I split the tests in MR and PET. I think this PR is a good starting point for a more complete unit testing for the Python layer. Thanks to these tests I actually fixed a few bugs in `Gadgetron.py`!

src/common/tests/CMakeLists.txt
```diff
@@ -0,0 +1,18 @@
+#========================================================================
```
KrisThielemans, line 1: entirely empty?

src/xSTIR/pSTIR/tests/test_algebra.py
```diff
@@ -0,0 +1,594 @@
...
+        tmp1.fill(2.)
+        image1.subtract(image2, out=tmp1)
+        
+        numpy.testing.assert_array_almost_equal(
+            numpy.zeros(image1.shape, dtype=numpy.float32), tmp1.as_array()
+            )
+
+
+
+class TestGadgetronAcquisitionDataAlgebra(unittest.TestCase):
```
KrisThielemans, line 396: can't be in `xSTIR`

src/xSTIR/pSTIR/tests/test_algebra.py
```diff
@@ -0,0 +1,594 @@
+#========================================================================
+# Copyright 2017 - 2021 University College London
```
KrisThielemans, line 2: do remove UCL. adjust date for next line

src/xGadgetron/pGadgetron/tests/test_algebra.py
```diff
@@ -0,0 +1,239 @@
+#========================================================================
+# Copyright 2017 - 2021 University College London
```
KrisThielemans, line 2: do remove UCL. adjust date for next line

KrisThielemans (changes requested):
I'd love to see more commonality, and less code, and re-use for `NiftiImageData` but can leave it for later if you want. :-)

src/xSTIR/pSTIR/tests/test_algebra.py
paskino: oops

src/common/tests/CMakeLists.txt
paskino: Shall I remove it? It'd be useful once we want to add it!

src/xGadgetron/pGadgetron/tests/CMakeLists.txt
```diff
@@ -1,6 +1,4 @@
 #========================================================================
-# Author: Edoardo Pasca
-# Copyright 2017 University College London
 # Copyright 2017 Science Technology Facilities Council
 # Author: Casper da Costa-Luis
 # Copyright 2017
```
KrisThielemans, line 4: now that I've started (?) to be annoying, checking the history, this needs to be 2017-2021. I'd remove Casper and the 2017 below as his change here is half a line (of course, the whole thing is rather trivial and copyright statements for this type is file is a bit silly)

src/common/tests/CMakeLists.txt
KrisThielemans: yeah, but there's nothing in it. I'd remove it

KrisThielemans: obviously, means removing the change in `common/CMakeLists.txt` as well

KrisThielemans (changes requested):
a squash merge after this?

paskino:
I reworked all the tests, I introduced the [`TestDataContainerAlgebra`](https://github.com/SyneRBI/SIRF/blob/add_python_algebra_test/src/common/Utilities.py#L550) class in `Utilities.py` so that the unit tests for all engines can inherit from it and perform the same tests without code duplication, just by changing the `setUp` method.

I stumbled upon missing or incomplete algebra methods, `* /` for `NiftiImageData` which I have implemented currently in Python, but would like to implement in cSIRF, @evgueni-ovtchinnikov .

src/Registration/pReg/Reg.py
```diff
@@ -146,7 +146,10 @@ def get_as_deformation_field(self, ref):
 
 class NiftiImageData(SIRF.ImageData):
     """General class for nifti image data."""
-
+    ADD      = 0
+    SUBTRACT = 1
+    MULTIPLY = 2
+    DIVIDE   = 3
```
KrisThielemans, line 152: Python doesn't have private members, but i believe there is a convention to start them with `_` or something like that. As  `help(NiftiImageData)` will show these members, I think it's important to at least convey to users that they shouldn't ever use this

src/Registration/pReg/tests/test_algebra.py
```diff
@@ -0,0 +1,41 @@
+#========================================================================
+# Copyright 2017 - 2021 Science Technology Facilities Council
```
KrisThielemans, line 2: I'm sorry, but you please be somewhat careful with dates.

src/common/Utilities.py
```diff
@@ -7,6 +7,7 @@
 import os
 import sirf.pyiutilities as pyiutil
 import re
+import numpy
 
 __licence__ = """SyneRBI Synergistic Image Reconstruction Framework (SIRF)
 Copyright 2015 - 2019 Rutherford Appleton Laboratory STFC
```
KrisThielemans, line 13: dates here

src/xGadgetron/pGadgetron/tests/test_algebra.py
```diff
@@ -0,0 +1,68 @@
+#========================================================================
+# Copyright 2017 - 2021 Science Technology Facilities Council
```
KrisThielemans, line 2: just 2021

src/xSTIR/pSTIR/tests/test_algebra.py
```diff
@@ -0,0 +1,62 @@
+#========================================================================
+# Copyright 2017 - 2021 Science Technology Facilities Council
```
KrisThielemans, line 2: just 2021

KrisThielemans (changes requested):
thanks for this improvement!

KrisThielemans:
`Reg.py` also needs its copyright header updating

src/xGadgetron/pGadgetron/tests/test_algebra.py
paskino: Also I removed the term SuperBuild which hasn't anything to do with this repo.

paskino:
@KrisThielemans I am quite confident that the status of this PR is good to squash-and-merge once travis is happy as well. ;)

Adding unittests in Python lead to discovering a few bugs and a few missing implementations.

KrisThielemans:
go ahead and enjoy pressing the button!

paskino:
After 5672cdc336f2d1d0a1dc0147973a4f58923a3b25 now all tests are the same for the 3 engines. Travis should be happy and then I'll merge.

paskino:
I don't understand the error on Travis as it works nice on my machine.

KrisThielemans (changes requested):
Also needs info in `CHANGES.md` (added missing operations, and added tests) and possibly checking the `UserGuide`

KrisThielemans:
I noticed a weird one at https://travis-ci.org/github/SyneRBI/SIRF/jobs/757041003#L30762
```
ERROR: test_six_adjoint.test_main
...
sirf.Utilities.error: ??? "'ProjDataInterfile: error opening output file tmp_1_1612170647450.s\\n' exception caught at line 638 of /home/travis/build/SyneRBI/SIRF/src/xSTIR/cSTIR/cstir.cpp; the reconstruction engine output may provide more information"
```
This doesn't seem to have anything to do with this PR. Also later [OSSPS test](https://travis-ci.org/github/SyneRBI/SIRF/jobs/757041003#L30806) and [test_two](https://travis-ci.org/github/SyneRBI/SIRF/jobs/757041003#L30827). Are you changing the directory somewhere?

paskino:
I create a temporary folder and then delete it [here](https://github.com/SyneRBI/SIRF/blob/9ab281949e8454856f122ae1963b12c38a67db6b/src/xSTIR/pSTIR/tests/test_algebra.py#L35L44) and [here](https://github.com/SyneRBI/SIRF/blob/9ab281949e8454856f122ae1963b12c38a67db6b/src/xSTIR/pSTIR/tests/test_algebra.py#L35L44)

I don't think this is at all relevant, though.

KrisThielemans:
Those bits contain a `chdir`
https://github.com/SyneRBI/SIRF/blob/9ab281949e8454856f122ae1963b12c38a67db6b/src/xSTIR/pSTIR/tests/test_algebra.py#L37
which could be the problem. There's another one above
https://github.com/SyneRBI/SIRF/blob/9ab281949e8454856f122ae1963b12c38a67db6b/src/xSTIR/pSTIR/tests/test_algebra.py#L31

These lines are a copy of the examples where I did the `chdir` to make sure that future reconstructions  etc all end-up in a specific place without cluttering the original repo, and where they can be removed easily. I can't see why we'd need it here.

I suggest trying
````python
self.image1 = pet.ImageData(os.path.join(examples_data_path('PET'), 'brain','emission.hv'))
```
it's shorter as well!

src/xSTIR/pSTIR/tests/test_algebra.py
```diff
@@ -0,0 +1,86 @@
...
+            tmp = self.image1 / self.image2
+            self.assertFalse(True)
+        except ZeroDivisionError as derror:
+            self.assertTrue(True)
+            print ("Caught error", derror)
+        except error as derror:
+            self.assertTrue(True)
+            print ("Caught error", derror)
+
+class TestSTIRAcquisitionDataAlgebra(unittest.TestCase, TestDataContainerAlgebra):
```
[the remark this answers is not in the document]
KrisThielemans: this comment is still the case in the code. The `.hv` files are images

paskino:
I need some help @casperdcl @KrisThielemans 
I've updated the tests and they all work on my machine. However `ctest` fails miserably. I think I know what happens, not why and I cannot understand how to fix it.

As an example, here is the `CMakeLists.txt` of the PET tests:

```python
set (exclude_tests "--ignore-files='test_algebra' --exclude='TestDataContainerAlgebra'")
# If not NiftyPET projector, dont test that
if (NOT STIR_WITH_NiftyPET_PROJECTOR)
	set(exclude_tests "--ignore-files='test_algebra|tests_NiftyPET' --exclude='TestDataContainerAlgebra'")
endif()

add_test(NAME PET_TESTS_PYTHON
  COMMAND ${PYTHON_EXECUTABLE} -m nose --with-coverage --cover-package=sirf.STIR src/xSTIR/pSTIR ${exclude_tests}
  WORKING_DIRECTORY ${CMAKE_SOURCE_DIR})
add_test(NAME PET_TESTS_PYTHON_ALGEBRA
  COMMAND ${PYTHON_EXECUTABLE} -m unittest test_algebra
  WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR})
```

We add the `PET_TESTS_PYTHON` which calls `nose`. Because `nose` seems quite greedy it'll execute everything it thinks it's a test, so it'll try to execute every file in the `tests` directory, so it'd execute the `test_algebra.py`. 
I managed to put together the appropriate `nose` command to skip both the file `test_algebra.py` AND the new class `TestDataContainerAlgebra` that is imported by it. Additionally we had some test exclusion beforehand. 

This works if I issue it in the same `WORKING_DIRECTORY` as in the `CMakeLists.txt` above. However, when `ctests` runs it fails, seemingly not passing `${exclude_tests}`

This is one output of a `ctest`:
```
10: Test command: /usr/bin/python3 "-m" "nose" "--with-coverage" "--cover-package=sirf.STIR" "src/xSTIR/pSTIR" "--ignore-files='test_algebra|tests_NiftyPET' --exclude='TestDataContainerAlgebra'"
```

And the failures are of this sort:

```
10: ======================================================================
10: ERROR: test_algebra.TestDataContainerAlgebra.test_add_datacontainer
10: ----------------------------------------------------------------------
10: Traceback (most recent call last):
10:   File "/home/ofn77899/.local/lib/python3.6/site-packages/nose/case.py", line 198, in runTest
10:     self.test(*self.arg)
10:   File "/home/ofn77899/devel/install/python/sirf/Utilities.py", line 656, in test_add_datacontainer
10:     image1 = self.image1
10: AttributeError: 'TestDataContainerAlgebra' object has no attribute 'image1'
```
This means that `nose` hasn't excluded `TestDataContainerAlgebra` class from the tests.I have the feeling that the REGEXP `test_algebra|tests_NiftyPET` isn't passed on well?

KrisThielemans:
It's probably a quoting issue. This
```
Test command: /usr/bin/python3 "-m" "nose" "--with-coverage" "--cover-package=sirf.STIR" "src/xSTIR/pSTIR" "--ignore-files='test_algebra|tests_NiftyPET' --exclude='TestDataContainerAlgebra'"
```
should probably be
```
Test command: /usr/bin/python3 "-m" "nose" "--with-coverage" "--cover-package=sirf.STIR" "src/xSTIR/pSTIR" "--ignore-files='test_algebra|tests_NiftyPET'" "--exclude='TestDataContainerAlgebra'"
```
I'd try
```cmake
set(exclude_tests "--ignore-files='test_algebra|tests_NiftyPET'" "--exclude='TestDataContainerAlgebra'")
```
or no `"` at all (They're often not necessary in CMake)

paskino:
Thanks @KrisThielemans you brought me on the right path: one problem were the single quotes. So far I managed to fix the MR tests (which are simple and without more than one file to skip.

I also found that some variables that were used to take some decisions were out of scope, so the boolean evaluation would always be false. 

One step at a time.

paskino:
OK, I found that loading a template `AcquisitionData` and use storage scheme `memory` fails, #862

paskino:
On travis none of the `SIRF_TESTS` run. :(

rijobro:
They may well be running but since they're not producing any output and they timeout, you don't know which test is causing the problem. You could temporarily modify the test script to run the tests from inside the SIRF build folder (rather than the SuperBuild folder).

`cd builds/SIRF/build && ctest --verbose`

At least that way you'll see the individual SIRF tests and you should see where it trips over. Put this command before the current `ctest`, otherwise it'll timeout before it gets the chance to do the individual SIRF tests.

rijobro:
This line: https://github.com/SyneRBI/SIRF/blob/master/.travis.yml#L251

paskino:
@KrisThielemans I think this is ready to be squashed and merged the tests have passed. https://travis-ci.org/github/SyneRBI/SIRF/builds/757696622

src/common/Utilities.py
```diff
@@ -67,7 +67,7 @@ def existing_filepath(data_path, file_name):
     data_path: path to the file
     file_name: file name
     '''
-    full_name = data_path + '/' + file_name
+    full_name = os.path.join(os.path.abspath(data_path), file_name)
```
KrisThielemans, line 70: nice one. This might explain why we currently need on Windows
```
 # need to be forward slashes 
 $Env:SIRF_PATH = "C:/Users/krisf/Documents/devel/buildVC/SIRF-SuperBuild/sources/SIRF"
```
https://github.com/SyneRBI/SIRF-SuperBuild/issues/460

@evgueni-ovtchinnikov after this has merged, could you try to set `SIRF_PATH` with backslashes and see if it works?

src/common/Utilities.py
```diff
@@ -544,3 +544,200 @@ def is_operator_adjoint(operator, num_tests = 5, max_err = 10e-5, verbose = True
...
+        if hasattr(self, 'cwd'):
+            os.chdir(self.cwd)
+        try:
+            self.image2 *= 0
+            tmp = self.image1 / self.image2
+            self.assertFalse(True)
+        except ZeroDivisionError:
+            self.assertTrue(True)
+        except error:
+            self.assertTrue(True)
```
KrisThielemans, line 743: This is testing for exceptions to be thrown, but see https://github.com/SyneRBI/SIRF/issues/863. want to merge this as-is and change it later?

KrisThielemans (commented):
one question, but good to squash-merge for me. (please clean-up the squash-merge commit message a bit such that it says what the overall change is, as opposed to the 31 steps along the way)

src/common/Utilities.py
paskino: I will change it when we take a decision on how to handle division by zero, #864.
So far it is a nice test that checks if the exception is thrown when expected.

PR #856, opened 2021-01-25
Add warning to stir.show()

AnderBiguri:
stir.show() will silently do nothing if you accidentally give out of index slice. 

Did it with print to avoid importing more  modules, but can change that to proper warning modules.


KrisThielemans (commented):
Would it not be better to raise `IndexError` aside from just returning? @evgueni-ovtchinnikov, is there a reason we don't?

Ideally we add a check on the `tof` index as well

AnderBiguri:
Sounds good to me, did not added error simply because I thought there may be a reason to just return and not crash. Happy to change whatever, let me know,

KrisThielemans:
@evgueni-ovtchinnikov could you tell us if we can throw an exception ? @AnderBiguri, once we know, can you merge `master` in here? Travis should now work

AnderBiguri:
Sure. Just let me know what you want the code to do and I'll merge master+commit.

evgueni-ovtchinnikov:
yes, raising `IndexError` is a good idea

PR #857, opened 2021-01-30
improve user experience for reading PET norm files

KrisThielemans:
- fix doc
- write a message what it will do (in case it fails cryptically afterwards)
- in Python, fix a small memory leak


KrisThielemans:
@evgueni-ovtchinnikov please check if I should have deleted the handle or not

evgueni-ovtchinnikov (approved):

PR #868, opened 2021-02-09
Require python3

paskino:
Currently this implementation only sends a deprecation [warning ](https://github.com/paskino/SIRF/blob/require_python3/CMakeLists.txt#L129)if building with Python2.
This can be changed to `FATAL_ERROR` at later stage.


src/common/CMakeLists.txt
```diff
@@ -57,8 +57,8 @@ if (BUILD_PYTHON)
   FIND_PACKAGE(PythonLibs REQUIRED)
```
KrisThielemans, line 57: this probably has to be removed

KrisThielemans (changes requested):
great PR. It will however need changes in the SuperBuild as well. https://github.com/SyneRBI/SIRF-SuperBuild/issues/313

CHANGES.md
```diff
@@ -1,6 +1,7 @@
 # ChangeLog
 
 ## vX.X.X
+* Use CMake variable names from find_package(Python) which are available with CMake 3.12+. A warning is sent when the python version found is 2. This can be changed to FATAL_ERROR at a later stage.
```
KrisThielemans, line 4: let's make this a bit clearer. If they passed `PYTHON_EXECUTABLE` before, it'll be ignored. 

Unless we put in some lines
```cmake
if (PYTHON_EXECUTABLE and not Python_EXECUTABLE)
  set(Python_EXECUTABLE %{PYTHON_EXECUTABLE})
endif()
# same for other python variables
```
that might be a good idea anyway?

paskino:
yes, this PR is backwards compatible, but it will use the new CMake [`findPython`](https://cmake.org/cmake/help/latest/module/FindPython.html) function. We can disable Python2 at any time by just checking once the `Python_VERSION_MAJOR`.

KrisThielemans:
> yes, this PR is backwards compatible,

it currently isn't as it will act differently when people set `PYTHON_EXECUTABLE` etc, as noted.

We could do something to go either way
```cmake
if (CMake version is too old)
  if (Python_EXECUTABLE and not PYTHON_EXECUTABLE)
    set(PYTHON_EXECUTABLE ${Python_EXECUTABLE})
  endif()
  # same for other python variables
else()
  if (PYTHON_EXECUTABLE and not Python_EXECUTABLE)
    set(Python_EXECUTABLE ${PYTHON_EXECUTABLE})
  endif()
# same for other python variables
endif()
```
too complicated?

KrisThielemans:
by the way, once we agree (!), and you have something that works, please do a PR for STIR to fix https://github.com/UCL/STIR/issues/655...

paskino:
Nothing is too complicated, ah ah ah. 

Well, maybe STIR is?

paskino:
I think [this](https://github.com/paskino/SIRF/blob/require_python3/CMakeLists.txt#L108L124) logic captures what you said.

The user can pass in both `Python_EXECUTABLE` and `PYTHON_EXECUTABLE`. They will get a deprecation warning if they use `PYTHON_EXECUTABLE`. 

Let's wait 4 hours for travis and then merge.

KrisThielemans:
I think this is fine. @rijobro had to pass Python libraries as well, hence [these lines](https://github.com/SyneRBI/SIRF-SuperBuild/blob/08ffeb5ca408a753b923d24a0489d5844d47a313/SuperBuild.cmake#L140-L149) (which will now be ignored), but with any luck, that won't be necessary anymore with the updated CMake module.

PR #869, opened 2021-02-10
Test only SIRF tests and force output on travis

paskino:
Following https://github.com/SyneRBI/SIRF/pull/853#issuecomment-774037988 I suggest to run only SIRF tests and force output on travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -248,16 +248,16 @@ script:
 # print for debugging
 - cat builds/SIRF/build/CMakeCache.txt
 - |
-   ctest --verbose V --output-on-failure; test_fail=$?
-   # echo "----------- Killing gadgetron server"
-   # killall gadgetron
-   if [[ $test_fail -ne 0 ]]; then
-     #echo "----------- Test output"
-     # cat builds/SIRF/build/Testing/Temporary/LastTest.log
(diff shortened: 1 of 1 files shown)


KrisThielemans:
I certainly agree to run only SIRF tests in the SIRF Travis, but I think useful to keep the gadgetron log to see why stuff failed.
Also, best to keep this travis as close as possible to the SIRF-Superbuild one.

Summary: I think only change needed is to say
```
cd builds/SIRF/build; ctest --verbose; test_fail=$?
```

paskino:
In the original command, `test_fail` will be populated only in case of some test failure. 

```bash 
ctest --verbose V --output-on-failure; test_fail=$?
```

In the new command `test_fail` will be always populated, won't it? That's the reason I commented out all the gadgetron log section.

KrisThielemans:
possibly there's some Travis trickery with the `| |` and subshells, but I cannot see the difference (aside from the fact that possibly I should have written `cd builds/SIRF/build && ctest --verbose; test_fail=$?` but I wouldn't know why that'd make any difference. In all of these, `test_fail` should be set to the exit value of `ctest`. no?

paskino:
Oh yes, you are right. I had thought it'd have the output of the tests, which isn't true.

PR #710, opened 2020-06-06
Coverage fixes

KrisThielemans:
coveralls and codecov were broken.

fixes #709

PR #870, opened 2021-02-11
Add requirements.txt

ashgillman:
Hopefully the Travis test will tell us if this is working - `pip install -r requirements.txt` seemed to be happy, but that was the extent of my testing.

The commented-out lines were not declared in the Wiki but were in Travis, so I expect they may be needed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -221,6 +221,7 @@ before_install:
 - $PY_EXE -m pip install --user -U pip setuptools wheel
 - $PY_EXE -m pip freeze
 - $PY_EXE -m pip install --user -r requirements.txt
+- $PY_EXE -m pip install --user -r ci-requirements.txt
 # no point re-downloading SIRF - just use local version
 # N.B.: don't put into build matrix to allow caching.
 - BUILD_FLAGS="$BUILD_FLAGS -DPYVER=$PYMVER -DSIRF_SOURCE_DIR:PATH=$PWD -DDISABLE_GIT_CHECKOUT_SIRF=ON"
--- a/.travis.yml
+++ b/.travis.yml
@@ -192,15 +192,6 @@ before_install:
        sudo apt install libvtk7-dev
      fi
    fi
-- $PY_EXE --version
-# get pip
-- curl -0 https://bootstrap.pypa.io/get-pip.py -o get-pip.py
-- $PY_EXE get-pip.py --user
-- echo "Using Python executable $PY_EXE"
-- $PY_EXE --version
-- $PY_EXE -m pip --version
-# setuptools may be out of date on osx
-- $PY_EXE -m pip install --user -U pip setuptools wheel
 # for counting clones, excluding ours
 - |
    if [[ -n "$GITHUB_API_TOKEN" ]]; then
@@ -210,7 +201,7 @@ before_install:
      # count unique clones, excluding travis, and print total
      TRAVIS_REPO_SLUG=CCPPETMR/SIRF-SuperBuild gh_stats_count -k uniques
    fi
-- $PY_EXE -m pip freeze
+- $PY_EXE --version
 # ccache compiler override
 - ln -s "$(which ccache)" g++
 - ln -s "$(which ccache)" g++-6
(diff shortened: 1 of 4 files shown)


.travis.yml
```diff
@@ -201,8 +201,6 @@ before_install:
 - $PY_EXE -m pip --version
 # setuptools may be out of date on osx
 - $PY_EXE -m pip install --user -U pip setuptools wheel
-# ensure python bin dir exists (and coverage dependencies installed)
-- $PY_EXE -m pip install --user -U nose codecov coveralls requests
```
ashgillman, line 205: Why is this done pre-installation? Probably that was important but I ignored it...

requirements.txt
```diff
@@ -0,0 +1,11 @@
...
+--only-binary=scipy
+--only-binary=matplotlib
+docopt
+psutil
+nose
+coverage
+#coveralls?
+#requests?
+#deprecation?
+#nibabel?
```
ashgillman, line 11: Probably we want nibabel for testing - it is optional. It seems the preferred way to do this is an optional-requirements.txt. I can add this?

.travis.yml
KrisThielemans: can't see why it'd matter, but probably doesn't do any harm either

requirements.txt
KrisThielemans: ok. `coveralls` is used for coverage testing as well I believe, so optional. I thought we needed `deprecation` but that should be easy to check. `requests` I have no idea

paskino:
[This ](https://github.com/SyneRBI/SyneRBI_VM/blob/master/scripts/INSTALL_python_packages.sh#L69)is what's installed in the VM, for comparison.

paskino:
As you suggested @ashgillman the `requirements.txt` is [not available](https://travis-ci.org/github/SyneRBI/SIRF/jobs/758531042#L1010) yet when issuing `pip install -r requirements.txt` because the repo hasn't been cloned yet.

KrisThielemans:
yes it is. see https://travis-ci.org/github/SyneRBI/SIRF/jobs/758531042#L749

ashgillman:
From here, before merging, we need to add `pip install -r optional-requirements.txt`. I just want to make sure it runs without them first (i.e., that they are indeed optional)

KrisThielemans:
`coverage` and `coveralls` are used by `ctest`. However, our upload is currently broken. See #710 for a "solution" that didn't work.

`requests` is no (longer?) used. Potentially it is used in SIRF-exercises, but then it should be added there.

`deprecation` is currently not used anymore but was in the past and could be in the future. Please add it (always, not optional)

As opposed to creating an optional file, we could apparently use  `coveralls[test]`, see https://www.python.org/dev/peps/pep-0508/#extras

ashgillman:
Thanks for the info on those deps.

What you suggest is a little different. The notation `coveralls[test]` would be to install the "test" dependencies of "coveralls", which would have to be in turn specified in coveralls' `setup.py` - `requirements.txt` doesn't support extras ([intentionally](https://github.com/pypa/pip/issues/1161)).

Reading through that, it is a bit more clear that requirements.txt is intended for users or deployment, as opposed to defining the dependencies of a project (which should be in setup.py).

So anything for CI etc. I guess should go in some form of requirements.txt (I would think maybe a ci-requirements.txt) and any dependencies could either go in a requirements.txt or we could put them into the setup.py.in.

### requirements.txt Pros
- easier for interested user to see in the source code - our setup.py.in is a little hard to find
- closer aligned with what we wanted - just a way to communicate the dependencies to users and give them a simple way to install them.
- Can install dependencies separate to installation

### setup.py.in Pros
- Can set up extras, like [test], [optional] (nibabel)
- Dependencies are automatically installed during installation
### Cons
- Because the Python installation directory is non-standard, I'm not sure how well this works and where the dependencies go.

Because I can't see the project going up on the likes of PyPI any time soon, I don't think going the setup.py route is particularly necessary.

KrisThielemans (approved):

PR #872, opened 2021-02-11
Copy examples demos

### LINKED ISSUE #181: install demos and documentation
we need to copy the examples into `${CMAKE_INSTALL_PREFIX}/share/SIRF/examples` and `${CMAKE_INSTALL_PREFIX}/share/SIRF/doc`

should use version number

create a script to "git clone SIRF-data".

Migrated from https://github.com/CCPPETMR/SIRF-SuperBuild/issues/102

### LINKED ISSUE #463: Install data to install directory
The environment scripts now contain the variable `SIRF_INSTALL_PATH` which points to the installation directory. 

I suggest to use the directory `${SIRF_INSTALL_PATH}/share/sirf/data` to put the data in. 

This can also be used by the download [scripts ](https://github.com/CCPPETMR/SIRF-Exercises/blob/master/scripts/download_MR_data.sh#L81)of the `SIRF-Exercises` instead of the current `$SIRF_PATH` which points to the source directory (which resides in the build directory). 

Also this is compatible with conda installation where one can use `sys.prefix/share/sirf`

The same directory `$SIRF_INSTALL_PATH/share/ccpi` is used if `BUILD_CIL=ON` to store the data for CIL.

paskino:
copies the `examples, doc` and `data` directory to the `${CMAKE_INSTALL_PATH}/share/SIRF` directory.

Also corrects `examples_data_path` to use `os` for path manipulation
`petmr_data_path` calls `examples_data_path`

closes #181 
closes #463


KrisThielemans:
I'd prefer to install in `${CMAKE_INSTALL_PATH}/share/SIRF-${VERSION_MAJOR}.${VERSION_MINOR}`, (set around [here](https://github.com/SyneRBI/SIRF/blob/41336634af15d83db7afbb57292a7752f6735e07/CMakeLists.txt#L58)). A first step towards adding version numbers...

paskino:
sure. Is a version string available to Python/MATLAB?

paskino:
This will now add

```python
import sirf

print (sirf.__version__)
```

It doesn't do anything wrt MATLAB.

KrisThielemans (changes requested):
we're now using a new env variable `SIRF_INSTALL_PATH`. This is new and should be documented in a few places. It would also need  changes in the SuperBuild `env*sh` scripts.

There's a tricky question with this strategy however. If you're developing, do you want examples_data_path to point to an install directory, or the developer version? Now we force it to the install. I think a better strategy might be to 
```
if SIRF_DATA_PATH exists, use it
else if SIRF_INSTALL_PATH exists, do what you do
else error (or could even hunt in /usr/share and /usr/local/share, but not on Windows...)
```
 that way, the developer can set things as they like.

it would need a similar change for Matlab. Might be easier/better to do this on the C++ level (with a `configure`).

By the way, why are the commits from #871 in this diff?

paskino:
The `data` directory is copied from SIRF source directory, which for a developer will be the one relative to the branch they are using. So I don't see the difference.

I think this has the benefit of separating the build directory from the install, which is a good step in the way of being able to distribute a binary package.

For the working of MATLAB I have no clue and would need some help.

paskino:
For both Python and MATLAB, the resolution order for the examples data path is:
1. in the directory pointed by `${SIRF_DATA_PATH}` that can be specified by the user externally
2. in the `${SIRF_INSTALL_PATH}/share/SIRF-<version>`

The version string is set by CMake
* for Python in [`sirf/__init__.py`](https://github.com/paskino/SIRF/blob/copy_examples_demos/cmake/sirf.__init__.py.in#L1)
* for MATLAB in [`examples_data_path.m`](https://github.com/paskino/SIRF/blob/copy_examples_demos/src/common/%2Bsirf/%2BUtilities/examples_data_path.m#L29)

CMakeLists.txt
```diff
@@ -195,3 +195,11 @@ ADD_SUBDIRECTORY(src)
 include("${CMAKE_CURRENT_SOURCE_DIR}/cmake/setup.py.cmake")
 configure_file("cmake/config.py.in" "${CMAKE_CURRENT_BINARY_DIR}/cmake/config.py")
 INSTALL(FILES "${CMAKE_CURRENT_BINARY_DIR}/cmake/config.py" DESTINATION "${PYTHON_DEST}/sirf")
+set(SHARE_DIR ${CMAKE_INSTALL_PREFIX}/share/SIRF-${VERSION_MAJOR}.${VERSION_MINOR}.${VERSION_PATCH})
```
KrisThielemans, line 198:
```suggestion
set(SHARE_DIR ${CMAKE_INSTALL_PREFIX}/share/SIRF-${VERSION_MAJOR}.${VERSION_MINOR})
```
no patch info (data and doc shouldn't really change (a lot) between patches

src/common/+sirf/+Utilities/examples_data_path.m
```diff
@@ -19,10 +19,15 @@
...
 
-    SIRF_PATH = getenv('SIRF_PATH');
-    if ~isempty(SIRF_PATH)
-        p = [SIRF_PATH '/data/examples/' petmr];
+    
+    SIRF_INSTALL_PATH = getenv('SIRF_INSTALL_PATH');
+    SIRF_DATA_PATH = getenv('SIRF_DATA_PATH')
+    if ~isempty(SIRF_DATA_PATH)
+        p = [ SIRF_DATA_PATH filesep 'examples' filesep petmr];
+    elseif ~isempty(SIRF_PATH)
```
KrisThielemans, line 27:
```suggestion
    elseif ~isempty(SIRF_INSTALL_PATH)
```

src/common/+sirf/+Utilities/examples_data_path.m
```diff
@@ -19,10 +19,15 @@
...
+        % p = [SIRF_PATH '/data/examples/' petmr];
+        p = [ SIRF_INSTALL_PATH filesep 'share' filesep 'SIRF-@VERSION_MAJOR@.@VERSION_MINOR@.@VERSION_PATCH@' filesep 'data' filesep 'examples' filesep petmr ];
```
KrisThielemans, line 29:
```suggestion
        p = [ SIRF_INSTALL_PATH filesep 'share' filesep 'SIRF-@VERSION_MAJOR@.@VERSION_MINOR@' filesep 'data' filesep 'examples' filesep petmr ];
```

src/common/CMakeLists.txt
```diff
@@ -93,5 +93,6 @@ if (BUILD_MATLAB)
   INSTALL(DIRECTORY +sirf DESTINATION "${MATLAB_DEST}")
   file(GLOB MatlabFiles "${CMAKE_CURRENT_LIST_DIR}/*.m")
   INSTALL(FILES ${MatlabFiles} DESTINATION "${MATLAB_DEST}")
+  configure_file("${CMAKE_CURRENT_SOURCE_DIR}/+sirf/+SIRF/+Utilities/examples_data_path.m" ${MATLAB_DEST}/+SIRF/+Utilities/examples_data_path.m)
```
KrisThielemans, line 96: I'd file more comfortable with having a `.m.in` file. Or better: we create a `version.m.in` , and get the info from there

KrisThielemans (changes requested):
As I wrote, I prefer to install in `share/SIRF-2.0`, i.e. without patch info. If you patch, you don't really think that doc etc is going to change. It's what ITK does for instance. Sadly this means that your current strategy of simply relying on `__version__` would fail (i.e. you'd need to remove the patch). sorry

KrisThielemans:
Could you also add a brief section in https://github.com/SyneRBI/SIRF/blob/master/doc/UserGuide.md, near the end I guess, that talks about the environment variables (used by tests and examples only)?

paskino:
Once this is approved I'll add the docs

KrisThielemans (changes requested):
All fine. let's just create `version.m.in` , `version_major.m.in` etc files for matlab, and use those. Like this
```matlab
% return SIRF version string (major.minor.patch)
% Usage:
% version_string = version()
function str = version()
  str = '@VERSION_MAJOR@.@VERSION_MINOR@.@VERSION_PATCH@';
```
and
```matlab
p = fullfile(SIRF_INSTALL_PATH , 'share', ['SIRF-' version_major() '.' version_minor() ], 'data' , 'examples',  petmr);
```

add the doc :smile:, and create an issue to create a `sirf_version` on the C++ side

paskino:
I believe the UserGuide is a programmer's guide. I cannot easily see where that info could fit. Maybe [here ](https://github.com/SyneRBI/SIRF/blob/master/doc/UserGuide.md#general-notes-of-usage-)but I'll have a think.

The info should go in the [wiki](https://github.com/SyneRBI/SIRF/wiki/Examples).

KrisThielemans:
>  Maybe here but I'll have a think.

sounds good to me

paskino:
I added info to the [wiki](https://github.com/SyneRBI/SIRF/wiki/Examples#environment-variables).

PR #881, opened 2021-03-12
MR Encoding

johannesmayer:
This PR is an overhaul of the `MRAcquisitionModel`.

- now a code basis is there to support other encoding schemes than cartesian. 
- this automatically re-generates the test data in the `$SIRF_PATH/data/MR/` folder s.t. conflicts with the ISMRMRD version are avoided.
- this can compute the coil sensitivities for different k-space sampling schemes and removes a large chunk of duplicate code.
- while radial readout reconstruction is not yet possible code is provided to reconstruct 3D radial phase encoding data.
- if the libraries for the gadgetron toolboxes are not detected then any code for non-cartesian reconstruction is excluded and a `runtime_error` is thrown if the user tries to reconstruct non-cartesian code anyway.

The backward and forward method now don't rely on the order of acquisition data in the memory any more.
Rather consistent sub-sets of data (i.e. acquisition that belong to the same slice, echo etc.) are created and then Fourier-transformed. Hence individual images based on these subsets are reconstructed. 

Forward projection works the other way around: each image is Fourier-transformed and the subset of acquisition data in the template containers that match the image are replaced. 

The `MRAcquisitionModel` and `CoilImageData` have a shared pointer to an encoding class member that performs these operations 

Potential drawbacks:
- this will only work if the `storage_scheme` is `'memory'` unless the `set_acquisition` method is implemented for the storage scheme `'file'`.


johannesmayer:
Eh this is not fair, why did one build fail? :( ... I finally got it to run!! 👎 👎

KrisThielemans:
No idea why that job keeps failing with that error and not others.
For this reason, I had put in to run `ctest` silently. However, @paskino changed that to `--verbose` in #869. It does help with debugging with time-outs, but in this creates trouble. I'd consider reverting that.

On the other hand I see lots of
```
GadgetronImagesVector::set_up_geom_info(): read_dir, phase_dir and slice_dir should all be unit vectors.
```
is that your PR or is it already in past logs?

johannesmayer:
I don't think the `set_up_geom_info()` come from this PR, e.g. https://travis-ci.org/github/SyneRBI/SIRF/jobs/760445436 also contains them.

KrisThielemans:
> No idea why that job keeps failing with that error and not others.

@johannesmayer wrote in email "this only happens if I use the non-cartesian FFT, somehow this got switched on in the travis build because it found the Gadgetron libs. "

KrisThielemans:
> I don't think the `set_up_geom_info()` come from this PR, e.g. https://travis-ci.org/github/SyneRBI/SIRF/jobs/760445436 also contains them.

indeed. I've created #884

johannesmayer:
Sorry, I should have written it here. I agree that it is suspicious. Still, the test running the code passes. 
I checked again to see if there is another log that is longer and that passes, but the others passing have ~20k, 30274 and 33746 lines ... The third one fails after 36062 lines

johannesmayer:
Is there anything I could try now to make this one test pass??

KrisThielemans:
In my opinion, it needs running the tests silently unless they fail, see https://github.com/SyneRBI/SIRF/pull/881#issuecomment-802916693, *i.e. `ctest --output-on-failure` as opposed to `ctest --verbose`. @paskino?

johannesmayer:
With Travis builds put to a stop, does this mean that this PR can not be tested until that is sorted? 

It did somehow complete the build in the Superbuild PR SyneRBI/SIRF-SuperBuild#503 as the `version_config` was set to this branch.

src/xGadgetron/cGadgetron/CMakeLists.txt
```diff
@@ -28,14 +28,51 @@
...
+                )
+find_library( GT_LOG NAMES gadgetron_toolbox_log
+                NAMES_PER_DIR
+                PATHS ${CMAKE_INSTALL_PREFIX}
+                )
+
+set (GT_LIBS ${GT_CPUCORE} ${GT_FFT} ${GT_NFFT} ${GT_LOG} )
+
+if(GT_CPUCORE AND GT_FFT AND GT_NFFT AND GT_LOG)
+    MESSAGE(STATUS "The Gadgetron Toolboxes requried for radial gridding were found. Non-caratesian encoding will be compiled.")
```
KrisThielemans, line 61: please correct spelling mistakes

src/xGadgetron/cGadgetron/CMakeLists.txt
```diff
@@ -28,14 +28,51 @@
...
+    set(CGADGETRON_SOURCES ${CGADGETRON_SOURCES} noncartesian_encoding.cpp)
+
+    if(${CMAKE_VERSION} VERSION_LESS "3.12.0")
+      add_definitions(-DGADGETRON_TOOLBOXES_AVAILABLE)
+    else()
+      add_compile_definitions(GADGETRON_TOOLBOXES_AVAILABLE)
+    endif()
+else()
+    set(GADGETRON_TOOLBOXES_AVAILABLE FALSE)
+    MESSAGE(WARNING "The Gadgetron Toolboxes requried for radial gridding were NOT FOUND. Non-caratesian encoding will NOT BE compiled.")
```
KrisThielemans, line 72: and here

src/xGadgetron/cGadgetron/encoding.cpp
```diff
@@ -0,0 +1,311 @@
+#include "sirf/Gadgetron/encoding.h"
+
+#include <sstream>
+#include <math.h>
+
+//#include <cassert>
+//#define assertm(exp, msg) assert(((void)msg, exp)
+
+#define SIRF_GOLDEN_ANGLE M_PI*0.618034
```
KrisThielemans, line 9:
```suggestion
#define constexpr double SIRF_GOLDEN_ANGLE  = M_PI*0.618034
```
However, I'm fairly sure that `M_PI` isn't always going to be there, including on Windows. Might need its own constant.

src/xGadgetron/cGadgetron/encoding.cpp
```diff
@@ -0,0 +1,311 @@
...
+}
+
+void sirf::GRPETrajectoryPrep::set_acquisition_trajectory(Acquisition& acq)
+{
+    acq.resize(acq.number_of_samples(),acq.active_channels(), this->traj_dim_);
+    std::vector<float> acq_traj = this->calculate_trajectory(acq);
+    acq.setTraj(&acq_traj[0]);
+}
+
+std::vector<float> sirf::GRPETrajectoryPrep::calculate_trajectory(Acquisition& acq)
```
KrisThielemans, line 94: can this be a `const` member function?

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/encoding.h
```diff
@@ -0,0 +1,156 @@
...
+#include <ismrmrd/xml.h>
+#include "sirf/Gadgetron/gadgetron_data_containers.h"
+//#include "sirf/Gadgetron/gadgetron_image_wrap.h"
+#include "sirf/iUtilities/LocalisedException.h"
+
+#include <ismrmrd/ismrmrd.h>
+
+/*!
+\file
+\ingroup Fourier Encoding
```
KrisThielemans, line 44: do we actually have a group like this? Can there be spaces in group names? @evgueni-ovtchinnikov where do we define our doxygen groups?

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/encoding.h
```diff
@@ -0,0 +1,156 @@
...
+\brief Specification file for preparing MRAcquisitionData for Fourier encoding.
```
KrisThielemans, line 45:
```suggestion
\brief Specification file for preparing sirf.MRAcquisitionData for Fourier encoding.
```
needs namespace as comment is outside it

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/encoding.h
```diff
@@ -0,0 +1,156 @@
...
+
+\author Johannes Mayer
+\author CCP PETMR
```
KrisThielemans, line 48:
```suggestion
```
should have been SyneRBI, but in fact I don't think we should have this

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/encoding.h
```diff
@@ -0,0 +1,156 @@
...
+*/
+
+/*!
+\ingroup aTrajectoryPreparation
```
KrisThielemans, line 52: can't invent a doxygengroup all the time :-)

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/encoding.h
```diff
@@ -0,0 +1,156 @@
...
+\brief Abstract class for trajectory preparation
+
+*/
+
+namespace sirf{
+
```
KrisThielemans, line 58: needs a doxygen comment, that includes an `\ingroup` (i.e. a comment for the file, one for every class)

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/encoding.h
```diff
@@ -0,0 +1,156 @@
...
+    CartesianTrajectoryPrep(): aTrajectoryPreparation() {
+        traj_type_ = ISMRMRD::TrajectoryType::CARTESIAN;
+        traj_dim_ = 0;
+    }
+
+    virtual void set_trajectory(sirf::MRAcquisitionData& mr_acq);
+
+protected:
+    virtual void set_acquisition_trajectory(ISMRMRD::Acquisition& acq){}
+    virtual std::vector<float> calculate_trajectory(ISMRMRD::Acquisition& acq){return std::vector<float>{};}
```
KrisThielemans, line 97: ??

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -94,6 +101,8 @@ namespace sirf {
 				deserialize();
 				have_header_ = true;
 			}
+            std::cout << "################ WE HAAAAVE " << this->header_.encoding.size() << " encodings now ####################### " << std::endl;
+
```
KrisThielemans, line 105: I guess delete this

src/xGadgetron/pGadgetron/tests/test1.txt
```diff
@@ -1,9 +1,9 @@
-2.212010e+02
-2.090210e+02
-2.090212e+02
-2.090210e+02
-2.155659e-07
-9.322899e-08
-1.430538e-06
-0.000000e+00
-0.000000e+00
+2.212586e+02
```
KrisThielemans, line 1: are most of these actually zero? Do they need to change?

KrisThielemans (changes requested):
some housekeeping. I didn't check the actual code.

Also, can you make member functions `const` wherever you can?

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/encoding.h
johannesmayer: hahaha I'm sorry I just extrapolated ... maybe I should actually find out what doxygen does

src/xGadgetron/pGadgetron/tests/test1.txt
evgueni-ovtchinnikov: ?1: 5 values are close to zero (less than 1e-4)
?2: no need to change if equal within default relative tolerance of 2e-3

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/cgadgetron.h
```diff
@@ -68,10 +68,16 @@ extern "C" {
...
 	void* cGT_appendAcquisition(void* ptr_acqs, void* ptr_acq);
 	void* cGT_createEmptyAcquisitionData(void* ptr_ad);
+    void* cGT_getAcquisitionsSubset(void* ptr_acqs, PTR_INT ptr_idx, size_t const num_elem_subset);
+
 	void* cGT_cloneAcquisitions(void* ptr_input);
 	void* cGT_sortAcquisitions(void* ptr_acqs);
 	void* cGT_sortAcquisitionsByTime(void* ptr_acqs);
 	void* cGT_setAcquisitionsInfo(void* ptr_acqs, const char* info);
+    void* cGT_setGRPETrajecotry(void* ptr_acqs);
+    void* cGT_getGRPETrajecotry(void* ptr_acqs, PTR_FLOAT ptr_traj);
```
KrisThielemans, line 78: misspelling

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/encoding.h
```diff
@@ -0,0 +1,153 @@
+/*
+SyneRBI Synergistic Image Reconstruction Framework (SIRF)
+Copyright 2015 - 2021 Rutherford Appleton Laboratory STFC
```
KrisThielemans: you'll need to add PTB with appropriate dates in all your files (potentially removing STFC )

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/encoding.h
```diff
@@ -0,0 +1,153 @@
...
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+/*!
+\file
+\ingroup Gadgetron Extensions
```
KrisThielemans: probably no spaces allowed in group names

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/encoding.h
```diff
@@ -0,0 +1,153 @@
...
+    virtual void set_acquisition_trajectory(ISMRMRD::Acquisition& acq) const =0;
+    virtual std::vector<float> calculate_trajectory(ISMRMRD::Acquisition& acq) const =0;
+};
+
+
+typedef std::vector< std::pair<float, float> > SIRFTrajectoryType2D;
+
+/*!
+\ingroup Gadgetron Extensions
+\brief Class to set the gold-angle radial phase encoding (RPE) trajecotry
```
KrisThielemans:
```suggestion
\brief Class to set the golden-angle radial phase encoding (GRPE) trajectory
```

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -119,7 +127,18 @@ namespace sirf {
         mutable bool have_header_;
 	};
 
-    class KSpaceSorting
+    /*!
+    \ingroup Gadgetron Data Containers
```
KrisThielemans, line 131: probably no space

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -429,7 +475,12 @@ namespace sirf {
 	*/
 	class AcquisitionsVector : public MRAcquisitionData {
 	public:
-		AcquisitionsVector(AcquisitionsInfo info = AcquisitionsInfo())
+        AcquisitionsVector(const char* filename_with_ext)
```
KrisThielemans, line 478: I'd prefer to have this
```suggestion
        AcquisitionsVector(const std::string& filename_with_ext)
```
It will convert automatically from a `char*` (while it doesn't the other way around)

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -429,7 +475,12 @@ namespace sirf {
...
+        {
+            this->read(std::string(filename_with_ext));
```
KrisThielemans, line 480:
```suggestion
            this->read(filename_with_ext);
```

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -429,7 +475,12 @@ namespace sirf {
...
+        }
+
+        AcquisitionsVector(AcquisitionsInfo info = AcquisitionsInfo())
```
KrisThielemans, line 483:
```suggestion
        AcquisitionsVector(const AcquisitionsInfo& info = AcquisitionsInfo())
```
presumably we want to avoid copying the object

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -997,34 +1074,35 @@ namespace sirf {
         void set_csm_smoothness(int s){csm_smoothness_ = s;}
 
         void calculate(CoilImagesVector& iv);
-        void calculate(const MRAcquisitionData& acq)
+        void calculate(MRAcquisitionData& acq)
```
KrisThielemans, line 1077: hopefully can keep it `const`

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -997,34 +1074,35 @@ namespace sirf {
...
+        CFImage get_csm_as_cfimage(const KSpaceSubset::TagType tag, const int offset) const;
+
 
         void get_dim(size_t const num_csm, int* dim) const
         {
             GadgetronImagesVector::get_image_dimensions(num_csm, dim);
 
         }
 
+        void forward(GadgetronImageData& img, GadgetronImageData& combined_img)const;
```
KrisThielemans, line 1094: add `const` where you can. Needs to have same order as elsewhere (input vs output)

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -997,34 +1074,35 @@ namespace sirf {
...
 
-        void forward(){
-            throw LocalisedException("This has not been implemented yet." , __FILE__, __LINE__);
-        }
-        void backward(){
-            throw LocalisedException("This has not been implemented yet." , __FILE__, __LINE__);
-        }
+        //bool flag_imgs_suitable_for_csm_computation_=false;
 
+        void calculate_csm(ISMRMRD::NDArray<complex_float_t>& cm, ISMRMRD::NDArray<float>& img, ISMRMRD::NDArray<complex_float_t>& csm);
```
KrisThielemans, line 1105: add `const` where you can

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_x.h
```diff
@@ -368,6 +374,7 @@ namespace sirf {
 			gadgetron::shared_ptr<GadgetronImageData> sptr_ic
 			) : sptr_acqs_(sptr_ac) //, sptr_imgs_(sptr_ic)
 		{
+            sptr_enc_ = std::make_shared<sirf::CartesianFourierEncoding>(sirf::CartesianFourierEncoding());
```
KrisThielemans, line 377:
```suggestion
            sptr_enc_ = std::make_shared<sirf::CartesianFourierEncoding>();
```
before you were creating an object, and then calling the copy-constructor. Now you just create.

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/noncartesian_encoding.h
```diff
@@ -0,0 +1,110 @@
...
+\brief Implementation to perform a non-cartesian FFT for RPE MR data
+*
+* backward(...): first data are sorted into a 4D matrix consisting of kx, ky, kz, coil
+* and then the Fourier-transformed along the kx dimension. In the second step a NUFFT
+* along the two remaining dimensions is performed.
+*
+*/
+
+
+class RPEFourierEncoding : public FourierEncoding
```
KrisThielemans, line 60: `RPE` or `GRPE`?

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/noncartesian_encoding.h
```diff
@@ -0,0 +1,110 @@
...
+
+
+typedef Gadgetron::hoNDArray<Gadgetron::floatd2> GadgetronTrajectoryType2D;
+
+/*!
+\ingroup Gadgetron Extensions
+\brief Implementation to perform a non-cartesian FFT for RPE MR data
+*
+* backward(...): first data are sorted into a 4D matrix consisting of kx, ky, kz, coil
+* and then the Fourier-transformed along the kx dimension. In the second step a NUFFT
```
KrisThielemans, line 54:
```suggestion
* and then Fourier-transformed along the kx dimension. In the second step a NUFFT
```

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/noncartesian_encoding.h
```diff
@@ -0,0 +1,110 @@
...
+\brief Class to perform a NUFFT for 2D data
+*
+*
+*/
+
+class Gridder_2D
+{
+public:
+
+    Gridder_2D(std::vector<size_t> img_dims_output, const GadgetronTrajectoryType2D &traj) : nufft_operator_(from_std_vector<size_t, 2>(img_dims_output), (float)this->oversampling_factor_, (float)this->kernel_size_)
```
KrisThielemans, line 87: `img_dims_output` is presumably not output (can't be with this signature). Maybe it is `img_output_dims`?

Also, make it a `const vector&`

src/xGadgetron/cGadgetron/tests/CMakeLists.txt
```diff
@@ -17,13 +17,53 @@
...
+                NAMES_PER_DIR
+                PATHS ${CMAKE_INSTALL_PREFIX})
+
+if(EXE_ISMRMRD_SL AND RERUN_ISMRMRD_SHEPP_LOGAN)
+    add_custom_target(GENERATE_SIMULATED_TESTDATA ALL
+        COMMAND rm -f simulated_MR_2D_cartesian.h5
+        COMMAND rm -f simulated_MR_2D_cartesian_Grappa2.h5
+        COMMAND env LD_LIBRARY_PATH=${CMAKE_INSTALL_PREFIX}/lib:$LD_LIBRARY_PATH ${EXE_ISMRMRD_SL} -r 2 -o simulated_MR_2D_cartesian.h5
+        COMMAND env LD_LIBRARY_PATH=${CMAKE_INSTALL_PREFIX}/lib:$LD_LIBRARY_PATH ${EXE_ISMRMRD_SL} -a 2 -o simulated_MR_2D_cartesian_Grappa2.h5
+        WORKING_DIRECTORY "${SIRF_SOURCE_DIR}/data/examples/MR/")
```
KrisThielemans, line 49: @evgueni-ovtchinnikov , can we add your "fixing-up" code here?

src/xGadgetron/cGadgetron/tests/mrtest_auxiliary_funs.cpp
```diff
@@ -96,3 +103,66 @@ void sirf::write_cfimage_to_raw(const std::string& fname_prefix, const ImageWrap
 
     sirf::write_cfimage_to_raw(fname_prefix, *ptr_img);
 }
+
+void sirf::write_imagevector_to_raw(const std::string& fname_prefix, const sirf::GadgetronImagesVector iv)
```
KrisThielemans, line 107:
```suggestion
void sirf::write_imagevector_to_raw(const std::string& fname_prefix, const sirf::GadgetronImagesVector& iv)
```

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -834,6 +834,27 @@ def append_acquisition(self, acq):
...
+        Returns AcquisitionData object with subset of acquisitions defined by idx
+        '''
+        assert self.handle is not None
+        subset = AcquisitionData()
+        subset.handle = pygadgetron.cGT_getAcquisitionsSubset(self.handle, idx.astype(numpy.intc).ctypes.data, idx.size)
+        check_status(subset.handle)
+        
+        return subset
+    
+    def set_user_floats(self, data, idx):
```
KrisThielemans, line 849: this needs doc. Only ISMRMRD afficionados will understand it

KrisThielemans (changes requested):

johannesmayer:
I don't know about @ckolbPTB, but I will eternally shut up about any adjoint vs inverse and k-space density weighting, and I will admit that PET > MR 

small_triumph_mind_over_matter 
(gif shows some low-res RPE conjugate gradient convergence without any density weighting from some weird blob to the correct image)

examples/Python/MR/rpe_recon.py
```diff
@@ -0,0 +1,136 @@
...
+if data_path is None:
+    data_path = examples_data_path('MR') + '/zenodo/'
+output_file = args['--output']
+show_plot = not args['--non-interactive']
+
+run_recon = str(args['--no-cart']) == 'True'
+
+import numpy as np
+    
+def calc_ramlak_dcf(acq_data):
```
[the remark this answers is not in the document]
johannesmayer: I got rid of the whole thing instead

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/noncartesian_encoding.h
johannesmayer: RPE.
RPE just means, FFT along readout, and NUFFT in the phase encoding plane. This works for any kind of pattern in the phase encoding plane.

The G for golden GRPE is only one pattern coming from a golden-angle increment, but the RPE will work no matter what trajectory you store in your acquisitions.

johannesmayer: RPE

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/encoding.h
johannesmayer: i just copied from gadgetron_x.h where they were present.

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
johannesmayer: I followed the developer's guide where it says if I pass an output via a pointer or reference it shall come first.
I didn't cross-check it against the rest of the code-base, though.

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/encoding.h
KrisThielemans: ok. I've created https://github.com/SyneRBI/SIRF/issues/886.
Let's not change this in this PR therefore

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
KrisThielemans: ok great. that should mean that `combined_image` should be a `const GadgetronImageData&`. Maybe you could (in the .h file) name the first arg of `forward` as `output`, just to make it obvious...

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
johannesmayer: ok so this will be addressed by #886 I assume

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
johannesmayer: As @evgueni-ovtchinnikov predicted the problem is that `operator () (x,y,z,channel)` of the `ISMRMRD::Image<T>` only returns `T&` and has no const-overload return to `const T&` . They only provide a method `template <typename T> const T * Image<T>::getDataPtr()` that would allow const access. But to use that to access the data would require to compute a linear index from the arguments `(x,y,z,channel)` called for example `linear_index_x_y_z_c`.

So I see three possibilities:
- I leave it as it is and whenever our image class is involved I can not use  `const`. (Potentially for this PR this would be sufficient.)
- I write the code s.t. when I call the `Image<T>::operator ()` function I compute a linear index and  wherever I use the `operator()` replace that with `data = *(Image::getDataPtr() + linear_index_x_y_z_c)`
- I write a function `const SomeType ImageWrap::operator()` that uses the `getDataPtr()` method and I convert the arguments to a linear index to access the data. (annoying since then for every call you need to get the image header first for the image dimensions) 
- and the 4th possiblity is to add that `const` overload to ISMRMRD, but then we would need to build against their latest version...

DANAJK:
> I don't know about @ckolbPTB, but I will eternally shut up about any adjoint vs inverse and k-space density weighting, and I will admit that PET > MR
> 
> small_triumph_mind_over_matter
> (gif shows some low-res RPE conjugate gradient convergence without any density weighting from some weird blob to the correct image)

Cool. My understanding is that you can get density compensation wrong (or leave it out) and still arrive at the solution, though it might take longer to get there. The forward model (from reguarly spaced image pixels) to k-space is still correct.

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
johannesmayer: same problem with const-ness as with the `ISMRMRD::Image`

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
evgueni-ovtchinnikov: @johannesmayer: third option can compute the data starting pointer and dimensions in the `ImageWrap` constructor

src/xGadgetron/cGadgetron/tests/CMakeLists.txt
evgueni-ovtchinnikov: @KrisThielemans: if you mean the one that computes non-zero slice directions then yes, we just need to send the outputs of line 47 and 48 to some temporary files, e.g. `tmp1.h5` and `tmp2.h5`, and make CMake run `MR_TESTS_CPLUSPLUS` binary first with arguments `tmp1.h5` and `simulated_MR_2D_cartesian.h5` and then with `tmp2.h5` and `simulated_MR_2D_cartesian_Grappa2.h5`, and delete `tmp1.h5` and `tmp2.h5`. @johannesmayer: can you do that, please (and also remove printing from `edit_acq`)?

johannesmayer: sure, I am currently in the CMakeCrash course anyway so when this is finished I will be an expert on that matter!

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
evgueni-ovtchinnikov: @johannesmayer: perhaps a better idea is to inherit from `ISMRMRD::Image` and add `operator()(x,y,z,c) const` to the derived class

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_x.h
johannesmayer: done

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
johannesmayer: done

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
johannesmayer: I agree, however, I would strongly opt for the 99% instead of the 100% solution here and add a "fix" for this non-constness another time. 

Otherwise, I might have to do a PR from master to my branch one day...

KrisThielemans: Of course, let's not introduce a lot of hard work for ourselves, there's enough already!

Let's take option 1 now, but please create an issue for SIRF that we need this, and one for ISMRMRD, or even do a PR there. At some point they'll merge it, and then we can use it.

By the way, there's a 5th option: use `const_cast`, but it's quite ugly, so I wouldn't do that.

(funny thing is that I can't see where we use `ISMRMRD::Image`)

johannesmayer: We use it only as a `typedef ISMRMRD::Image<complex_float_t> CFImage` ... 
Then when we get the data the `operator ()` is used, e.g. [here](https://github.com/johannesmayer/SIRF/blob/25393fef1c85b38063e9158ad9d57f19fb1aef92/src/xGadgetron/cGadgetron/fourierencoding.cpp#L104)

KrisThielemans: ok. By the way, if it's ISMRMRD derived code, I wonder if those lines actually are just a straightforward copy, in which case they could be replaced by
```C++
std::copy(img.getDataPtr(), img.getDataPtr()+img.getNumberOfDataElements(), ci.getDataPtr());
```
that wouldn't work if `img.operator()` does some reordering of course.

johannesmayer: I thought of that already, but then it interfered with
```
Loop nx ny nz nc and do:
dst_img(nx, ny, nz, nc) = coilmap(nx, ny, nz, nc) * ((*ptr_src_img)(nx, ny, nz, 0))
```
so I had dropped it...
But actually that loop problem above is just coil-sensitivity specific, so I think I'll be able to make the Fourier transforms `const` by using your std::copy ...

KrisThielemans: great. Actually
```c++
std::copy(img.begin(). img.end(), ci.getDataPtr());
```
should do the trick I guess. Much neater.

The other example can be done as well with iterators.
```c++
auto coilmap_iter = coilmap.begin();
auto src_img_iter = ptr_src_img->begin();
auto dst_img_iter = dstImg.begin(); // or getDataPtr()
while (coilmap_iter != coilmap_end)
{ ....}
```
or indeed using [std::transform](https://www.cplusplus.com/reference/algorithm/transform/) and [std::multiplies](https://en.cppreference.com/w/cpp/utility/functional/multiplies)
```c++
std::transform(coilmap.begin(), coilmap,end(), ptr_src_image.begin(),
                        dst_img.begin(), // destination
                         std::multiplies<complex_float_t>());
```
or possibly with `foreach` since C++11.

johannesmayer: Alright looks nice! I'll fiddle that in ...

src/xGadgetron/cGadgetron/tests/CMakeLists.txt
johannesmayer: Alright, I added an executable and run it to preprocess the data and made the MR C++ tests depend on that target.
So far, this preprocessing only sets the acquisition orientation, but if in the future something comes up we could add it to that preprocessing.

johannesmayer:
Here's the content of the PR with a description of the changes in a presentation

[2021.03.25 MREncodingPR_JohannesMayer.pdf](https://github.com/SyneRBI/SIRF/files/6367036/2021.03.25.MREncodingPR_JohannesMayer.pdf)

johannesmayer:
I don't know how this latest error in the CI appears...
```
/home/runner/work/SIRF/SIRF/build/builds/SIRF/build/src/xGadgetron/cGadgetron/tests/MR_PROCESS_TESTDATA: error while loading shared libraries: libgadgetron_toolbox_cpucore_math.so.3.17: cannot open shared object file: No such file or directory
```
Is it because I link against `libgadgetron_toolbox_cpu.so` and that one links against the one with the `_math` suffix, and that one is not there?
But the `MR_PROCESS_TESTDATA` executable does not use any of those gadgetron libraries, it only links against our own `cgadtetron` library...
I use `find_library` to find the `libgadgetron_toolbox_cpu`, how am I supposed to find if any libraries that one depends on are also there?

KrisThielemans:
This is a run-time error, not a compilation/linking error. As I assume that the library should be in the same place as the others, I'm not sure why this error appears. Presumably `LD_LIBRARY_PATH` is set?

> how am I supposed to find if any libraries that one depends on are also there?

Ideally, Gadgetron would use a CMake export mechanism, which case this comes for free (i.e. CMake exports that one library depends on another). But if doesn't, it is up to us to figure that out.

But again, to me this doesn't look like the cause.

> Also the MR_PROCESS_TESTDATA executable does not use any of those gadgetron libraries, it only links against our own cgadtetron library...

sure, but the executable will want to load any shared libraries (as we don't do anything clever ourselves for loading only the libraries that we need at run-time), so it errors out.

johannesmayer:
Do you know why the builds fail with some download errors?
I recall some email saying something was off about a week ago but I cannot find it any more.
```
-- Downloading (if not already done)...
		Path:     /home/runner/work/SIRF/SIRF/data/examples/MR/zenodo/SIRF_recon.h5
		URL:      https://zenodo.org/record/3571228/files/SIRF_recon.h5?download=1
		MD5:      0d69892502d3a5cd220716a717cc2c26
-- [download 0% complete]
CMake Error at src/Synergistic/tests/CMakeLists.txt:24 (file):
  file DOWNLOAD HASH mismatch

    for file: [/home/runner/work/SIRF/SIRF/data/examples/MR/zenodo/SIRF_recon.h5]
-- Downloading (if not already done)...
		Path:     /home/runner/work/SIRF/SIRF/data/examples/MR/zenodo/dicom_as_nifti.nii
		URL:      https://zenodo.org/record/3571228/files/dicom_as_nifti.nii?download=1
		MD5:      c5f5b519e5c032560ead912713479b53
      expected hash: [0d69892502d3a5cd220716a717cc2c26]
        actual hash: [d41d8cd98f00b204e9800998ecf8427e]
             status: [22;"HTTP response code said error"]
```

KrisThielemans:
Just some networking problems between GitHub and Zenedo. Can someone restart the jobs please

KrisThielemans:
The first job failed because of antoehr network problem, but the second job failed due to a real errror I guess
https://github.com/SyneRBI/SIRF/pull/881/checks?check_run_id=2675644450#step:6:10874
```
[ 99%] Built target MR_PROCESS_TESTDATA
/home/runner/work/SIRF/SIRF/build/builds/SIRF/build/src/xGadgetron/cGadgetron/tests/MR_PROCESS_TESTDATA: error while loading shared libraries: libgadgetron_toolbox_cpucore_math.so.3.17: cannot open shared object file: No such file or directory
```

src/xGadgetron/cGadgetron/tests/CMakeLists.txt
```diff
@@ -17,13 +17,77 @@
...
+    message(ERROR "We could not find the ISMRMRD executable for the Shepp Logan Generation." )
+else()
+    add_custom_target(GENERATE_SIMULATED_TESTDATA ALL
+        COMMAND "")
+endif()
+
 add_library(MR_TESTS_CPP_AUXILIARY ${CMAKE_CURRENT_SOURCE_DIR}/mrtest_auxiliary_funs.cpp)
 target_link_libraries(MR_TESTS_CPP_AUXILIARY PUBLIC csirf cgadgetron)
 
+add_executable(MR_PROCESS_TESTDATA ${CMAKE_CURRENT_SOURCE_DIR}/mrtests_prep_testdata.cpp)
```
johannesmayer, line 67: This error is because the target `PROCESS_SIMULATED_TESTDATA` can't find the gadgetron library. 
The code used to build the home `MR_PROCESS_TESTDATA` used in the target does not use any gadgetron code, but it uses our `cgadgetron` library.

Do you have an idea of what is missing? Do I need to add another dependency as 
```add_dependencies(PROCESS_SIMULATED_TESTDATA MR_PROCESS_TESTDATA cgadgetron)```?
I thought that would have been covered by linking the `PROCESS_SIMULATED_TESTDATA` publicly against cgadgetron.

**Comment by johannesmayer** [comment]
Ok so switching the postprocessing of simulated data off, there is now an "error", due to the python tests checking against some numbers, which they fail to meet within the tolerance...

I don't know what happened to the testdata again, or what's going on there.

Is the policy with these tests that these numbers in the .txt files are magical (in a row with 0, 1, π, e, i,), and they must be achieved for the program to be considered running successfully, or can I just increase tolerance by a tad and then it will work?

Because all the C++ tests they run, and they do check for adjoint-ness of the acquisition model and the norm of the operator...

**Comment by KrisThielemans** [comment]
error is https://github.com/SyneRBI/SIRF/runs/2687040183?check_suite_focus=true#step:7:946
```
  File "/home/runner/work/SIRF/SIRF/src/xGadgetron/pGadgetron/tests/test2.py", line 34, in test_main
    test.check(input_data.norm())
  File "/home/runner/install/python/sirf/Utilities.py", line 299, in check
    self.check_if_equal_within_tolerance(expected, value, abs_tol, rel_tol)
  File "/home/runner/install/python/sirf/Utilities.py", line 333, in check_if_equal_within_tolerance
    raise ValueError(msg)
ValueError: +++ test 0 failed: expected 156.4, got 148 (tolerance 0.313)
```

I don't know what this test is (and have no time to check). Maybe @evgueni-ovtchinnikov can help. The tolerance seems quite high so. (@johannesmayer did you change this test/test-data at all?)

By the way, in future, please paste error output here. I know it's only 5 clicks to find out, but it'd save us time...

**Comment by evgueni-ovtchinnikov** [comment]
@johannesmayer you somehow managed to end up with wrong `test2.txt`, mine (dated 26 Mar 2021) has expected value 1.479937e+02 in the first line.

**Comment by johannesmayer** [comment]
Alriiiiight :) thank you... green light at the end of the tunnel!

**Comment by KrisThielemans** [comment]
Great! Could you try and address some of the codacy new issues? A lot of them are just trailing white-space, or unused imports so should be very easy to do. A few need more thinking and could be wrongly flagged (I didn't check the code at all). About half of them are due to our use of `import *` which we need to get rid of, but feel free to leave that one for later!

@evgueni-ovtchinnikov, any final comments on this? I don't have time to review.

**Review by evgueni-ovtchinnikov — APPROVED** [review]
- please make sure that all MR tests and **all demos** in `SIRF/examples/*/MR/*` work ok before merging
- merge your branch on the new VM first and check that all is well before pushing the post-merge master

**Comment by johannesmayer** [comment]
> Great! Could you try and address some of the codacy new issues? A lot of them are just trailing white-space, or unused imports so should be very easy to do. A few need more thinking and could be wrongly flagged (I didn't check the code at all). About half of them are due to our use of `import *` which we need to get rid of, but feel free to leave that one for later!

Yes!

> * please make sure that all MR tests and **all demos** in `SIRF/examples/*/MR/*` work ok before merging
> * merge your branch on the new VM first and check that all is well before pushing the post-merge master

Yes!

One more comment, I have only deactivated the re-running of the Shepp Logan Generator by switching the default value of a CMake option to off... Should I delete the code instead since if you run it crashes the automated tests?

**Comment by evgueni-ovtchinnikov** [comment]
@johannesmayer do not delete anything unless you know exactly what went wrong, otherwise you might be "papering over the crack" - keep it deactivated and make an issue

**Comment by johannesmayer** [comment]
> @johannesmayer do not delete anything unless you know exactly what went wrong, otherwise you might be "papering over the crack" - keep it deactivated and make an issue

The lines in the CMake file that do this are not on SIRF/master, so I would delete them in order not to introduce them with this PR when they don't run... I could add them in a separate PR once I understand what is going wrong.

**Comment by johannesmayer** [comment]
Alright so I progressed a bit with
* I corrected the Codacy issues except for the ones that are due to `import *`
* I build the branch on the a new 3.0.0 VM that I downloaded from Zenodo
* `ctest -VV -R MR_` ran through with all tests passing.
* I double checked running all scripts in `SIRF/examples/Python/MR/*` and everything worked fine 
* I can't run it in Matlab, so I can't tell what's going on there unfortunately :(

**Comment by KrisThielemans** [comment]
great. thanks!

@evgueni-ovtchinnikov ready to merge? Please feel free to press the button!

File: examples/Python/MR/rpe_recon.py [hunk 33]
```diff
@@ -0,0 +1,117 @@
...
+  -o <file>, --output=<file>  output file for simulated data
+  -e <engn>, --engine=<engn>  reconstruction engine [default: Gadgetron]
+  -n <bool>, --non-cart=<bool> run recon iff non-cartesian code was compiled
+                              [default: False]
+  --non-interactive           do not show plots
+'''
+
+## SyneRBI Synergistic Image Reconstruction Framework (SIRF)
+## Copyright 2015 - 2020 Rutherford Appleton Laboratory STFC
+## Copyright 2015 - 2017 University College London.
```
KrisThielemans, line 26: adjust

examples/Python/MR/rpe_recon.py
```diff
@@ -0,0 +1,117 @@
...
+
+    # locate the k-space raw data file
+    input_file = existing_filepath(data_path, data_file)
+    
+    # acquisition data will be read from an HDF file input_file
+    AcquisitionData.set_storage_scheme('memory')
+    acq_data = AcquisitionData(input_file)
+    
+    print('---\n acquisition data norm: %e' % acq_data.norm())
+
```
KrisThielemans, line 69: let's delete this. doesn't have any function here

examples/Python/MR/rpe_recon.py
```diff
@@ -0,0 +1,117 @@
...
+run_recon = str(args['--non-cart']) == 'True'
+
+def main():
+
+    # locate the k-space raw data file
+    input_file = existing_filepath(data_path, data_file)
+    
+    # acquisition data will be read from an HDF file input_file
+    AcquisitionData.set_storage_scheme('memory')
```
KrisThielemans, line 65: delete I guess

src/xGadgetron/cGadgetron/CMakeLists.txt
```diff
@@ -28,14 +28,52 @@
...
+                )
+find_library( GT_LOG NAMES gadgetron_toolbox_log
+                NAMES_PER_DIR
+                PATHS ${CMAKE_INSTALL_PREFIX}
+                )
+
+set (GT_LIBS ${GT_CPUCORE} ${GT_FFT} ${GT_NFFT} ${GT_LOG} )
+
+if(GT_CPUCORE AND GT_FFT AND GT_NFFT AND GT_LOG)
+    MESSAGE(STATUS "The Gadgetron toolboxes required for radial gridding were found. Non-caratesian encoding will be compiled.")
```
KrisThielemans, line 62:
```suggestion
    MESSAGE(STATUS "The Gadgetron toolboxes required for radial gridding were found. Non-cartesian encoding will be compiled.")
```

src/xGadgetron/cGadgetron/CMakeLists.txt
```diff
@@ -28,14 +28,52 @@
...
+    set(CGADGETRON_SOURCES ${CGADGETRON_SOURCES} noncartesian_encoding.cpp)
+
+    if(${CMAKE_VERSION} VERSION_LESS "3.12.0")
+      add_definitions(-DGADGETRON_TOOLBOXES_AVAILABLE)
+    else()
+      add_compile_definitions(GADGETRON_TOOLBOXES_AVAILABLE)
+    endif()
+else()
+    set(GADGETRON_TOOLBOXES_AVAILABLE FALSE)
+    MESSAGE(WARNING "The Gadgetron Toolboxes required for radial gridding were NOT FOUND. Non-caratesian encoding will NOT BE compiled.")
```
KrisThielemans, line 73:
```suggestion
    MESSAGE(WARNING "The Gadgetron Toolboxes required for radial gridding were NOT FOUND. Non-cartesian encoding will NOT BE compiled.")
```

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
KrisThielemans: still planning to do this? If not, create an issue for later

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/noncartesian_encoding.h
```diff
@@ -0,0 +1,142 @@
...
+    RPEFourierEncoding(): FourierEncoding() {}
+
+    virtual void forward(MRAcquisitionData& ac, const CFImage& img) const;
+    virtual void backward(CFImage& img, const MRAcquisitionData& ac) const;
+protected:
+    GadgetronTrajectoryType2D get_trajectory(const MRAcquisitionData& ac) const;
+
+};
+
+using namespace Gadgetron;
```
KrisThielemans, line 104: please do not use `using` statement in `.h` files.  It means we could have name conflicts. (Feel free to use them in the `.cpp` of course).

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/trajectorypreparation.h
```diff
@@ -0,0 +1,118 @@
...
+    virtual void set_trajectory(sirf::MRAcquisitionData& mr_acq) =0;
+    
+protected:
+
+    void update_acquisitions_info(sirf::MRAcquisitionData& mr_acq);
+
+    ISMRMRD::Encoding kspace_encoding_;
+    ISMRMRD::TrajectoryType traj_type_;
+
+    uint16_t traj_dim_;
```
KrisThielemans, line 65: does this need to be `protected`? Or can it be `private`? (the more `private`, the better)

src/xGadgetron/cGadgetron/tests/mrtests_prep_testdata.cpp
```diff
@@ -0,0 +1,51 @@
+/*
+CCP PETMR Synergistic Image Reconstruction Framework (SIRF)
+Copyright 2020 Rutherford Appleton Laboratory STFC
+
+This is software developed for the Collaborative Computational
+Project in Positron Emission Tomography and Magnetic Resonance imaging
+(http://www.ccppetmr.ac.uk/).
```
KrisThielemans, line 7: synerbi

KrisThielemans (changes requested):
some housekeeping comments.

can you please rename `fourierencoding.cpp` and `.h` to `FourierEncoding.*`? That's how the class is called. Sorry.

I suppose `noncartesian_encoding.h` then needs to be `NoncartesianEncoding.h`. Or actually `RPEFourierEncoding`? Same for `TrajectoryPreparation`.

Also, have a look at copyright statements. Are there soe lines in CHANGES.md and the documentation?

src/xGadgetron/cGadgetron/tests/CMakeLists.txt
KrisThielemans: you have
```
target_link_libraries(cgadgetron PUBLIC ${GT_LIBS})
```
so that should be fine. I'm a bit confused as ` error while loading shared libraries` is normally a run-time error I thought (not a linking error). can we reproduce this somewhere else than GHA?

src/xGadgetron/cGadgetron/tests/CMakeLists.txt
```diff
@@ -17,13 +17,77 @@
...
+add_dependencies(MR_PROCESS_TESTDATA GENERATE_SIMULATED_TESTDATA)
+target_link_libraries(MR_PROCESS_TESTDATA PUBLIC MR_TESTS_CPP_AUXILIARY csirf cgadgetron)
+
+if(RERUN_ISMRMRD_SHEPP_LOGAN)
+        add_custom_target(PROCESS_SIMULATED_TESTDATA ALL
```
johannesmayer, line 72: @KrisThielemans, as you mentioned [here](https://github.com/SyneRBI/SIRF/pull/881/files/1257f2fb1e01f730fb3605365d4b6a5ad0f00c67#r642557433) it is indeed a runtime error, since I run the executable during the build.

The logic was initially:
* problem: the ISMRMRD-simulated Shepp-Logan in the data repository is outdated, and the XML version in the header can lead to the program rejecting the test data as input
* solution: I run the Shepp-Logan simulation from `CMake` during the build to generate the test data on the fly
* next problem: then the geometry fields in the acquisitions were not populated, so we got all these warnings that the orientation vectors are zero.
* next solution: @evgueni-ovtchinnikov and I had written code that populates these fields after the simulation had finished, and I compiled this into an executable and run it after I generated the test data on the fly
* next problem: during this run it does not find this gadgetron library that I never assumed it would need.

Since the above first problem was now apparently solved by updating the data repository, I could just delete the part that re-runs the SheppLogan simulation and everything that follows, since it looks like we don't need it after all.

src/xGadgetron/cGadgetron/tests/CMakeLists.txt
```diff
@@ -17,13 +17,77 @@
...
+set(TESTDATA_OUTPUT_NAME_GRAPPA "simulated_MR_2D_cartesian_Grappa2.h5")
+
+set(TESTDATA_TEMP_PREFIX "testdata_tmp")
+
+if(EXE_ISMRMRD_SL AND RERUN_ISMRMRD_SHEPP_LOGAN)
+    add_custom_target(GENERATE_SIMULATED_TESTDATA ALL
+        COMMAND rm -f ${TESTDATA_OUTPUT_NAME_FULL}
+        COMMAND rm -f ${TESTDATA_OUTPUT_NAME_GRAPPA}
+        COMMAND env LD_LIBRARY_PATH=${CMAKE_INSTALL_PREFIX}/lib:$LD_LIBRARY_PATH ${EXE_ISMRMRD_SL} -r 2 -o "${TESTDATA_TEMP_PREFIX}_1.h5"
+        COMMAND env LD_LIBRARY_PATH=${CMAKE_INSTALL_PREFIX}/lib:$LD_LIBRARY_PATH ${EXE_ISMRMRD_SL} -a 2 -o "${TESTDATA_TEMP_PREFIX}_2.h5"
```
KrisThielemans, line 55: `LD_LIBRARY_PATH` is linux specific. Tricky problem related to the `SuperBuild` (not `SIRF`). See other comment on shared libraries.

src/xGadgetron/cGadgetron/tests/CMakeLists.txt
KrisThielemans: ok. thanks for that. This explains it all.

You run a SIRF executable during the build, but that depends on `LD_LIBRARY_PATH` etc being set correctly. But this is set via env_sirf.sh, which on GHA is only done after the build. On your local machine, you more than likely have it set before you build, so you don't see this problem.

creating an issue...

KrisThielemans: https://github.com/SyneRBI/SIRF-SuperBuild/issues/546

KrisThielemans: @johannesmayer what about editing
https://github.com/SyneRBI/SIRF/blob/0349fe6d16b8bb268378c5c37b2a4dee9cb55f0a/.github/workflows/build-test.yml#L67-L68
to insert `source ~/install/env_sirf.sh` before the build?

johannesmayer: I tried but it would not accept my access token when I tried to push to my PR branch after I modified the workflow....

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/trajectorypreparation.h
johannesmayer: inherited classes use it so I left it protected

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/noncartesian_encoding.h
johannesmayer: done

src/xGadgetron/cGadgetron/CMakeLists.txt
johannesmayer: done

evgueni-ovtchinnikov:
@johannesmayer ok to merge now or are you still working on this PR?

johannesmayer:
I updated the `Changes.md` describing what is new, and if the tests clear now I am good to merge.

src/xGadgetron/cGadgetron/tests/CMakeLists.txt
KrisThielemans: ?? There is nothing magical about that file, so maybe you had a temporary problem.

I'm assuming this problem is unresolved. If so, can you create a new issue or PR addressing only this problem? Otherwise we will just forget until it hits us again.

PR #887, opened 2021-04-09
Use memory storage only for MR data

evgueni-ovtchinnikov:
This PR makes preparations for impending MR encoding #881 that works for memory storage only.


examples/Python/MR/Gadgetron/run_all.py
```diff
@@ -30,6 +30,7 @@
     args = ''
     for a in range(1, narg):
         args += ' ' + sys.argv[a]
-    err = os.system(sys.executable + ' ' + i + args)
+    exe = 'python ' + i + args
```
KrisThielemans, line 33: This will fail on systems where the executable is not called `python` (such as our new VM, where it is `python3`). Instead, use quotes, see https://stackoverflow.com/questions/6977215/os-system-to-invoke-an-exe-which-lies-in-a-dir-whose-name-contains-whitespace

examples/Python/MR/acquisition_data.py
```diff
@@ -51,9 +51,9 @@
 slcs = int(args['--slices'])
 show_plot = not args['--non-interactive']
 
-AcquisitionData.set_storage_scheme('memory')
-scheme = AcquisitionData.get_storage_scheme()
-print('storage scheme: %s' % repr(scheme))
+#AcquisitionData.set_storage_scheme('memory')
+#scheme = AcquisitionData.get_storage_scheme()
+#print('storage scheme: %s' % repr(scheme))
```
KrisThielemans, line 56: remove

examples/Python/MR/fully_sampled_recon.py
```diff
@@ -55,7 +55,7 @@ def main():
 
     # all acquisition data generated by this script will be stored in memory
     # (the input data will remain in the input file)
-    AcquisitionData.set_storage_scheme('memory')
+#    AcquisitionData.set_storage_scheme('memory')
```
KrisThielemans, line 58: remove

examples/Python/MR/run_all.py
```diff
@@ -30,6 +30,7 @@
     args = ''
     for a in range(1, narg):
         args += ' ' + sys.argv[a]
-    err = os.system(sys.executable + ' ' + i + args)
+    exe = 'python ' + i + args
```
KrisThielemans, line 33: as above

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
```diff
@@ -40,7 +40,7 @@ limitations under the License.
 using namespace gadgetron;
 using namespace sirf;
 
-std::string MRAcquisitionData::_storage_scheme;
+//std::string MRAcquisitionData::_storage_scheme;
```
KrisThielemans, line 43: remove

KrisThielemans (changes requested):
I haven't tried this, nor did check the core lines. sorry.

Please also add a line in `CHANGES.md`.

I think we have to remove all the comments. They create clutter in the files. We have git to go back and recover them.

src/xGadgetron/cGadgetron/gadgetron_x.cpp
```diff
@@ -208,7 +210,7 @@ AcquisitionsProcessor::process(MRAcquisitionData& acquisitions)
 	}
 
 	uint32_t na = sptr_acqs->number();
-	//std::cout << na << '\n';
+	//std::cout << na << " acquisitions processed\n";
```
johannesmayer, line 213: Do these lines do anything, because they are comments, right?

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -435,14 +438,14 @@ namespace sirf {
 		}
 		static void init()
```
johannesmayer, line 439: This function does nothing any more, right? Is it still necessary?

johannesmayer (approved):
Just had two questions, but looks good to me.
I agree with Kris about deleting all lines instead commenting them.

examples/Matlab/MR/acquisition_data.m
```diff
@@ -36,8 +36,8 @@ function acquisition_data(engine)
 % all acquisition data generated by this script will be stored in memory
 % (the input data will remain in the input file)
 AD = MR.AcquisitionData();
-scheme = AD.get_storage_scheme();
-AD.set_storage_scheme('memory');
+%scheme = AD.get_storage_scheme();
+%AD.set_storage_scheme('memory');
```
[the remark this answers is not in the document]
evgueni-ovtchinnikov: I always comment out before deleting in case I delete something that actually should have stayed and do not notice it until much later, when restoring could be a lot of hassle. All these commented out lines will go before merge.

examples/Python/MR/Gadgetron/run_all.py
evgueni-ovtchinnikov: @KrisThielemans: Had to do it this way because neither the solution you quoted nor any other solutions I found on the Internet worked. Will think about yet another solution.

src/xGadgetron/cGadgetron/gadgetron_x.cpp
evgueni-ovtchinnikov: @johannesmayer of course they do not, this is a debugging print, which is often needed so I keep it commented out.

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
evgueni-ovtchinnikov: @johannesmayer no it is not, just forgot to remove

examples/Python/MR/Gadgetron/run_all.py
KrisThielemans: what did `sys.executable` report?

evgueni-ovtchinnikov: @KrisThielemans 'C:\Program' is not recognized as an internal or external command, operable program or batch file.

evgueni-ovtchinnikov: @KrisThielemans this works:
```
    exe = os.path.split(sys.executable)[1] + ' ' + i + args
    err = os.system(exe)
```

KrisThielemans: this is effectively like calling `os.path.basename`. I think it will fail when the directory is not on the path.

it's possible there are some issues with spaces and quotes. I think it's important to understand what `sys.executable` is exactly.

what did you try with the quotes?

evgueni-ovtchinnikov: @KrisThielemans "directory is not on the path" - do you mean Python executable directory? Well, then using Python would be very inconvenient, would it not?

Of what I was trying, I remember
```
    err = os.system('"' + exe + '"')
```
and (somebody suggested this weird solution specifically for Windows)
```
    err = os.system('""' + exe + '""')
```
and
```
    err = subprocess.call([exe])
```
all with
```
    exe = sys.executable + ' ' + i + args
```

KrisThielemans: >  "directory is not on the path" - do you mean Python executable directory? Well, then using Python would be very inconvenient, would it not?

you don't know how they call it. From jupyter or whatever, it might not be in the path. Or there might be another python in the path which conflicts.

Please try
```python
exe = '"' + sys.executable + '"  ' + i + args
```
the quotes need to be around the executable only

evgueni-ovtchinnikov: @KrisThielemans it worked, thanks!

KrisThielemans (approved):

PR #894, opened 2021-04-19
Closes #741

### LINKED ISSUE #741: Resample vs Resampler
We have `Reconstructor`, not `Reconstruct`. Naming was because it's an "object that can reconstruct".

`Resample` has its name because initially it didn't follow the `Reconstructor` pattern, but now it does.

Some other thoughts.
- At present `Reconstructor` doesn't follow the `Operator` type of naming (see #737) although in principle it could (`AcquisitionData`->`ImageData`). Not sure if that'd be beneficial, but I don't think it affects the naming in any case.
- In C++, you can have function objects (I guess in Python as well). Aliasing `resample.forward(z)`  to `resample(z)` would make the original `Resample` name more logical. `resample.backward(z)` has a ring to it as well.
- we can introduce type aliases etc for backwards incompatibility.

We should use consistent naming though. Sigh.

johannesmayer:
Refactored code and added `DEPRECATED` public child class on the C++ level.
Did not rename variable names.
Closes #741


KrisThielemans (changes requested):
I think best to use an alias for the type, as opposed to making a derived class (I guess that won't work without extra stuff as it'd have to forward constructors etc).
See e.g. https://subscription.packtpub.com/book/programming/9781786465184/1/ch01lvl1sec6/creating-type-aliases-and-alias-templates and https://en.cppreference.com/w/cpp/language/type_alias.

This is an old post on doing that with deprecation https://stackoverflow.com/questions/19791904/deprecate-templated-class-name-with-template-alias-type-alias-using-c11

I think we should do this for Python as well, but there it's simple: https://stackoverflow.com/questions/840969/how-do-you-alias-a-python-class-to-have-another-name-without-using-inheritance

We'll need changes in the Matlab files as well.

johannesmayer:
I'm sorry, but our PTB proxy server is killing me (like so often). I can not push or pull any more, so I cannot do the Python and Matlab corrections,  unfortunately.

KrisThielemans:
> I can not push or pull any more

ever again? Aaargh!

great that all checks worked. thanks!

johannesmayer:
ok pushing is back, so I quickly did the Python bit
The tests pass, but it does not test if the code using the old name is still functional, but it should be ...

KrisThielemans:
> The tests pass, but it does not test if the code using the old name is still functional, but it should be ...

yes. you've been working too hard 😄 . What about on your VM running an old demo with new SIRF? (i.e. just clone `master` somewhere and execute a relevant example)

johannesmayer:
I did do the whole thing with string replacements, I edited all the MATLAB files as well.

I just didn't do the deprecation bit yet in MATLAB.

KrisThielemans:
> I edited all the MATLAB files as well.

oops. I didn't even notice!

johannesmayer:
It seems to work, you can generate a `NiftyResample` object and it will give `type(NiftyResample)=NiftyResampler`.

However, it does not issue any deprecation warning and I can not say why.
Also Matlab-wise I have no idea how to do the aliasing and Google's giving me nothing.

src/Registration/pReg/Reg.py
```diff
@@ -1238,6 +1240,13 @@ def range_geometry(self):
         """Get range geometry."""
         return self.reference_image
 
+@deprecation.deprecated(details="The use of NiftyResample is discouraged and will be dropped. \
+                        Please use NiftyResampler instead.")
+
+class NiftyResample(object):
+    pass
```
KrisThielemans, line 1247: this definition will be removed by the assignment below I believe.

johannesmayer: That is correct, I guess, but I wrote it s.t. I could put the deprecation in front of it...because 
```
@deprecation.deprecated(details="The use of NiftyResample is discouraged and will be dropped. Please use NiftyResampler instead.")
NiftyResample = NiftyResampler
```
will generate an error.
On the other hand, none of the depreciation does anything, actually.
I can either remove both the deprecation, and the class definition, or you need to give me a hint on how to make this deprecation issue any warning :D .

KrisThielemans: sure. it's seems tricky to get deprecate a variable. possibly cannot be done.

let's just remove the deprecation then and the class def (less code is better, especially if it's confusing!), but please add a comment that contains the word "deprecation", such that we find it when searching for it...

Otherwise, good to go from me.

KrisThielemans (approved):

PR #895, opened 2021-04-20
xapyb pet

gfardell:
- There is no xapyb where a and b are mixed scalar and vector, therefore I've implemented this in python with the vector multiplication first before the scalar version of axpby is called. If we want to implement this at a lower level I am happy to, but it'll need it to be added to STIR too.
 - I have not implemented anything in matlab, I don't have a matlab to test the code and don't know matlab well at all, so it would be good if someone else could handle this
 -  I've added unit tests at the python level
- I haven't implemented the vector xapyb versions for gadgetron data containers
- Where python tries to call the vector version and this is not implemented the error is caught and python is used for the calculation instead however it still prints:
```
File: /home/sirfuser/devel/install/python/sirf/SIRF.py
Line: 254
check_status found the following message sent from the engine:
```
I'm not sure if there's a way to handle this try/except better.


src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -1626,6 +1626,37 @@ void NiftiImageData<dataType>::axpby(
...
+
+template<class dataType>
+void NiftiImageData<dataType>::xapyb(
+    const DataContainer& a_x, const DataContainer& a_a,
+    const DataContainer& a_y, const DataContainer& a_b)
+{
+    const NiftiImageData<dataType>& a = dynamic_cast<const NiftiImageData<dataType>&>(a_a);
+    const NiftiImageData<dataType>& b = dynamic_cast<const NiftiImageData<dataType>&>(a_b);
+    const NiftiImageData<dataType>& x = dynamic_cast<const NiftiImageData<dataType>&>(a_x);
+    const NiftiImageData<dataType>& y = dynamic_cast<const NiftiImageData<dataType>&>(a_y);
```
KrisThielemans, line 1645: could potentially throw unexpectedly if the type isn't ok. This is fine, but will confuse the user (certainly so a Python user).

should we cope with cases where the data-type isn't `NiftiImageData`? Might be quite hard. But if we don't, I think we should put the whole thing in a `try catch` block with some sensible error.

Also, in future, this can be shorter as
```c++
auto& a = dynamic_cast<const NiftiImageData<dataType>&>(a_a);
```

src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -1626,6 +1626,37 @@ void NiftiImageData<dataType>::axpby(
...
+
+    // If the result hasn't been initialised, make a clone of one of them
+    if (!this->is_initialised())
+        *this = *x.clone();
+
+    ASSERT(_nifti_image->nvox == x._nifti_image->nvox, "axpby operands size mismatch");
```
KrisThielemans, line 1651: is `ASSERT` one of our macros? Normally `ASSERT` only does checks in debug-mode, which isn't good enough for something called via Python.

src/common/SIRF.py
```diff
@@ -202,6 +202,8 @@ def add(self, other, out=None):
         return z
     def axpby(self, a, b, y, out=None, **kwargs):
         '''
+        This function has been deprecated and will be removed in a future release. Please use `sapyb` instead.
```
KrisThielemans, line 205: better to use https://python-deprecated.readthedocs.io/en/latest/tutorial.html?

KrisThielemans (changes requested):
this was quite a lot of work! thanks

evgueni-ovtchinnikov:
@KrisThielemans 
```
#define THROW(msg) throw LocalisedException(msg, __FILE__, __LINE__)
#define ASSERT(condition, msg) if (!(condition)) THROW(msg)
```

KrisThielemans:
ok!

src/Registration/cReg/NiftiImageData.cpp
gfardell: At the python end `sapyb` for vectors is in a try/except so it will default to python if there's an issue.

It does seem a bit strange restricting a and b to the same types, but I don't know how data is organised so whether a numpy array (for example) is compatible with nifty or stir arrays.

gfardell:
> this was quite a lot of work! thanks

400+ lines of code does seem a little excessive for basically some wrappers, but at least I understand SIRF a little now!

evgueni-ovtchinnikov (changes requested):

src/common/SIRF.py
```diff
@@ -211,30 +213,55 @@ def axpby(self, a, b, y, out=None, **kwargs):
...
-                (alpha.ctypes.data, self.handle, beta.ctypes.data, y.handle)
-        else:
+
+        if out is not None:
             assert_validities(self, out)
             z = out
-            try_calling(pysirf.cSIRF_axpbyAlt \
-                (alpha.ctypes.data, self.handle, beta.ctypes.data, y.handle, z.handle))
+        else:
+            z = self.copy()
```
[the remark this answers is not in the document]
gfardell: Maybe allocate is better if SIRF has a version of that? But by doing it in this order I don't have to duplicate the code below with `cSIRF_xapby_ss()` and `cSIRF_xapby_ss_Alt()` versions.

src/common/SIRF.py
```diff
@@ -211,30 +213,55 @@ def axpby(self, a, b, y, out=None, **kwargs):
...
+                tmp = self.multiply(a)
+                pysirf.cSIRF_xapyb_ss_Alt(tmp.handle, one.ctypes.data, y.handle, beta.ctypes.data, z.handle)
+            else:
+                assert_validities(self, b)
+                try:
+                    try_calling(pysirf.cSIRF_xapyb_vv_Alt(self.handle, a.handle, y.handle, b.handle, z.handle))
+                except:
+                    tmp = self.multiply(a)
+                    y.multiply(b, out=z)
+                    z.add(tmp, out=z)
```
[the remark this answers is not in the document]
gfardell: `axpby` previously covered vector and scalar methods, but with python computing the vector case, so I don't think it's covering more cases. But yes, it is a busy method.

In terms of the users passing a data-container to `out` that is also one of the inputs, I guess we'd have to check on the ID to stop this, but as the operation is pixel-wise anyway I assumed it would be safe to allow. It doesn't change the implementation apart from the python code order avoids an intermediary step writing to out.

I think CIL currently wraps the C++ code with ctypes and passes the type of a and b as parameters. I guess this would be an option to simplify the python and matlab ends. but `sapyb` needs to be consistent in behaviour across SIRF and CIL, I'm not sure we'd want to split it in to `sapyb_vv` or `sapyb_ss` at the python end as there's algorithms where a and b could be either.

KrisThielemans:
Turns out we have a `DEPRECATED` macro already in https://github.com/SyneRBI/SIRF/blob/843eb25760ae8e1e15bc816748d8b80d9c34f7d2/src/iUtilities/include/sirf/iUtilities/iutilities.h#L24

I like it better that it has its own file though. I think `SIRF_DEPRECATED` might be better than `DEPRECATED`, but that seems to have to be for a different PR

src/common/SIRF.py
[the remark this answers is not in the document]
evgueni-ovtchinnikov: using input container as `out` is ok - see e.g. `__iadd__`.

I still believe unnecessary copies can be avoided, but let us leave this for now - after all, as I mentioned, in SIRF we try to do all computation in C++, so `sapyb` will slim down eventually.

src/Registration/cReg/NiftiImageData.cpp
KrisThielemans: ok. Nevertheless, it's the C++ library that should give sensible errors as much as we can, not the "client" Python/MATLAB

KrisThielemans:
@gfardell @evgueni-ovtchinnikov I have no clear feeling if this is feasible to merge in the next few days, or if we better postpone it (which would be fine). can you comment?

gfardell:
> @gfardell @evgueni-ovtchinnikov I have no clear feeling if this is feasible to merge in the next few days, or if we better postpone it (which would be fine). can you comment?

I am updating it now and don't imagine it will take too long for it's current functionality, however if we want to delay it until the gadegtron xapyb is implemented I guess it would be more complete.

src/common/SIRF.py
gfardell: I'm happy to use this. But I'll need to add the `Python-Deprecated` to the requirements, could you point out where these live?

KrisThielemans: should be there already
https://github.com/SyneRBI/SIRF/blob/master/requirements.txt#L7

evgueni-ovtchinnikov:
@KrisThielemans I would postpone it - for one thing, I believe we should move the computation in Python `sapyb` to C++ and do Matlab and MR before merge. (Now that Johannes took over `Resampler` from me, I can take over `xapyb` from Gemma :-))

src/common/SIRF.py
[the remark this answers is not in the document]
gfardell: The only unnecessary copy is when `a` and `b` are of different types, and the user has passed `out`, as it will pre-calculate the `vector * vector` side. Is this where you mean?

There is a specific case if the users passes `y` and `out` as the same object, where `a` is a vector and `b` a scalar. Then an intermediate step `x.multiply(a, out=out)` would overwrite `y` before it was used. So without checking on the datacontainer IDs it seems safer that where it has to calculate in steps it writes initially for a temporary container rather than `out`. Of course if it was implemented in the C++ code it would be pixel-wise and not an issue.

gfardell:
I can't figure out why this stopped building. I can build and run the tests locally still.

KrisThielemans:
As the error is with recent changes in `master`, I think it is best to merge `master` on here again as the `DEPRECATED` stuff changed there. Presumably might give a conflict (although github doesn't think so).

gfardell:
I've made a stab at the matlab wrappers but I haven't managed to get a license from SCD yet to run and test it. I've also not used matlab much so have no clue on error handling, when I get a license I'll be able to figure that out.

src/common/csirf.cpp
```diff
@@ -136,7 +139,96 @@ void* ptr_z
...
+extern "C"
+void*
+cSIRF_xapyb_vv(
+const void* ptr_x, const void* ptr_a,
+const void* ptr_y, const void* ptr_b
+) {
+	try {
+		DataContainer& x =
+			objectFromHandle<DataContainer >(ptr_x);
+		DataContainer& y =
```
evgueni-ovtchinnikov, line 176: these two functions have substantial overlap - why not have one `cSIRF_xapyb` with extra `const char* args` parameter that may have values `"ss"` or `"vv"` (and potentially other values, making it future proof)?

this would also solve lesser problem of breaking our naming scheme `cSIRF_nameOfFunctionInCamelCase`

src/common/include/sirf/common/csirf.h
```diff
@@ -43,6 +43,14 @@ void* cSIRF_axpby(const PTR_FLOAT ptr_a, const void* ptr_x,
...
 void* cSIRF_axpbyAlt(const PTR_FLOAT ptr_a, const void* ptr_x,
 	const PTR_FLOAT ptr_b, const void* ptr_y, void* ptr_z);
+void* cSIRF_xapyb_ss(const void* ptr_x, const PTR_FLOAT ptr_a,
+	const void* ptr_y, const PTR_FLOAT ptr_b);
+void* cSIRF_xapyb_ss_Alt(const void* ptr_x, const PTR_FLOAT ptr_a,
+	const void* ptr_y, const PTR_FLOAT ptr_b, void* ptr_z);
+void* cSIRF_xapyb_vv(const void* ptr_x, const void*  ptr_a,
+	const void* ptr_y, const void* ptr_b);
+void* cSIRF_xapyb_vv_Alt(const void* ptr_x, const void*  ptr_a,
+	const void* ptr_y, const void* ptr_b, void* ptr_z);
```
evgueni-ovtchinnikov, line 53: see the above suggestion to replace these four functions with two functions with an extra `const char* args` parameter

evgueni-ovtchinnikov (changes requested):

KrisThielemans:
merged via another branch

PR #899, opened 2021-04-20
clean-up CHANGES/zenodo etc for v3.0.0

KrisThielemans:
reorganised, hopefully making more sense

@johannesmayer could you clarify what you mean with `Handling of coil images and sensitivities data simplified`? Is it backwards compatible? (Probably not on the C++ front)


CHANGES.md
```diff
@@ -1,32 +1,48 @@
...
+
+### New features
+* PET
+  - Addition of `sirf.STIR.ScatterEstimation` and `ScatterSimulation` to allow (non-TOF) scatter estimation in PET
+  - GE Signa PET/MR reading of listmode data, sinograms, normalisation and randoms support added.
+  - If STIR is at least version 5 or built from the master branch, [Georg Schramm's parallel (computing) projector](https://github.com/gschramm/parallelproj proj) is now made available from SIRF (use `AcquisitionModelUsingParallelproj`). This uses Joseph interpolation, but importantly can use your GPU (if CUDA was found during building).
+  - Implemented extraction of the operator representing the linear part of PET acquisition model and computation of its norm.
+  - When adding a shape to a `sirf.STIR.ImageData`, optionally give the number of times to sample a voxel. This is useful when the shape partially - but not completely - fills a voxel.
+  - If `storage_scheme` is set to `memory`, `PETAcquisitionData` allows direct modification, whereas before a copy would need to be created first. (Internally, it uses STIR `ProjDataInMemory`, instead of `ProjDataFromStream`).
+* MR
```
johannesmayer, line 23: The existing classes to handle coil image data were mostly deleted and the new ones were inherited from the `ISMRMRDImageData` class. This makes it backwards incompatible only on the C++ level.
On the Python we deleted `CoilImageData` in #646 and then @evgueni-ovtchinnikov restored the class in #786 but it's now names `CoilImagesData` (with an s inbetween), so I think it should not be backwards compatible (although only slightly so).

KrisThielemans: @evgueni-ovtchinnikov can you attempt to rephrase the `CHANGES` accordingly (it's on a branch which you can edit). Move it to the start (backwards incompatible) section.

KrisThielemans:
Current output of `git-fame`
image

@gfardell can you add yourself to `.zenodo.json` and `.mailmap`?

CHANGES.md
evgueni-ovtchinnikov: done

KrisThielemans: great. thanks. however (😉 ) as a user I don't know what it means. How does a user need to adapt her code? can you add 2 lines on that as well please?

evgueni-ovtchinnikov: more (hopefully) user-helpful explanation provided

PR #913, opened 2021-05-03
add Doxygen to CMake

KrisThielemans:
- changed Doxyfile to Doxyfile.in
- moved all doxygen files to doxygen folder
- copy CMake lines from STIR

This introduces a new target `RUN_DOXYGEN`, and an option `BUILD_DOCUMENTATION` (if `ON`, building the project will now run doxygen and install the files).


KrisThielemans:
@evgueni-ovtchinnikov maybe you could give this a go. Either set `BUILD_DOCUMENTATION=ON`, or do `make RUN_DOXYGEN`, (or build the target in VC if you have `doxygen` installed)

evgueni-ovtchinnikov:
@KrisThielemans builds ok except for `AcquisitionModel` and `NiftyMoMo` formulas that use latex

I vaguely remember removing latex formatting - somebody must have put it back. Should I remove it again? `AcquisitionModel` formulas are pretty simple and will not be affected much - cannot say the same for `NiftyMoMo` ones though, as they use partial derivatives.

KrisThielemans:
For me this works ok (on the VM, after installing a load of Latex stuff), see for `PETAcquisitionModel`:
image

Weirdly I don't get any NiftyMoMo output in my doxygen (which I don't find too surprising as the namespace isn't documented etc).

Is the above behaviour specific to this PR, or just something that you have when running `doxygen` on SIRF?

evgueni-ovtchinnikov:
For me, latex formatting in comments was always a problem. The difference that this PR made is that Visual Studio build now fails on the doxygen project:
```
1>'latex' is not recognized as an internal or external command,
1>operable program or batch file.
1>CUSTOMBUILD : error : Problems running latex. Check your installation or look for typos in _formulas.tex and check _formulas.log!
```
It would be better to have this as a warning. If that is not possible, then I would remove latex stuff from the doxygen comments. (At any rate, e.g. the formula `y = A(x)` hardly needs latex.)

KrisThielemans:
There's a possibility to remove the requirement for latex with doxygen formulas by enabling some MATH_JAX stuff (a search will tell you) but that can't be easy either.

If `BUILD_DOCUMENTATION=OFF`, Visual Studio won't run it, so I guess that isn't a huge problem.

I think we could use a bit more math here and there, so therefore suggest can leave this  as-is, and just make sure we create the doxygen on systems that have latex installed.

Issue #915, opened 2021-05-06
Proper solution to #914


evgueni-ovtchinnikov:
#914 can be urgently quick fixed in Python for now. However, proper not urgent solution is to provide access to existing C++ image and acquisition templates in Python so that they will become automatically available in the linear part of the acquisition model.

PR #916, opened 2021-05-06
adds img_templ and acq_templ to linear acquisition model

### LINKED ISSUE #914: Python: get_linear_acquisition_model does not contain acq_templ and img_templ
For CIL we require the members `acq_templ` and `img_templ` to be part of the acquisition model. Once we use `get_linear_acquisition_model` these are not part of the new python object.

A quick fix is to add the members to the new `AcquisitionModel`.

paskino:
closes #914

I tested on the CIL1 paper reconstruction.


evgueni-ovtchinnikov (approved):

evgueni-ovtchinnikov:
ok as a quick fix pending #915

PR #929, opened 2021-05-20
clarify the func set_time_interval in doc

szho42:
make it explict in the documentation that the function set_time_interval in Listmode2Sinograms takes time intervals in seconds.


KrisThielemans:
thanks for that!

PR #934, opened 2021-06-01
Constructor for MR image data from MR acquisition data

### LINKED ISSUE #933: MR Image from MR Acquisition Data
Generate an empty image from MR Acquisition data. 
This is required for example in setting up an acquisition model without having to perform any reconstructions beforehand.

This addresses #824 on the MR side.

johannesmayer:
Fixes #933

The following operation is performed:
* for each consistent subsets of k-space a `CFImage` is created and appended. 
* the header is populated based on the first acquisition in each subset

While there are fields in the ISMRMRD XML header of the whole rawdata file comprising all acquisitions, the image headers are populated from the `Acquisition` object to include the geometry information that can vary between different subsets of k-space, for example for different 2D slices.
It is assumed that the `Acquisition` objects in each subset contain the same header information.

### DIFF
--- a/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
+++ b/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
@@ -881,6 +881,17 @@ namespace sirf {
 
 		GadgetronImagesVector() : images_()
 		{}
+
+		/*!
+		\ingroup MR
+		\brief Constructor for images from MR Acquisition data.
+
+		The images are generated with the dimensions given in the recon-space of the 
+		MRAcquisitionData's ISMRMRD header information.
+		The geometry information and header of the individual images are populated
+		based on all consistent subsets of acquisitions in the MRAcquisition object.
+		The images can also be created coil-resolved.
+		*/
 		GadgetronImagesVector(const MRAcquisitionData& ad, const bool coil_resolved=false);
         GadgetronImagesVector(const GadgetronImagesVector& images);
 		GadgetronImagesVector(GadgetronImagesVector& images, const char* attr,
--- a/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
+++ b/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
@@ -1074,8 +1074,8 @@ namespace sirf {
     public:
 
         CoilSensitivitiesVector() : GadgetronImagesVector(){}
-		CoilSensitivitiesVector(MRAcquisitionData& ad, const bool coil_resolved=true) :
-			GadgetronImagesVector(ad, coil_resolved){}
+		CoilSensitivitiesVector(MRAcquisitionData& ad) :
+			GadgetronImagesVector(ad, true){}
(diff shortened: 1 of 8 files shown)


src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -876,7 +876,7 @@ namespace sirf {
 
 		GadgetronImagesVector() : images_()
 		{}
-		GadgetronImagesVector(const MRAcquisitionData& ad);
+		GadgetronImagesVector(const MRAcquisitionData& ad, const bool coil_resolved=false);
```
KrisThielemans, line 879: some doxygen doc please

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -1058,6 +1058,8 @@ namespace sirf {
     public:
 
         CoilSensitivitiesVector() : GadgetronImagesVector(){}
+		CoilSensitivitiesVector(MRAcquisitionData& ad, const bool coil_resolved=true) :
+			GadgetronImagesVector(ad, coil_resolved){}
```
KrisThielemans, line 1062: would it make more sense to have
```suggestion
		CoilSensitivitiesVector(MRAcquisitionData& ad) :
			GadgetronImagesVector(ad, /*coil_resolved=*/true){}
```

KrisThielemans (changes requested):
Obviously I have no idea if this is correct, but you know what you're doing so I trust you!

some tiny cosmetics therefore only. Do check the copyright statements as well, and add something to CHANGES.md

KrisThielemans:
Please addd a test for set-up acquisition model (and maybe even basic recon)

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
johannesmayer: done

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
johannesmayer: done

KrisThielemans (approved):
ready to merge from my perspective. @johannesmayer ?

johannesmayer:
I have found that in the special case of an empty `AcquisitionData` object an exception will be thrown.
So there is the option to
* leave it that way, and say if you don't have any acquired data it makes no sense to set up images from them.
* add some changes to use the k-space dimensions defined in the header to make images, but these won't contain geometry information? 

Which would you prefer?

KrisThielemans:
I think best to use option 1 (throw an exception). Better safe than sorry.

Of course, this will mean we need to be able to create an `AcquisitionData` from scratch...

KrisThielemans:
let us know if this needs any more changes then or not. I'd prefer to merge it soon such that it can be used in the SIRF-Exercises

johannesmayer:
Ok then let's go with option 1, and I think it is good to merge.

PR #938, opened 2021-06-02
Fix CSM Artefacts

johannesmayer:
For cartesian acquisition data that are undersampled, the reconstructions underlying the coil sensitivities computation can contain high-frequency artefacts that will propagate to the coil maps and hence to the resulting reconstructions.

The `Acquisition` objects' header, however, contain flags that label whether the acquisitions are suitable to be used for coil calibration. These flags are given by: 
```
ISMRMRD_ACQ_IS_PARALLEL_CALIBRATION             = 20
ISMRMRD_ACQ_IS_PARALLEL_CALIBRATION_AND_IMAGING = 21,
```

This PR uses only the subset of `Acquistions` that have these flags set for cartesian trajectories.
If no `Acquisition` contains these two flags, the entire set of raw data is employed in the coil sensitivity estimation.

### DIFF
--- a/examples/Python/MR/grappa_and_steepest_descent.py
+++ b/examples/Python/MR/grappa_and_steepest_descent.py
@@ -144,8 +144,8 @@ def main():
             print('pylab not found')
 
     if output_file is not None:
-      print('writing to %s' % output_file)
-      image_data.write(output_file)
+        print('writing to %s' % output_file)
+        image_data.write(output_file)
 
 try:
     main()
--- a/examples/Python/MR/grappa_and_steepest_descent.py
+++ b/examples/Python/MR/grappa_and_steepest_descent.py
@@ -106,7 +106,7 @@ def main():
     # image_data
     acq_model = AcquisitionModel(preprocessed_data, image_data)
     acq_model.set_coil_sensitivity_maps(csms)
-    
+
     start_val = 0.001
     image_data = image_data.fill(start_val + start_val * 1j)
 
--- a/examples/Python/MR/grappa_and_steepest_descent.py
+++ b/examples/Python/MR/grappa_and_steepest_descent.py
@@ -106,6 +106,9 @@ def main():
     # image_data
     acq_model = AcquisitionModel(preprocessed_data, image_data)
     acq_model.set_coil_sensitivity_maps(csms)
+    
+    start_val = 0.001
+    image_data = image_data.fill(start_val + start_val * 1j)
 
     res = numpy.ndarray((niter,))
     scale = 1 # reconstructed image and phantom may have different scale,
--- a/examples/Python/MR/grappa_and_steepest_descent.py
+++ b/examples/Python/MR/grappa_and_steepest_descent.py
@@ -108,7 +108,7 @@ def main():
     acq_model.set_coil_sensitivity_maps(csms)
 
     start_val = 0.001
-    image_data = image_data.fill(start_val + start_val * 1j)
+    image_data.fill(start_val + start_val * 1j)
 
     res = numpy.ndarray((niter,))
     scale = 1 # reconstructed image and phantom may have different scale,
--- a/src/xGadgetron/pGadgetron/Gadgetron.py
+++ b/src/xGadgetron/pGadgetron/Gadgetron.py
@@ -577,10 +577,6 @@ class CoilSensitivityData(ImageData):
 
         dcw = compute_kspace_density(csm_data)
         csm_data = csm_data * dcw
-        try_calling(pygadgetron.cGT_computeCoilImages(self.handle, csm_data.handle))
-
-        dcw = compute_kspace_density(csm_data)
-        csm_data = csm_data * dcw
 
         if method_name == 'Inati':
             try:
--- a/src/xGadgetron/pGadgetron/Gadgetron.py
+++ b/src/xGadgetron/pGadgetron/Gadgetron.py
@@ -1619,12 +1619,18 @@ def extract_calibration_data( ad ):
     ref_flag = 20
     ref_img_flag = 21
 
-    subset_idx = []
+    subset_idx = numpy.array([])
     for i in range(ad.number()):
         a = ad.acquisition(i)
         flags = decode_ismrmrd_flag(a.flags())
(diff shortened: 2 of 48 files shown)


examples/Python/MR/grappa_and_steepest_descent.py
```diff
@@ -107,6 +107,9 @@ def main():
     acq_model = AcquisitionModel(preprocessed_data, image_data)
     acq_model.set_coil_sensitivity_maps(csms)
 
+    start_val = 0.001
+    image_data = image_data.fill(start_val + start_val * 1j)
```
KrisThielemans, line 111:
```suggestion
    image_data.fill(start_val + start_val * 1j)
```

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -1590,3 +1601,45 @@ def calc_rpe_dcw(ad):
...
     return dcw
+
+def extract_calibration_data(ad):
+    '''
+    Function to extract all Acquisitions from an AcquisitionData object that
+    carry the coil calibration flag.
+    '''
+
+    if not isinstance(ad, AcquisitionData):
+        raise AssertionError("The trajectory you asked for is not among the available trajectoryies")
```
KrisThielemans, line 1612:
```suggestion
        raise AssertionError("The trajectory you asked for is not among the available trajectories")
```

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -1590,3 +1601,45 @@ def calc_rpe_dcw(ad):
...
+        acq = ad.acquisition(i)
+        flags = decode_ismrmrd_flag(acq.flags())
+        if ref_flag in flags or ref_img_flag in flags:
+            subset_idx = numpy.append(subset_idx, i)
+
+    if subset_idx.size > 0:
+        calib_data = ad.get_subset(subset_idx)
+        return calib_data
+    else:
+        print("Warning, there are no calibration data in this dataset. We use them all for CSM computation.")
```
KrisThielemans, line 1628:
```suggestion
        print("Warning, there are no calibration data in this dataset. We use them for CSM computation.")
```
but ideally you add something about what the consequences are (if easy?)

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -1590,3 +1601,45 @@ def calc_rpe_dcw(ad):
...
+    '''
+    Function to extract all Acquisitions from an AcquisitionData object that
+    carry the coil calibration flag.
+    '''
+
+    if not isinstance(ad, AcquisitionData):
+        raise AssertionError("The trajectory you asked for is not among the available trajectoryies")
+
+    ref_flag = 20
+    ref_img_flag = 21
```
KrisThielemans, line 1615: can you add some of the comments if what you wrote in the PR description here?

KrisThielemans (changes requested):

evgueni-ovtchinnikov (changes requested):
Far too much Python coding, I am afraid - must be done in C++, otherwise we will have to code it all again in Matlab (Octave, Julia, ...). Ideally, all Python interface methods must just prepare and make calls to C interface functions and check their return status.

I would suggest adding `int calibration` argument to `MRAcquisitionData::get_subset` so that for a non-zero value of this argument only the calibration area acqusitions are returned.

KrisThielemans:
> Far too much Python coding, I am afraid - must be done in C++, otherwise we will have to code it all again in Matlab (Octave, Julia, ...). Ideally, all Python interface methods must just prepare and make calls to C interface functions and check their return status.
> 

Pretty much all of the CSM estimation code is Python anyway. So I guess you're mostly/only taking about `extract_calibration_data`? That does seem a prime candidate for a C++ member, but  while I heartily agree with that statement, but I think we have to accept contributions in Python (or MATLAB), and then convert them to C++ in a next PR.

> I would suggest adding `int calibration` argument to `MRAcquisitionData::get_subset` so that for a non-zero value of this argument only the calibration area acqusitions are returned.

I don't think that'd work. `get_subset` returns some k-space lines I guess. Calibration data are just a few of those. I would just add `extract_calibration_data` (later).

evgueni-ovtchinnikov:
@KrisThielemans 

> Pretty much all of the CSM estimation code is Python anyway.

This is largely due to the option to use Inati method from `ismrmrdtools` - without it, Johannes' methods `__calc_from*` would just call respective `cGT_computeCoilSensitivities*` interface functions.

> I don't think that'd work. get_subset returns some k-space lines I guess. 

It would just take adding
```
if (!calibration || acq.isFlagSet(ISMRMRD::ISMRMRD_ACQ_IS_PARALLEL_CALIBRATION) ||
acq.isFlagSet(ISMRMRD::ISMRMRD_ACQ_IS_PARALLEL_CALIBRATION_AND_IMAGING))
```
in front of `subset.append_acquisition(acq);` in `MRAcquisitionData::get_subset`.

KrisThielemans:
> > I don't think that'd work. get_subset returns some k-space lines I guess.
> 
> It would just take adding in `MRAcquisitionData::get_subset`
> ...

I meant that it might not make logical sense, but I'll leave that to @johannesmayer to decide :-)

johannesmayer:
Okay, I agree that @evgueni-ovtchinnikov is right... 
I shifted it all to the C++ level doing:

* `MRAcquisitionData::keep_flagged_acquisitions()` is a method that you can call with a vector of ISMRMRD flags as an argument. It will keep only the acquisitions that have the flag in the container (except if none of them has the flag, then it keeps them all).
* Now only the coil computation code checks for a cartesian trajectory and calls the method with the two calibration flags on the data used for coil sensitivity computation.

src/xGadgetron/pGadgetron/Gadgetron.py
johannesmayer: Done

src/xGadgetron/pGadgetron/Gadgetron.py
johannesmayer: Done by deletion.

PR #951, opened 2021-06-16
add CUDA includes if Gadgetron built with CUDA

paskino:
When Gadgetron is built with CUDA, `gadgetron_x.cpp` needs to have those includes.


KrisThielemans:
@paskino, Feel free to merge once GHA passes

PR #956, opened 2021-06-16
Shifted setup of geometry to the end of the MR Image Constructor

johannesmayer:
I had noticed that I call the set up the metadata before actually filling the image vector.
This will do it after the image vector is full.


KrisThielemans (commented):
no longer call `set_up_geom_info()`? sounds dangerous

KrisThielemans:
I have no idea what the consequences of this PR are. required for 3.1? i.e. bug fix?

johannesmayer:
That is called already inside of `set_meta_data()` so I didn't call it explicitly

PR #958, opened 2021-06-17
Fix flag checking for CSM computation

johannesmayer:
This is a fix of the shift of the last CSM fix into the C++ layer (I don't know what the hell I had tested there).

The strategy is the same:
- `CoilImagesVector::calculate(MRAcquisitionData& )` extracts a subset of the acquisitions that carry a flag that marks them as coil calibration data.
- if none of the acquisitions has that flag, then all acquisitions are used.

The problem with the current code is that the `ISMRMRD::Acquisition::isFlagSet()` method behaves unexpectedly and I don't know why. I will open an issue for it. Right now I use a different approach to check for which flags are set.

New things:

*I wrote a method `CoilImagesVector::extract_calibration_data` 
*I check if the flags are set without the `ISMRMRD` methods

### DIFF
--- a/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
+++ b/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
@@ -1730,9 +1730,13 @@ std::unique_ptr<MRAcquisitionData> CoilImagesVector::extract_calibration_data(co
 
     if(ad.get_trajectory_type() == ISMRMRD::TrajectoryType::CARTESIAN)
     {   
-        std::vector<int> flagged_positions = ad.get_flagged_acquisitions_index(calibration_flags);
+        std::vector<int> idx_calib_acquisitions = ad.get_flagged_acquisitions_index(calibration_flags);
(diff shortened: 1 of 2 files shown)


KrisThielemans:
test failures I'm afraid. @johannesmayer can you fix them? @evgueni-ovtchinnikov can you then have a quick look and merge? We can fix behaviour afer 3.1-rc.1 (if necessary!)

johannesmayer:
The failures are strange complaining about dependencies :
```
# SIRF external dependencies
apt-get update -qq
E: The repository 'https://packages.microsoft.com/ubuntu/20.04/prod focal Release' no longer has a Release file.
```

johannesmayer:
After merging master again it seems to go past the dependencies now.

KrisThielemans:
probably a GHA issue. let's see

johannesmayer:
ok went through

KrisThielemans:
@evgueni-ovtchinnikov ok to merge? Just do so!

PR #959, opened 2021-06-18
Use ISMRMRD methods in flag checking

johannesmayer:
This PR 
- uses ISMRMRD methods to check if acquisitions have a certain flag set or not instead of a manual implementation.
- renamed a variable according to the conventions of adding the `uptr_` prefix for unique pointers.


KrisThielemans (approved):

KrisThielemans:
@evgueni-ovtchinnikov please merge if you're happy

PR #962, opened 2021-06-21
get_info() consistency fixes

### LINKED ISSUE #829: get_info() improvements
After #815, I think we need a few others
.
- [x] deprecate the C++ `print_info` function, in the future reducing our code just a little bit more.
- [x] check the docstrings for `get_info()`? We have a number of them (great!) but most don't seem to say they return a string. 
- [x] does `MRAcquisitionData::get_info()` return something else? If so, we need to make that very clear in the doc, and I suggest to rename that function (probably in 2 stages: in next major release providing a duplicate with `get_info()` writing a warning), in the one after that let `get_info` return a string as well)
- [x] STIR has `parameter_info`, so STIR derived classes in SIRF have that as well. I think nice to provide in the sirf.stir classes `get_info()` as well (which just calls `parameter_info()` such that all SIRF classes have the same name.

### LINKED ISSUE #947: MR Acquisition*.parameter_info() doc incomplete
`AcquisitionData` doesn't give any clue what parameters I should pass
https://github.com/SyneRBI/SIRF/blob/a6573fc2b252e380ad00290dbd3b4d360940b343/src/xGadgetron/pGadgetron/Gadgetron.py#L923-L925
and `ImageData` talks about `Image` but that still doesn't tell me what to do
https://github.com/SyneRBI/SIRF/blob/a6573fc2b252e380ad00290dbd3b4d360940b343/src/xGadgetron/pGadgetron/Gadgetron.py#L307-L310

I only checked the Python doc.

KrisThielemans:
another attempt

closes #829
closes #947 (once I know what name to use)

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -11,7 +11,8 @@
 
 * SIRF Python interface
   - `range_geometry` and `domain_geometry` methods of `AcquisitionModel` classes, required by CIL algorithms, now obtain data via respective C++ `AcquisitionModel` classes accessors, in line with our strategy of keeping interface code minimal
-
+  - `sirf.Gadgetron.AcquisitionData.get_info` was renamed to `get_ISMRMRD_info` to avoid
+    confusion with the other `get_info()` methods that return a string. (`get_info` still works but issues a deprecation warning).
 * Build system
   - fix bug with older CMake (pre-3.12?) that the Python interface was not built
   [#939](https://github.com/SyneRBI/SIRF/issues/939).
--- a/doc/UserGuide.md
+++ b/doc/UserGuide.md
@@ -285,7 +285,7 @@ An engine-specific image data container class for data representing 3D objects. 
     as_array            Returns the object data as an array. 
     read_from_file      Reads the image data from file.
     get_ISMRMRD_info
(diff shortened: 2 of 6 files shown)


DANAJK:
I'm not clear what I am reviewing. If you wish to revert, that is fine for me for now.

Generally my preference would be to get a structure or class with all the relevant information - similar to `dicominfo` or `niftiinfo` in MATLAB for a structure, or the ismrmrd class https://github.com/ismrmrd/ismrmrd/tree/master/matlab/%2Bismrmrd  

A structure is self-documenting (look at the field names) and field values can be used directly in calculations or the fields can be converted to strings for printing.

KrisThielemans:
> I'm not clear what I am reviewing. If you wish to revert, that is fine for me for now.

sorry. I only pushed the MR stuff now. got caught in a meeting!

> Generally my preference would be to get a structure or class with all the relevant information

yes, I guess this is in `sirf.Gadgetron.Image` etc but I don't understand this well enough. In any case, that's too ambitious for this PR 😉

ckolbPTB:
Looks fine to me.

PR #966, opened 2021-06-22
STIR.IterativeReconstructor get/set_current_estimate now clone

### LINKED ISSUE #965: STIR.IterativeReconstructor.set_current_estimate stores a reference
When doing
```python
recon.set_current_estimate(initial_image)
recon.process()
reconstructed_image=recon.get_current_estimate()
```
the reconstruction updates the `initial_image` variable. Also, re-running a reconstruction but assigning to a different variable would still update `reconstructed_image`. I don't think anyone will expect this.

Reason is https://github.com/SyneRBI/SIRF/blob/ce6c26c6de6ad6b11a57b4b12fc97170ef4ad26e/src/xSTIR/pSTIR/STIR.py#L2470-L2473 and https://github.com/SyneRBI/SIRF/blob/ce6c26c6de6ad6b11a57b4b12fc97170ef4ad26e/src/xSTIR/pSTIR/STIR.py#L2482-L2484
(and equivalent in MATLAB). I think these should be using `clone()`. Less efficient, but less surprising.

@ckolbPTB would you agree?

KrisThielemans:
It was surprising that they stored/returned a reference to the image, causing unexpected changes.

Fixes #965

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -9,6 +9,10 @@
   - `CoilSensitivitiesVector` class now has forward and backward method using the encoding classes getting rid of the duplicate FFT code used to compute `coil sensitivities` from `MRAcquisitionData`.
   - Added constructor for `GadgetronImagesVector` from `MRAcquisitionData`. This allows setting up an MR acquisition model without having to perform a reconstruction first. 
 
+* PET/STIR
+   - iterative reconstructors `set_current_estimate` and `get_current_estimate` now create a clone to avoid surprising modifications of arguments. The old behaviour of `set_current_estimate` can still be achieved by `set_estimate`.<br>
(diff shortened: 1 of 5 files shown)


KrisThielemans:
Matlab still to do

KrisThielemans:
@ckolbPTB any opinion on the name for the "old" functionality? I've called it `set_estimate` (i.e. "this is the image that will be estimated"), but maybe that's not very clear.

KrisThielemans:
> @ckolbPTB any opinion on the name for the "old" functionality? I've called it `set_estimate` (i.e. "this is the image that will be estimated"), but maybe that's not very clear.

I've just disabled this functionality for now, and left the code in as comments.

KrisThielemans:
this should be ready.

PR #977, opened 2021-06-29
Small fixes

### LINKED ISSUE #970: VoxelisedGeometricalInfo::operator==() uses == to compare floats
The comparison between `float` numbers is made using the `==` operator.

```
virtual bool operator==(const GeometricalInfo<num_dimensions, num_dimensions>& gi) const
{
const VoxelisedGeometricalInfo& vgi = (const VoxelisedGeometricalInfo&)gi;
return
	_offset == vgi.get_offset() &&
	_spacing == vgi.get_spacing() &&
	_size == vgi.get_size() &&
	_direction == vgi.get_direction();
}
```
This sometimes leads to a failing `GadgetronImagesVector::reorient()` method even though the actual geometrical information is the same up to `1e-5 mm`.

evgueni-ovtchinnikov:
Fixes #970 and makes few other small corrections/amendments.

### DIFF
--- a/src/common/include/sirf/common/GeometricalInfo.h
+++ b/src/common/include/sirf/common/GeometricalInfo.h
@@ -90,11 +90,12 @@ public:
 	virtual bool operator==(const GeometricalInfo<num_dimensions, num_dimensions>& gi) const
 	{
 		const VoxelisedGeometricalInfo& vgi = (const VoxelisedGeometricalInfo&)gi;
+		float eps = 1e-4;
 		return
-			_offset == vgi.get_offset() &&
-			_spacing == vgi.get_spacing() &&
+			near_(_offset, vgi.get_offset(), eps) &&
+			near_(_spacing, vgi.get_spacing(), eps) &&
 			_size == vgi.get_size() &&
-			_direction == vgi.get_direction();
+			near_(_direction, vgi.get_direction(), eps);
 	}
 	virtual bool operator!=(const GeometricalInfo<num_dimensions, num_dimensions>& gi) const
 	{
@@ -117,6 +118,34 @@ private:
 	Spacing _spacing;
 	Size _size;
 	DirectionMatrix _direction;
+	static bool near_(const Coordinate& x, const Coordinate& y, float eps)
+	{
+		float s = 0;
+		float t = 0;
+		for (int i = 0; i < num_dimensions; i++) {
+			float xi = x[i];
+			float yi = y[i];
+			s = std::max(s, std::abs(xi));
+			s = std::max(s, std::abs(yi));
+			t = std::max(t, std::abs(xi - yi));
+		}
+		return t <= eps * s;
+	}
+	static bool near_(const DirectionMatrix& x, const DirectionMatrix& y, float eps)
+	{
+		float s = 0;
+		float t = 0;
+		for (int i = 0; i < num_dimensions; i++) {
+			for (int j = 0; j < num_dimensions; j++) {
+				float xij = x[i][j];
+				float yij = y[i][j];
+				s = std::max(s, std::abs(xij));
+				s = std::max(s, std::abs(yij));
+				t = std::max(t, std::abs(xij - yij));
+			}
+		}
+		return t <= eps * s;
+	}
 };
 
 typedef GeometricalInfo<3, 3> GeometricalInfo3D;
--- a/src/common/include/sirf/common/GeometricalInfo.h
+++ b/src/common/include/sirf/common/GeometricalInfo.h
@@ -90,7 +90,7 @@ public:
 	virtual bool operator==(const GeometricalInfo<num_dimensions, num_dimensions>& gi) const
 	{
 		const VoxelisedGeometricalInfo& vgi = (const VoxelisedGeometricalInfo&)gi;
-		float eps = 1e-4;
+		float eps = 1e-2;
 		return
 			near_(_offset, vgi.get_offset(), eps) &&
 			near_(_spacing, vgi.get_spacing(), eps) &&
--- a/src/common/include/sirf/common/GeometricalInfo.h
+++ b/src/common/include/sirf/common/GeometricalInfo.h
@@ -90,12 +90,13 @@ public:
 	virtual bool operator==(const GeometricalInfo<num_dimensions, num_dimensions>& gi) const
 	{
 		const VoxelisedGeometricalInfo& vgi = (const VoxelisedGeometricalInfo&)gi;
-		float eps = 1e-2;
+		float eps = 0.01;
+		float delta = 0.1;
 		return
 			near_(_offset, vgi.get_offset(), eps) &&
 			near_(_spacing, vgi.get_spacing(), eps) &&
 			_size == vgi.get_size() &&
-			near_(_direction, vgi.get_direction(), eps);
+			near_(_direction, vgi.get_direction(), delta);
 	}
 	virtual bool operator!=(const GeometricalInfo<num_dimensions, num_dimensions>& gi) const
 	{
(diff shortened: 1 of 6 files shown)


src/xGadgetron/pGadgetron/tests/test1.py
```diff
@@ -172,9 +172,14 @@ def test_main(rec=False, verb=False, throw=True):
...
-    for n in (2,2.0,numpy.int32(2),numpy.int64(2),numpy.float128(2),numpy.complex(2)):
+    twos = (2,2.0,numpy.int32(2),numpy.int64(2),numpy.complex(2))
+    try:
+        twos += numpy.float128(2)
+    except:
+        pass
+    for n in twos:
         complex_images.fill(n)
-        test.check_if_zero_within_tolerance((complex_images-2).norm())
+        test.check_if_equal(0, (complex_images-2).norm())
```
KrisThielemans, line 182: why replace with `check_if_equal`? You can never know with numerical calculations.

src/common/include/sirf/common/GeometricalInfo.h
```diff
@@ -90,11 +90,12 @@ class VoxelisedGeometricalInfo :
 	virtual bool operator==(const GeometricalInfo<num_dimensions, num_dimensions>& gi) const
 	{
 		const VoxelisedGeometricalInfo& vgi = (const VoxelisedGeometricalInfo&)gi;
+		float eps = 1e-4;
```
KrisThielemans, line 93:
```suggestion
		float eps = 1e-4;
```
```suggestion
		float eps = 1e-2;
```
Offsets etc are in mm, testing for 0.1 micrometer accuracy is far too much. Even 0.01mm is asking for a lot, but I guess we can see how that goes.

src/xSTIR/cSTIR/tests/test1.cpp
```diff
@@ -101,6 +101,9 @@ int test1()
 		image_data.fill(1.0);
 		float im_norm = image_data.norm();
 		std::cout << "image norm: " << im_norm << '\n';
+		const VoxelisedGeometricalInfo3D &geom_info = *image_data.get_geom_info_sptr();
+		std::cout << geom_info.get_info().c_str() << '\n';
+		std::cout << bool(geom_info == geom_info) << '\n';
```
KrisThielemans, line 106: can we convert this to a test?

src/xSTIR/pSTIR/tests/tests_one.py
```diff
@@ -132,12 +132,8 @@ def test_main(rec=False, verb=False, throw=True):
...
     geom_info = image.get_geometrical_info()
     geom_info.print_info()
-    if geom_info.get_size() != (image_size[0],image_size[1],image_size[2]):
-        #raise AssertionError("SIRF get_geometrical_info().get_size() failed.")
-        warnings.warn("SIRF get_geometrical_info().get_size() failed.")
-    if geom_info.get_spacing() != (voxel_size[0],voxel_size[1],voxel_size[2]):
-        #raise AssertionError("SIRF get_geometrical_info().get_spacing() failed.")
-        warnings.warn("SIRF get_geometrical_info().get_spacing() failed.")
+    test.check_if_equal(geom_info.get_size(), image_size[::-1])
+    test.check_if_equal(geom_info.get_spacing(), voxel_size[::-1])
```
KrisThielemans, line 137: Please revert this. This will make the test work at the moment, but the underlying code is wrong, so we test if it's wrong...

KrisThielemans (changes requested):

src/xGadgetron/pGadgetron/tests/test1.py
evgueni-ovtchinnikov: `complex_images` is filled with 2s, which are represented exactly

src/common/include/sirf/common/GeometricalInfo.h
evgueni-ovtchinnikov: ok

src/xGadgetron/pGadgetron/tests/test1.py
KrisThielemans: fair enough, but suppose someone changes it to some other value (admittedly a bit hard to get a non-exact representation with integers, unless it'd be really big). In any case, it seems best practice to me to check real numbers with tolerance.

src/xSTIR/cSTIR/tests/test1.cpp
evgueni-ovtchinnikov: too trivial as it is now

@johannesmayer can you do something more meaningful with `==`?

src/xSTIR/pSTIR/tests/tests_one.py
evgueni-ovtchinnikov: ok

src/xGadgetron/pGadgetron/tests/test1.py
evgueni-ovtchinnikov: nobody is supposed to change anything in our regression tests without giving a sound reason, and I see no such reason for changing 2 to any other value - the whole point of this test is to ensure `fill` can handle scalars

KrisThielemans: There is no reason to change it from 2, but I see no good reason to have a test that is stronger than what will work for most scalars. I find it misleading. I'd prefer that you revert it.

evgueni-ovtchinnikov: ok

src/xSTIR/cSTIR/tests/test1.cpp
evgueni-ovtchinnikov: test added

evgueni-ovtchinnikov:
@KrisThielemans @johannesmayer ok to merge?

KrisThielemans:
ok. I assume you and GHA have run a recent version (as there have been a lot of [ci skips]...)? If not, maybe just add a line to CHANGES and see what GHA says.

src/common/include/sirf/common/GeometricalInfo.h
```diff
@@ -117,6 +118,34 @@ class VoxelisedGeometricalInfo :
...
+		float s = 0;
+		float t = 0;
+		for (int i = 0; i < num_dimensions; i++) {
+			float xi = x[i];
+			float yi = y[i];
+			s = std::max(s, std::abs(xi));
+			s = std::max(s, std::abs(yi));
+			t = std::max(t, std::abs(xi - yi));
+		}
+		return t <= eps * s;
```
johannesmayer, line 132: This near function just looks at the maximum distance between two `Coordinate` instances and not the element-wise comparison, right?

so if `x = [100.0, 0.1, 3.0]` and `y=[100.0, 1.0, 3.0]` then `s` should be `100` and `t = 0.9` so then if you call it with `eps= 1e-2` the condition reads true because `0.9 < 1e-2 * 100` but the two Coordinates are in fact different... or did I read the code wrong?

same goes for the comparison of direction matrices below

johannesmayer: Shouldn't there be an element-wise float comaprison between two floats a and b as `equal(a,b,eps)= abs(a-b) <= eps*max(a,b)`... then it would read for the above example for the 2nd component that `0.9 < 1e-2 * 1` and that would evaluate to `false` (which is not a proof of course :D )

johannesmayer (changes requested):

src/common/include/sirf/common/GeometricalInfo.h
KrisThielemans: ok. good point! In fact, there is no need to doing a relative comparison here. We want to compare distances in up to .01 mm, and directions vectors up to 0.1.

We can therefore simply use `norm(x-y)<=eps`

evgueni-ovtchinnikov: @johannesmayer but if `x[0] = 0` becomes `y[0] = 1e-7` purely because of the round-off error in calculating `y` then `near_` would return `false`.

@KrisThielemans I went for relative distances because I was not sure about measurement units, switched to absolute now.

src/common/include/sirf/common/GeometricalInfo.h
```diff
@@ -90,12 +90,13 @@ class VoxelisedGeometricalInfo :
 	virtual bool operator==(const GeometricalInfo<num_dimensions, num_dimensions>& gi) const
 	{
 		const VoxelisedGeometricalInfo& vgi = (const VoxelisedGeometricalInfo&)gi;
-		float eps = 1e-2;
+		float eps = 0.01;
```
KrisThielemans, line 93: could get rid of this one now. not used

src/common/include/sirf/common/GeometricalInfo.h
KrisThielemans: sure. looks fine now

KrisThielemans:
funny how everything always escalates 😢

src/common/include/sirf/common/GeometricalInfo.h
evgueni-ovtchinnikov: @KrisThielemans `eps` is used in lines 96 and 97

GHA builds failed because of some data download troubles - should I re-run right away or should I wait a while?

PR #980, opened 2021-07-02
Fix MR Geometry

### LINKED ISSUE #960: GadgetronImagesVector::reorient potentially wrong in special cases
In the reorientation method for the MR image data, there is a loop over all images and the z-position is computed based on the  index of the loop.

The line 
https://github.com/SyneRBI/SIRF/blob/e214a7d4c59630e29f84adcabbc2888ee85b23fb/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp#L1546

in 
```

for (unsigned im=0; im<number(); ++im) {
        // Get image header
        ISMRMRD::ImageHeader &ih = image_wrap(im).head();

...
        for (unsigned i=0; i<3; ++i)
                ih.position[i] = offset[i]
                                    + direction[i][0] * (ih.field_of_view[0] / 2.0f)
                                    + direction[i][1] * (ih.field_of_view[1] / 2.0f)
                                    + direction[i][2] * float(im) * geom_info_out.get_spacing()[2];
```

However, the different images in a GadgetronImagesVector are not necessarily 2D slices, so if I acquire multiple repetitions for example in a cine, or multiple contrasts all these 2D slices should be in the same location but I think this will shift them around in space as well. 

A potential better implementation is to use the header flag for the slice number that could be used to shift them instead image number in the container.

johannesmayer:
Changes include:
*  `GadgetronImagesVector::set_up_geom_info()` now takes slices into account correctly.
* `GadgetronImagesVector::reorient()` takes the FOV of 2D and 3D data into account correctly.
* tests in MR CPP tests if a random reorientation of Shepp-Logan reconstruction is possible.
* `GadgetronImagesVector::sort` is now sorting the images first w.r.t. the projection of the position onto the slice direction and only secondary for contrast and slice.

Fixes #960.


src/xGadgetron/pGadgetron/tests/tests_mrimage_geometry.py
```diff
@@ -0,0 +1,63 @@
+
+# -*- coding: utf-8 -*-
+"""sirf.Gadgetron test.
+v{version}
+
+Constructor of MR image data from MR acquisition data
```
KrisThielemans, line 6: - the name of the file implies it's testing something else. Maybe the file should be renamed?
- add some more info on what this "test" is actually doing. Would be useful to mention it's writing a DICOM file for instance...
- are there any actual tests that could be done? (I guess only appropriate if we know what the input is for sure is going to be)

KrisThielemans (changes requested):
This looks pretty good to me! Thanks.

Could you also
- add something to CHANGES.md
- find/add somewhere in our documentation where we mention something about supported MR acquisitions and what we do with geometry.
- ideally also add some doxygen comments

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
```diff
@@ -1637,6 +1682,14 @@ GadgetronImagesVector::set_up_geom_info()
 
         // Calculate the spacing!
         ISMRMRD::ImageHeader &ih2 = image_wrap(1).head();
+
+        if( spacing[2] != get_slice_spacing(ih1, ih2) )
```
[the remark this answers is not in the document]
KrisThielemans: discussed that the warning is still valid. However need to take tolerance into account

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
```diff
@@ -1637,6 +1682,14 @@ GadgetronImagesVector::set_up_geom_info()
...
+        {
+            std::cout << "\nGadgetronImagesVector::set_up_geom_info(). "
+                        "Warning, you set up geometry for slices whose width is not their distance."
+                        "This setup does probably not account for overlaps or gaps between slices.\n";
+        }
+        
         spacing[2] = get_slice_spacing(ih1, ih2);
```
KrisThielemans, line 1693:
```suggestion
       // just making sure they are the same, as opposed to "up to tolerance"
        spacing[2] = get_slice_spacing(ih1, ih2);
```

src/xGadgetron/pGadgetron/tests/tests_mrimage_geometry.py
KrisThielemans: let's just remove this file

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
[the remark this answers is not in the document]
johannesmayer: done

src/xGadgetron/pGadgetron/tests/tests_mrimage_geometry.py
johannesmayer: done

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
johannesmayer: done

johannesmayer:
Merged `master` and pushed again, and updated `Changes.md `
I checked again, if the last build clears, this PR is good to go as far as I see.

KrisThielemans (approved):
let's hope the tests work!

PR #986, opened 2021-07-23
Replace wherever possible boost types with std ones

### LINKED ISSUE #241: remove boost dependency as much as possible
We're having problems with conflicts with boost versions. This is a serious problem for the boost libraries that are compiled (i.e. not header-only), see e.g. #240, although header-only libraries could still throw up surprising things.

Doing this completely might be hard, but we should at least aim to get rid of explicit includes of boost in our .h files. That way, the boost dependency could be made private. But far better would be to get rid of it completely.

It seems to me that there are at present only few places where we use boost. For instance, in our .h files. 
```
find ../../.. -type f -name \*.h -exec grep -nH -e boost/ {} +
../../../src/xGadgetron/cGadgetron/cgadgetron_shared_ptr.h:24:#include "boost/shared_ptr.hpp"
../../../src/xGadgetron/cGadgetron/gadget_lib.h:35:#include <boost/algorithm/string.hpp>
../../../src/xGadgetron/cGadgetron/gadgetron_client.h:33:#include <boost/asio.hpp>
../../../src/xGadgetron/cGadgetron/gadgetron_client.h:34:#include <boost/shared_ptr.hpp>
../../../src/xGadgetron/cGadgetron/gadgetron_client.h:35:#include <boost/thread/mutex.hpp>
../../../src/xGadgetron/cGadgetron/gadgetron_client.h:36:#include <boost/thread/thread.hpp>
../../../src/xGadgetron/cGadgetron/gadgetron_data_containers.h:33:#include <boost/algorithm/string.hpp>
../../../src/xGadgetron/cGadgetron/gadgetron_x.h:38:#include <boost/thread/mutex.hpp>
../../../src/xGadgetron/cGadgetron/gadgetron_x.h:39:#include <boost/shared_ptr.hpp>
../../../src/xGadgetron/cGadgetron/xgadgetron_utilities.h:36:#include <boost/thread/mutex.hpp>
../../../src/xSTIR/cSTIR/stir_types.h:24:#include <boost/algorithm/string.hpp>
../../../src/Registration/cReg/SIRFReg.h:55:#include <boost/filesystem.hpp>
../../../src/Registration/cReg/SIRFRegMisc.h:33:#include <boost/filesystem.hpp>
../../../src/Registration/cReg/SIRFRegNiftyResample.h:40:#include <boost/filesystem.hpp>
../../../src/Registration/cReg/SIRFRegParser.h:37:#include <boost/filesystem.hpp>
```
Quite a few of these seem to be left-overs from the past. Most can and should be replaced by their `std` versions, in particular uses of `boost/shared_ptr.hpp`. Hopefully `mutex` as well.

SIRFReg depends on `boost/filesystem`. Sadly, `std::filesystem` exists from C++-14 only. I'm a bit reluctant to require C++-14, although Gadgetron does it now anyway. I suggest to do a CMake check on existence of `std::filesystem`, and only use `boost::filesystem` if we need to.

If we cannot remove the dependency completely, a trick to get rid of it in the .h files is to use pre-declaration in the .h, and include it in the .cpp. This works for pointers and references. So, as opposed to have a `boost::filesystem` member, we could have a `std::shared_ptr<boost::filesystem>`. That's a bit ugly though.

evgueni-ovtchinnikov:
Trying to fix #241.


src/common/include/sirf/common/iequals.h
```diff
@@ -0,0 +1,12 @@
+#pragma once
+
+#ifndef CASE_INSENSITIVE_STRING_COMPARISON
+#define CASE_INSENSITIVE_STRING_COMPARISON
+
+#include <string>
+
+namespace sirf {
+	bool iequals(const std::string& a, const std::string& b);
```
KrisThielemans, line 9: great, please add a doxygen line

src/common/iequals.cpp
```diff
@@ -0,0 +1,16 @@
+#include <cctype>
+#include <string>
+
```
KrisThielemans, line 3: best to include the `iequals.h` header, such that we're sure they're consistent.

KrisThielemans (changes requested):
this looks great to me. Possibly the CMakeLists.txt could be adapted to drop some boost dependencies explicitly?

evgueni-ovtchinnikov:
@KrisThielemans re drop boost in CMakeLists.txt: so far as I can see, `boost::algorithm` is not referred there, some other boost libraries are still in use unfortunately

KrisThielemans:
> some other boost libraries are still in use unfortunately

ok.

src/common/include/sirf/common/iequals.h
```diff
@@ -0,0 +1,21 @@
...
+
+#ifndef CASE_INSENSITIVE_STRING_COMPARISON
+#define CASE_INSENSITIVE_STRING_COMPARISON
+
+#include <string>
+
+/*!
+\file
+\ingroup Common
+\brief Case insensitive string comparison, replaces boost::iequals.
```
KrisThielemans, line 11:
```suggestion
\brief Case insensitive string comparison sirf::iequals.
```

src/common/include/sirf/common/iequals.h
```diff
@@ -0,0 +1,21 @@
...
+
+\author Evgueni Ovtchinnikov
+\author SyneRBI
+*/
+
+namespace sirf {
+	bool iequals(const std::string& a, const std::string& b);
```
KrisThielemans, line 18:
```suggestion
       /*!
          \ingroup Common
          \brief Case insensitive string comparison, replaces boost::iequals.
        */
	bool iequals(const std::string& a, const std::string& b);
```

KrisThielemans (changes requested):
sorry. doxygen needs to be before the function (not just the file)

KrisThielemans (approved):

PR #989, opened 2021-08-03
Moved retrieving ISMRMRD::Acquisition parameters to C++

### LINKED ISSUE #981: get_ISMRMRD_info is very slow for 3D data
`get_ISMRMRD_info` loopes over all the Acquisitions (and also Images but there it is less of a problem) in Python. This seems to be very slow. Calling this function on a 3D data set takes a couple of minutes. I guess we need a better C++ implementation of this function.

evgueni-ovtchinnikov:
fixes #981

### DIFF
--- a/examples/Python/MR/acquisition_data.py
+++ b/examples/Python/MR/acquisition_data.py
@@ -94,7 +94,7 @@ def main():
         print('first readout shape: %dx%d' % a0.shape)
         
     # display flags
-    print('Flags'),
+    print('Flags')
     print(flags)
     
     # inspect some kspace_encode_step_1 counters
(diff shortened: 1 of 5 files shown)


KrisThielemans (commented):
certainly looks fine to me but I haven't looked at the detailed code, not tried it :-)

Do we have any tests for this? Not so easy to do I suppose as it would require "real" MR data. Maybe we already download some from zenodo in our tests?

In any case, seems that a manual check would work for now, which i presume you have done already

evgueni-ovtchinnikov:
@KrisThielemans tested on `meas_MID80_2d_cart_cine_30ph_headcoil_FID17006.h5` I got from Christoph some time ago using our demo `acquisition_data.py`.

@ckolbPTB have you tried it yet? Do you need it for images too?

evgueni-ovtchinnikov:
merging, images to be tackled when needed

PR #995, opened 2021-08-12
Replaced where possible returning Succeeded::no with throwing exception

### LINKED ISSUE #994: sirf/STIR set_up(),process() shouldn't return stir::Succeeded in C++
SIRF follows the general strategy of throwing an exception when there's a run-time error, as opposed to hoping that the user will check a return value. This is a bit different from STIR (although we're slowly converting to that strategy as well).

In Python, we already just call set_up() (hopefully checking STIR's return value), but in C++ we still return stir::Succeeded in some places. We should replace those with `void` return, and an check on the STIR return value.

evgueni-ovtchinnikov:
fixes #994


KrisThielemans (approved):

KrisThielemans:
can you add a line in `CHANGES.md` as well please (with `[ci skip]`)

PR #1005, opened 2021-08-27
Fix Call to AffineTransformation Constructor

johannesmayer:
Fixes erroneous use of arguments in constructor of Python `AffineTransformation`.
Currently the user has to make sure that numpy arrays are of type `numpy.float32` for them to be passed to the C++ layer correctly.

- Made sure that the correct arrays are passed in the Python `AffineTransformation` constructor from translation and Euler angles.
- Added method `format_numpy_array_for_setter` to utilities to make sure that the arrays are passed as default `numpy.float32` or another type specified. Currently used in `AffineTransformation` constructor only.


evgueni-ovtchinnikov:
checking data types is also needed in `sirf.Gadgetron.AcquisitionData.set_user_floats` and `sirf.Reg.Quaternion.__init__`.

PR #1008, opened 2021-08-31
Radial 2D MR Reconstruction

johannesmayer:
This PR implements 2D non-cartesian encoding. It is added in the fashion of
This class is also suitable to reconstruct 3D stack-of-stars type data, i.e. cartesian slice encoding with 2D radial readout.

For a user to employ this reconstruction they are only required to call the trajectory setter method `sirf.Gadgetron.set_radial2D_trajectory()` on the `AcquisitionData` object, and the `MRAcquisitionModel` and `CoilSensitivityData` will pick up the rest on their own.

The use is demonstrated in the example script `noncartesian_recon.py`, formerly `rpe_recon.py`.

New features:
- templatisation of the class performing non-cartesian FFT to work in 2D and 3D now.
- rewriting of the `TrajectoryPreparation` class to be able to handle 3D trajectories.
- implementation of linear angle increment 2D radial trajectory preparation.
- implementation of the `NoncartesianFourierEncoding`.


examples/Python/MR/noncartesian_cg_sense.py
```diff
@@ -17,6 +17,10 @@
   -e <engn>, --engine=<engn>  reconstruction engine [default: Gadgetron]
   -n <bool>, --non-cart=<bool> run recon iff non-cartesian code was compiled
                               [default: False]
+  -r <bool>, --recon=<bool>   run recon iff non-cartesian code was compiled
+                              [default: False]
+  --traj=<str>                trajectory type, must match the data supplied in file
+                              [default: grpe]
```
KrisThielemans, line 23: I guess need to list options

examples/Python/MR/noncartesian_cg_sense.py
```diff
@@ -54,104 +58,121 @@
...
 
-    # sort processed acquisition data;
-    print('---\n sorting acquisition data...')
-    processed_data.sort()
-    
-    #set the trajectory and compute the dcf
-    print('---\n setting the trajectory...')
-    processed_data = set_grpe_trajectory(processed_data)
+    print('---\n Backward projection ...')
+    # this is our first residual
```
KrisThielemans, line 86: comment out of place?

examples/Python/MR/noncartesian_cg_sense.py
```diff
@@ -54,104 +58,121 @@
...
 
 import numpy
 
 # define symmetrical operator for cg-optimisation
-def EhE(E, image ):
+def EhE(E, image):
     return E.backward( E.forward(image) )
 
-def main():
+def SENSE(rawdata, num_iter = 10, stop_criterion = 1e-7):
```
KrisThielemans, line 70: is it correct to call this `SENSE`? I thought that was an image-based technique

examples/Python/MR/noncartesian_recon.py
```diff
@@ -16,8 +16,10 @@
                               subfolder of SIRF root folder
   -o <file>, --output=<file>  output file for simulated data
   -e <engn>, --engine=<engn>  reconstruction engine [default: Gadgetron]
-  -n <bool>, --non-cart=<bool> run recon iff non-cartesian code was compiled
+  -r <bool>, --recon=<bool>   run recon iff non-cartesian code was compiled
                               [default: False]
+  --traj=<str>                trajectory type, must match the data supplied in file
```
KrisThielemans, line 21: list options

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/NonCartesianEncoding.h
```diff
@@ -75,66 +75,107 @@ IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
+    {
+        if( img_output_dims.size() != D)
+            throw LocalisedException("The image dimensions of the output should be of the dimensions of the Gridder." , __FILE__, __LINE__);
+
+        traj.get_dimensions(this->trajdims_);
+        this->output_dims_ = img_output_dims;
+        this->nufft_operator_.preprocess(traj);
+    }
+
+    void ifft(CFGThoNDArr& img, const CFGThoNDArr& kdata)
```
KrisThielemans, line 109: `const` member?

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/NonCartesianEncoding.h
```diff
@@ -75,66 +75,107 @@ IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
...
+
+        sptr_const_dcw ->fill(normed_dcw_value);
+
+        img.create(this->output_dims_);
+        img.fill(std::complex<float>(0.f, 0.f));
+
+        this->nufft_operator_.compute(kdata, img, sptr_const_dcw.get(), Gadgetron::NFFT_comp_mode::BACKWARDS_NC2C);
+    }
+
+    void fft(CFGThoNDArr& kdata, const CFGThoNDArr& img)
```
KrisThielemans, line 122: `const` member?

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/TrajectoryPreparation.h
```diff
@@ -49,26 +49,108 @@ namespace sirf{
 * reconstruction. The ISMRMRD format has a 3D trajectory data field in their ISRMRMRD::Acquisition classe.
 * The interface provides set_trajectory() which populates this data field depending on the implementation.
 */
+template <uint16_t D>
```
KrisThielemans, line 52: This should be `std::uint16_t` (also in other places). any specific reason we don't just use `unsigned`?

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/TrajectoryPreparation.h
```diff
@@ -49,26 +49,108 @@ namespace sirf{
...
-    void update_acquisitions_info(sirf::MRAcquisitionData& mr_acq);
 
+protected:
     ISMRMRD::Encoding kspace_encoding_;
     ISMRMRD::TrajectoryType traj_type_;
 
-    uint16_t traj_dim_;
-    virtual void set_acquisition_trajectory(ISMRMRD::Acquisition& acq) const =0;
-    virtual std::vector<float> calculate_trajectory(ISMRMRD::Acquisition& acq) const =0;
+    virtual void set_acquisition_trajectory(ISMRMRD::Acquisition& acq) const
```
KrisThielemans, line 122: could use a doxygen comment

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/TrajectoryPreparation.h
```diff
@@ -49,26 +49,108 @@ namespace sirf{
...
+    {
+        acq.resize(acq.number_of_samples(),acq.active_channels(), D);
+        TrajPointSet acq_traj = this->calculate_trajectory(acq);
+        acq.setTraj(&(acq_traj[0][0]));
+    }
+    virtual void update_acquisitions_info(sirf::MRAcquisitionData& mr_acq)
```
KrisThielemans, line 128: doxygen here as well (what's the difference with the `set_acquisition_trajectory` (aside from the input type)? Should they be called similarly? (why `set` and `update`?)

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/TrajectoryPreparation.h
```diff
@@ -49,26 +49,108 @@ namespace sirf{
...
+    }
+
+    virtual TrajPointSet calculate_trajectory(ISMRMRD::Acquisition& acq) const =0;
+    virtual void append_to_trajectory(TrajPointSet& tps, ISMRMRD::Acquisition& acq)=0;
+    virtual void get_ky_sorted_trajectory(TrajPointSet& tps, ISMRMRD::Acquisition& acq)=0;
+
 };
 
+typedef TrajectoryPreparation<2> TrajPrep2D;
+typedef TrajectoryPreparation<3> TrajPrep3D;
```
KrisThielemans, line 152: I'd rather call them `TrajectoryPreparation2D` etc. It otherwise becomes hard to remember when you used the abbreviation and when not.

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -1690,3 +1714,36 @@ def calc_rpe_dcw(ad):
     dcw.fill(density_weight)
     
     return dcw
+
+
+
+def calc_radial_dcw(ad):
+    '''
+    Function that computes the kspace weight depending on the distance to the center
+    as in a filtered back-projection. Stricly valid only for equally angular-spaced
```
KrisThielemans, line 1723:
```suggestion
    as in a filtered back-projection. Strictly valid only for equally angular-spaced
```

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -1690,3 +1714,36 @@ def calc_rpe_dcw(ad):
     dcw.fill(density_weight)
     
     return dcw
+
+
+
+def calc_radial_dcw(ad):
```
KrisThielemans, line 1720: any reason we don't have this one in C++?

KrisThielemans (changes requested):
looks good overall (as usual, I didn't check the actual implementation)

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/NonCartesianEncoding.h
johannesmayer: can't be `const`, the `Gadgetron::hoNFFT_plan::compute` method is not `const`.

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/TrajectoryPreparation.h
johannesmayer: Renamed it

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/NonCartesianEncoding.h
KrisThielemans: I haven't checked the code but potentially this needs an issue in Gadgetron then. they don't have a `set_up` and hten `do_it` type of functionality?

For us, it seems solvable by making `nufft_operator_` a `mutable` member.

johannesmayer: They do. 
So I call the `set_up` method in the constructor of the `Gridder` and then the `compute` method in our `(i)fft` method.
I can take a look at mutable.

johannesmayer:
Hey @evgueni-ovtchinnikov, I can't get this to run because of this XML header version error. (https://github.com/SyneRBI/SIRF/runs/3953431886#step:7:5639).

However, I cannot replicate this locally, and all the tests seem to run first without a problem and then this exception is caught. 
How do I find out where it is from? Could you please give me a hand?

johannesmayer:
I am confused because in the radial test [I read in the same test file](https://github.com/johannesmayer/SIRF/blob/2405929d80eaa787bb8a1538dce8c2f3f7a4c495/src/xGadgetron/cGadgetron/tests/mrtests.cpp#L642) as we do in the same MR C++ tests [just a couple of lines above.](https://github.com/johannesmayer/SIRF/blob/2405929d80eaa787bb8a1538dce8c2f3f7a4c495/src/xGadgetron/cGadgetron/tests/mrtests.cpp#L599) Yet somehow then this XML version header is thrown.

KrisThielemans:
ah. red-herring... I believe this is because the Cartesian only *read* to ISMRMRD file, while the Cartesian ones also *write*
https://github.com/SyneRBI/SIRF/blob/9ab43cc091711254f1131070e98ba3883564f91d/src/xGadgetron/cGadgetron/tests/mrtests.cpp#L668
@evgueni-ovtchinnikov had spotted that the ISMRMRD error is generated in `serialize` (i.e. when writing), but the penny didn't drop for me.

So i think this will be resolved once we upgrade ISMRMRD and generate the file.

Of course, I think it'd be best not to modify an existing file, so the test should make a copy first (either within C++ (not so easy I guess), or in the `CMakeLists.txt`

johannesmayer:
Okay, good thank you for discovering that.
Then I will wait for the upgrade

KrisThielemans:
all done? Want to merge `master` on here yet again first? (would reduce the diff that we see now)

johannesmayer:
I'm happy with it

KrisThielemans:
celebration time!

PR #1009, opened 2021-09-09
In place divide

epapoutsellis:
removes python2 divide

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ## v3.x.x
 
 * Removed `__div__` ,  `__idiv__` operators for `DataContainers` required for Python2.
+* Added `__truediv__` and `__itruediv__` Python3 operators to DataContainer algebra.
 
 ## v3.1.1
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 # ChangeLog
 
+## v3.x.x
+
+* Removed `__div__` ,  `__idiv__` operators for `DataContainers` required for Python2.
+
 ## v3.1.1
 
 * Build system
--- a/src/common/SIRF.py
+++ b/src/common/SIRF.py
@@ -620,6 +620,13 @@ class DataContainer(ABC):
         Same as __div__ but for Python 3.*
         '''
         return self.__div__(other)
+
+    def __itruediv__(self, other):
+        '''
+        Same as __idiv__ but for Python 3.*
+        '''
+        return self.__idiv__(other)    
+            
     @property
     def shape(self):
         '''Returns the shape of the data array
--- a/src/common/SIRF.py
+++ b/src/common/SIRF.py
@@ -399,10 +399,10 @@ class DataContainer(ABC):
 
         return NotImplemented
 
-    def __div__(self, other):
+    def __truediv__(self, other):
         '''
         Overloads / for data containers division by a scalar or (elementwise)
-        another data container (Python 2.*)
+        another data container (Python 3.*)
 
         Returns the ratio self/other if other is a scalar
         or the elementwise ratio if other is of the same type as self.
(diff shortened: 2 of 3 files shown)


KrisThielemans (commented):
I cannot get my head round Python div functions. If you say this is fine, ok for me. Just wondering if we really need to break Python2 compatibility (i.e. does it cost us anything?)

paskino (changes requested):
I think we need to add unittests for in-place division (but more generally all algebra) here
https://github.com/SyneRBI/SIRF/blob/e04334905ac3d9e6a136970b6f1ff5028b7328cd/src/common/Utilities.py#L565-L605

Notice that `divide` is tested in-place and not. But what's not tested is the form 

```python

a /= b
```
which is what was failing!

KrisThielemans:
> we do not have to break Python2 compatibility, but keeping it lowers readability and reduces maintainability.

ok. This needs to be very clear in the `CHANGES.md` then

paskino:
Yes, this is already part of the Changelog from v3.0.0: Python 2 is no longer supported.

https://github.com/SyneRBI/SIRF/blob/master/CHANGES.md#backwards-incompatible-changes

KrisThielemans:
the general statement, yes. I believe we need a specific line about the current change.

CHANGES.md
```diff
@@ -1,5 +1,9 @@
 # ChangeLog
 
+## v3.x.x
+
+* Removed `__div__` ,  `__idiv__` operators for `DataContainers` required for Python2.
```
KrisThielemans, line 5: and something about the fix for python3

KrisThielemans (changes requested):
@paskino, feel free to merge once done

KrisThielemans:
great. thanks. @epapoutsellis in future, please use `[ci skip]` in the body of the commit message if appropriate (see our [contribution guide](https://github.com/SyneRBI/SIRF/blob/master/CONTRIBUTING.md)

ashgillman:
Wow, I thought for sure this had something to do with the fact that I'd changed from CIL images to SIRF images, but despite days of debugging I hadn't found this. Great work @paskino and @epapoutsellis and whoever else helped!

It is weird that both in-place and return versions of `L2NormSquared.proximal()` use `\=`, yet I was getting different results, https://github.com/vais-ral/CCPi-Framework/blob/master/Wrappers/Python/cil/optimisation/functions/L2NormSquared.py#L123-L151

paskino:
@gfardell and @epapoutsellis did the job. 

I do not quite understand the [in-place operators](https://docs.python.org/3/library/operator.html#in-place-operators) explanation in the python library, especially in view of the following behaviour (Python 3.8.5):

Python does not (seem to) require to have an `__itruediv__` to be able to call it!!! Rather it just needs you to have defined `__truediv__` 

If the `__itruediv__` is not present Python will run `__truediv__` and reassign the output to the original variable name. So you think you've done an in-place operation while you have not. I think this is a Python bug???

Our bug was that this happened in a method where the new variable got assigned to the old variable (name), but it went out of scope at the end of the method, thereby invalidating the result! 

Try this code:

```python
class A(object):
    def __init__(self, a):
        self.list = a[:]
class B(A):
    def __truediv__(self, other):
        '''divides 2 lists element by element'''
        return B([el1/el2 for el1,el2 in zip(self.list, other.list)])

def method(a,b):
    a/=b

if __name__ == '__main__':
    a0 = B([2,2])
    a1 = B([1,1])
    b = a1/a0
    print (b.list)
    print (id(a1), a1.list)

    a1 /= a0

    print (id(a1), a1.list)
    
    a1bis = B([1,1])
    method(a1bis,a0)
    print (id(a1bis), a1bis.list)
```

this returns

```
[0.5, 0.5]
1998060834976 [1, 1]
1998063103568 [0.5, 0.5]
1998060834976 [1, 1]
```

paskino:
> Wow, I thought for sure this had something to do with the fact that I'd changed from CIL images to SIRF images

In a way that's true as in CIL we had a correct implementation of `__itruediv__` which was missing in SIRF.

ashgillman:
Oh wow. So the logic in Python must be like, if `__itruediv__` doesnt exist, `a/=b -> a = a/b`. I've looked through the docs also, and this is kind-of implied, but never explicit.

So after `a/=b`, a points to the new result, but the original a was never updated.

PR #1018, opened 2021-10-28
Implemented conjugate methods in C++ and Python

### LINKED ISSUE #1017: Add conjugate method for SIRF
There is a `conjugate` method for MR objects in [Gadgetron.py](https://github.com/SyneRBI/SIRF/blob/f8b29afdcd8a64e0cabe2d5b0e5368139ad29df2/src/xGadgetron/pGadgetron/Gadgetron.py#L446). It would be nice to add the same method in the `SIRF.py` which basically will be the same method as `.conjugate()` does nothing on real data.

evgueni-ovtchinnikov:
fixes #1017


ClaireDelplancke:
`ImageData` object has no attribute `conjugate`. This causes an error in CIL's `TotalVariation`, see https://github.com/TomographicImaging/CIL/pull/1028.

(I installed SIRF and CIL via Superbuild 2 weeks ago)

KrisThielemans:
I guess so, you can check by doing
```
cd build/sources/SIRF
git log
```
and see if these commits are in there.

PR #1021, opened 2021-11-15
Implemented workaround enabling new Gadgetron to read old ISMRMRD acquisition data

evgueni-ovtchinnikov:
Recent versions of Gadgetron fail to read ISMRMRD acquisition data generated by old ISMRMRD because of the line
```
<version>3</version>
```
in the acquisition data header XML.

This workaround simply makes method `MRAcquisitionData::read` ignore this line.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,10 @@
 
 ## v3.x.x
 
+* MR Geometry
+  - fixed GadgetronImagesVector::reorient() to only consider slice index 
+  and ignore dimensions such as contrast, repetition etc.
+
 * PET/STIR
   - (C++) Replaced where possible returning `stir::Succeeded::no` with throwing exception.
   - (C++) Fixed a bug in `PETAcquisitionDataInMemory::norm`.
--- a/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
+++ b/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
@@ -1036,15 +1036,20 @@ GadgetronImagesVector::sort()
 	for (int i = 0; i < ni; i++) {
       ImageWrap& iw = image_wrap(i);
       ISMRMRD::ImageHeader& head = iw.head();
-		t[0] = head.contrast;
-        t[1] = head.repetition;
-        // Calculate the projection of the position in the slice direction
-        t[2] = -( head.position[0] * head.slice_dir[0] +
+		// It is crucial to sort the images first by their projection onto the slice direction
+        // and only then consider other dimensions.
+        // In the function this->reorient() the computation of the position of the 2D slices
+        // requires them to be ordered with respect to their projection onto the slice direction to
+        // ensure that they are located next to each other.
+        t[0] = -( head.position[0] * head.slice_dir[0] +
                 head.position[1] * head.slice_dir[1]   +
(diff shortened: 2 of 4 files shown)


johannesmayer:
The code seems to just cut out the `<version> ... <\version>` part, and I think that is correct.
What I find strange is that they check the version, but if no version is present, however, they don't complain, only if the wrong one is present.

DANAJK:
Sorry, I don't know. There are some hints here as to what has changed:
https://github.com/ismrmrd/ismrmrd/releases

evgueni-ovtchinnikov:
@KrisThielemans actually, the very idea of my workaround came to me when I noticed that David's acquisition data files `grappa2_*rep.h5` had headers without that troublesome version line

KrisThielemans:
ok, well, what about checking the version number, if it's the same as the current ISMRMRD version, we keep it. If it's higher, we throw an error, if it's lower, we remove it and write a (very noticeable) warning. This would have to be documented in the User's Guide (and CHANGES)

KrisThielemans (changes requested):
ok, but I think the warning/error should be a bit more helpful really, like "ISMRMRD software version 4 is older than the version in the header (5). Failed to read file XXXX"

DANAJK (commented):
I spent a while looking at this code and the title of this Issue, but it isn't clear to me which way around the error is thrown. If one of these things called 'header' refers to the .h5 file in ISMRMRD format, can that be more explicitly stated? If the other 'header' refers to the running software, something like "Input ISMRMRD file has version XX but software only able to read versions less than YY"

evgueni-ovtchinnikov:
@DANAJK @KrisThielemans How about "Input ISMRMRD acquisitions file header is written in XML version XX, but the version of ISMRMRD used presently by SIRF supports XML version YY or less only"?

DANAJK:
> @DANAJK @KrisThielemans How about "Input ISMRMRD acquisitions file header is written in XML version XX, but the version of ISMRMRD used presently by SIRF supports XML version YY or less only"?

Yes, that is clearer.

KrisThielemans:
> > @DANAJK @KrisThielemans How about "Input ISMRMRD acquisitions file header is written in XML version XX, but the version of ISMRMRD used presently by SIRF supports XML version YY or less only"?
> 
> Yes, that is clearer.

Some suggested change:
"Input acquisition file was written in with ISMRMRD XML version XX, but the version of ISMRMRD used presently by SIRF supports XML version YY or less only"

evgueni-ovtchinnikov:
@johannesmayer how come my PR got lots of commits from your PR #980?

KrisThielemans:
a bit weird. However, I don't think @johannesmayer has push access, so maybe you merged his version into yours at some point by mistake?

We can get rid of it by using interactive rebase (a bit dangerous), or merge #980 first.

evgueni-ovtchinnikov:
@KrisThielemans I remembered: 13 days ago I merged #980 onto `master`, which created conflicts between `master` and `qfix-xml`. I merged `master` onto `qfix-xml` and resolved conflicts. Safe to merge this PR then?

KrisThielemans:
ok. I guess I'd recommend to do another merge of master here (as I see no merge commit for `master`) to be safe. of course, if you see locally that this didn't do anything, you can go ahead and merge this PR.

PR #1027, opened 2021-12-14
Trying to accelerate PET algebra tests for larger data

### LINKED ISSUE #1013: test PET__PYTHON_ALGEBRA is very slow
It takes 120s on my machine. Would be worth finding out which bit is so slow (or is it everything?)

evgueni-ovtchinnikov:
- trying to fix #1013 for larger data

### DIFF
--- a/src/xSTIR/pSTIR/tests/test_algebra.py
+++ b/src/xSTIR/pSTIR/tests/test_algebra.py
@@ -41,16 +41,30 @@ class TestSTIRImageDataAlgebra(unittest.TestCase, TestDataContainerAlgebra):
 class TestSTIRAcquisitionDataAlgebraFile(unittest.TestCase, TestDataContainerAlgebra):
 
     def setUp(self):
-        path = os.path.join(
-            examples_data_path('PET'), 'thorax_single_slice', 'template_sinogram.hs')
+        self.set_storage_scheme()
+        sirf_path = os.environ.get('SIRF_PATH')
+        data_path = os.path.join(sirf_path, 'tests', 'data.hs')
+        if os.path.exists(data_path):
+            print('reading data from %s' % data_path)
+            data = pet.AcquisitionData(data_path)
+            self.image1 = data.get_uniform_copy(0)
+            self.image2 = data.get_uniform_copy(0)
+            self.set_storage_scheme()
+            return
+        path = os.path.join(examples_data_path('PET'),
+                            'mMR', 'mMR_template_span11_small.hs')
+#                            'thorax_single_slice', 'template_sinogram.hs')
         if os.path.exists(path):
             template = pet.AcquisitionData(path)
-            self.image1 = template.get_uniform_copy(0)
-            self.image2 = template.get_uniform_copy(0)
-            # assert False
-            self.set_storage_scheme()
-        
-        
+            print(template.dimensions())
+            data = template.get_uniform_copy(0)
+            data = data.rebin(3, num_views_to_combine=6, num_tang_poss_to_trim=200)
+            print('rebinned to ', data.dimensions())
+            self.image1 = data
+            self.image2 = data.get_uniform_copy(0)
+            print('saving data to %s' % data_path)
+            self.image1.write(data_path)
+
     def tearDown(self):
         pass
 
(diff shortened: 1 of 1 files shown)


src/xSTIR/pSTIR/tests/test_algebra.py
```diff
@@ -63,16 +77,32 @@ def test_division_by_datacontainer_zero(self):
...
-            examples_data_path('PET'), 'thorax_single_slice', 'template_sinogram.hs')
+        sirf_path = os.environ.get('SIRF_PATH')
+        data_path = os.path.join(sirf_path, 'tests', 'data.hs')
+        if os.path.exists(data_path):
+            print('reading data from %s' % data_path)
+            data = pet.AcquisitionData(data_path)
+            self.image1 = data.get_uniform_copy(0)
+            self.image2 = data.get_uniform_copy(0)
+            return
+        path = os.path.join(examples_data_path('PET'),
```
KrisThielemans, line 91: not used?

KrisThielemans (commented):
looks good. I didn't check current run-time.

Where does tests/data.hs sit? Is it checked in ? (I didn't check)

paskino:
I think `setUp` should just rebin the data as you propose:

```
        self.set_storage_scheme()
        sirf_path = os.environ.get('SIRF_PATH')
        
        path = os.path.join(examples_data_path('PET'),
                            'mMR', 'mMR_template_span11_small.hs')
        
        template = pet.AcquisitionData(path)
        # print(template.dimensions())
        data = template.get_uniform_copy(0)
        data = data.rebin(3, num_views_to_combine=6, num_tang_poss_to_trim=200)
        # print('rebinned to ', data.dimensions())
        self.image1 = data
        self.image2 = data.get_uniform_copy(0)
```

Also the file `data.hs` that you create the first time you run the unit test (actually `setUp`) should be deleted at the end of the test, during [`tearDown`](https://github.com/SyneRBI/SIRF/blob/0565a5085983228d7e30f39718ff267983632792/src/xSTIR/pSTIR/tests/test_algebra.py#L69), making the trick not really viable.

paskino (changes requested):
See my comment https://github.com/SyneRBI/SIRF/pull/1027#issuecomment-994491179

src/xSTIR/pSTIR/tests/test_algebra.py
evgueni-ovtchinnikov: > looks good. I didn't check current run-time.
> 
> Where does tests/data.hs sit? Is it checked in ? (I didn't check)

testing time is now 14 sec

`data.hs` and `data.s` are created, if not present, in `SIRF/tests` by the first test - do not know how to delete them after all tests finished (CMake can, I presume?)

> not used?

sorry, do not get what you refer to

evgueni-ovtchinnikov:
> I think setUp should just rebin the data

rebinning in each test takes twice as long (30 sec vs 14 sec)

admittedly my fix is quite ugly, but I do not understand how unittest works, so cannot suggest anything else at the moment

paskino:
Each unit tests will be preceded by a call to `setUp` and call to `tearDown`. The idea here is to create whatever temporary thing required for the test and clear it away even if the test fails.

Your fix is functional, but does go against the whole idea of `setUp` and `tearDown`. Why not add small PET `AcquisitionData` and `ImageData` files in the repository?

paskino:
https://github.com/SyneRBI/SIRF/blob/0565a5085983228d7e30f39718ff267983632792/src/xSTIR/pSTIR/tests/test_algebra.py#L46

The file created during the first unittest will end up in `${SIRF_PATH}/tests`, which is in the[ source directory](https://github.com/SyneRBI/SIRF-SuperBuild/blob/c53f708390da2f055f8552c180fb968703f7f786/env_sirf.sh.in#L6-L7). Does this mean that git will not like it when running a second time?

If so, I believe we should delete the `data.hs` file once done with the tests. I'm not so sure if there is a single test class created to keep the count of how many tests have been run.

A possible solution is to be pragmatic and think that this file will be used by the PET Python algebra only and we can create it in the build directory so that we reduce the pollution of the source and install directories to a minimum. I'd be in favour of this solution.

paskino:
I suggest to close this as issue has been fixed by using a smaller template for the test data.

KrisThielemans:
fine for me

paskino:
Closing as completed by https://github.com/SyneRBI/SIRF/commit/bb1c4374cd96df5a62e7b1304b16ae936f4295ae

PR #1028, opened 2021-12-15
Subset object in SIRF

evgueni-ovtchinnikov:
Uses `subset_object` from `ashgillman/STIR`


evgueni-ovtchinnikov:
To quick test `get_subset()`, I tried it on the subset of all views, i.e. with `range(nv)` as the argument, where `nv` is the number of views (I used `PET\acquisition_data.py` in SIRF branch `subset_object` modified accordingly).

This is what I got for `my_forward_projection.hs` (the number of sinograms shown reduced to 15 for convenience):
Figure_1
Figure_2
Clearly, something went wrong.

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1053,6 +1053,16 @@ def get_info(self):
         pyiutil.deleteDataHandle(handle)
         return info
 
+    def get_subset(self, views):
+        """Returns the subset of self data formed by specified views
+
+        views: numpy.int32 array of views
```
KrisThielemans, line 1059: would need to check this if it is required to have `int32`

src/xSTIR/cSTIR/include/sirf/STIR/stir_data_containers.h
```diff
@@ -528,6 +536,10 @@ namespace sirf {
 			ptr->fill(0.0f);
 			_data.reset(ptr);
 		}
+		PETAcquisitionDataInMemory(std::unique_ptr<stir::ProjData> uptr_pd)
+		{
+			_data = std::move(uptr_pd);
```
KrisThielemans, line 541: this makes little sense to me. it wouldn't necessarily be in memory, would it?

src/xSTIR/cSTIR/include/sirf/STIR/stir_data_containers.h
```diff
@@ -485,6 +487,12 @@ namespace sirf {
 				(_template->same_acquisition_data(this->get_exam_info_sptr(),
 				this->get_proj_data_info_sptr()->create_shared_clone()));
 		}
+		virtual std::unique_ptr<PETAcquisitionData> get_subset(const std::vector<int>& views) const;
+//		{
+//			auto uptr_sub = _data->get_subset(views);
+//			auto ptr_ad = new PETAcquisitionDataInMemory(uptr_sub);
+//			return std::unique_ptr<PETAcquisitionData>(ptr_ad);
+//		}
```
KrisThielemans, line 495: let's remove this

src/xSTIR/cSTIR/include/sirf/STIR/stir_data_containers.h
```diff
@@ -160,6 +160,8 @@ namespace sirf {
 			return false;
 		}
 
+		virtual std::unique_ptr<PETAcquisitionData> get_subset(const std::vector<int>& views) const = 0;
```
KrisThielemans, line 163: not sure why a pure virtual here. Is it not going to be the same for both?

KrisThielemans (changes requested):
https://github.com/UCL/STIR/pull/969 has now been merged. Can you check if things are working here now?

src/xSTIR/cSTIR/include/sirf/STIR/stir_data_containers.h
evgueni-ovtchinnikov: So far as I understand, `ProjData::get_subset` returns `unique_ptr` to `ProjDataInMemory`, does it not?

src/xSTIR/cSTIR/include/sirf/STIR/stir_data_containers.h
evgueni-ovtchinnikov: ok

src/xSTIR/cSTIR/include/sirf/STIR/stir_data_containers.h
KrisThielemans: it does, but this is a general constructor, where you don't know how it will be called. I guess you'd have to do a check if `*uptr_pd` is a `ProjDataInMemory`, if so, `std::move` would be ok, otherwise it'd need to make a copy.

evgueni-ovtchinnikov: @KrisThielemans  Unfortunately, all my attempts to implement your last suggestion failed. My guess is that something is not quite right with `ProjData::get_subset` (see the strange subset sinograms above).

src/xSTIR/cSTIR/include/sirf/STIR/stir_data_containers.h
```diff
@@ -438,6 +440,22 @@ namespace sirf {
 			ptr->fill(0.0f);
 			_data.reset(ptr);
 		}
+		PETAcquisitionDataInFile(std::unique_ptr<stir::ProjData> uptr_pd) : _owns_file(true)
```
KrisThielemans, line 443: Isn't this a slightly weird constructor? If it's passed a `ProjDataInterfile` it will use it (and therefore if it's writable will modify), otherwise it will make a copy of the data with a temporary filename. Any modifications won't go to the argument it was constructed with.

I think this is surprising for the user.

src/xSTIR/cSTIR/include/sirf/STIR/stir_data_containers.h
```diff
@@ -528,6 +547,21 @@ namespace sirf {
 			ptr->fill(0.0f);
 			_data.reset(ptr);
 		}
+		PETAcquisitionDataInMemory(std::unique_ptr<stir::ProjData> uptr_pd)
```
KrisThielemans, line 550: same comment as for the `File` version

KrisThielemans:
it looks like in STIR we're not testing `ProjDataInMemory(ProjData&)` with subsets

src/xSTIR/pSTIR/STIR.py
KrisThielemans: done now

KrisThielemans:
The problem was the type of the `views` argument. `numpy.arange` created an `int64` numpy array, while the C code expected `int32`. This seems to work fine now, although we will of course need to add tests on projection etc.

@evgueni-ovtchinnikov how do we know that in C++ an `int` corresponds to a Python `int32`? This is going to break at some point. We might need to use `int32_t` in C++ as well. Does this occur in other places?

evgueni-ovtchinnikov:
> how do we know that in C++ an int corresponds to a Python int32?

I added function `cpp_int_bytes()` to `Utilities.py`, which can be used to create `ndarray`s of correct `dtype`.

> Does this occur in other places?

We mostly use integer arrays for retrieving dimensions (7 out of 11 C interface helper functions), so it should not take much work to safeguard against different `int` sizes on different architectures.

KrisThielemans:
Best to finish #1088 first, merge it to master and then merge master here. you'll get some conflicts no doubt.

examples/Python/PET/acquisition_data.py
```diff
@@ -86,13 +86,14 @@ def main():
...
     acq_array = acq_data.as_array()
 
-    nv = dim[2]//2
-    views = numpy.arange(nv)
-    acq_subset = acq_data.get_subset(views)
-    dim_subset = acq_subset.dimensions()
-    print('subset dimensions: %d x %d x %d x %d' % dim_subset)
-    if show_plot:
-        acq_subset.show(range(dim[1]//4), title='Sinograms of a subset of views')
+    if storage[0] == 'm':
```
ashgillman, line 89: Maybe not obvious what this means for a casual reader looking for examples. Would it be better to try/catch and print an explanation?

```suggestion
    if storage[0] == 'm':  # for now, we can only subset acquisition data stored in memory
```

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1095,6 +1096,20 @@ def get_info(self):
         pyiutil.deleteDataHandle(handle)
         return info
 
+    def get_subset(self, views):
+        """Returns the subset of self data formed by specified views
+
+        views: array of views (will be converted to numpy ndarray)
+        """
+        '''Ensure the array passed to C++ is a contiguous array of C++ int's
```
ashgillman, line 1104: Should this be a comment not a triple-quote?

ashgillman (commented):
I have only a few minor comments on the code. Also to flag that the feature hasn't been documented in the UserGuide.

KrisThielemans:
@epapoutsellis is there anything that needs to be done here?

epapoutsellis:
@KrisThielemans I don't think so, I  have tested the functionality with CIL. Everything works fine.

src/xSTIR/pSTIR/STIR.py
```diff
@@ -1095,6 +1096,20 @@ def get_info(self):
...
+        '''
```
KrisThielemans, line 1105:
```suggestion
        # Ensure the array passed to C++ is a contiguous array of C++ int's
```

PR #1045, opened 2022-02-03
update message text and set to FATAL_ERROR

### LINKED ISSUE #1042: [CMake] During RUN_ISMRMRD_SHEPP_LOGAN missing FATAL_ERROR leads to build failure
The error at this line

https://github.com/SyneRBI/SIRF/blob/e844ae8b18b4b0ad377c4da11af1ea17afc8d360/src/xGadgetron/cGadgetron/tests/CMakeLists.txt#L61

But `ERROR` type in CMake message does not exist. Use `FATAL_ERROR` with the message:

```We could not find the ismrmrd_generate_cartesian_shepp_logan executable. Either set the variable to the -DEXE_ISMRMRD_SL=<path_to>ismrmrd_generate_cartesian_shepp_logan executable or disable the generation with -DRUN_ISMRMRD_SHEPP_LOGAN=OFF```
.

### PR DESCRIPTION
**Pull request by paskino**
closes #1042

### CODE REVIEW

**Comment by evgueni-ovtchinnikov** [comment]
We give the user an option to disable Shepp-Logan data re-generation, but if they choose to disable it, demos and tests will not work because they will try to open files `autogenerated_*.h5`.
 I believe we should drop `autogenerated_` prefix and, if the re-generation is not chosen, copy data from `data` subfolder to `INSTALL`.

**Comment by KrisThielemans** [comment]
could do, but then we need to update the files in `data`

**Comment by paskino** [comment]
If the user sets `RUN_ISMRMRD_SHEPP_LOGAN` to `OFF` the files will not be generated and the `mrtests.cpp` executable will run against the old files, not the `autogenerated_` ones, see line 111 below.

https://github.com/SyneRBI/SIRF/blob/f4b014db675a02b41b9c44a9381280d5c6c238f0/src/xGadgetron/cGadgetron/tests/CMakeLists.txt#L105-L121

**Comment by paskino** [comment]
The problem you faced @evgueni-ovtchinnikov is that in your case CMake didn't error and you didn't see that the ismrmrd executable to create the phantom had not been found. 

Instead CMake just sent you a `message` which you didn't see and, because `RUN_ISMRMRD_SHEPP_LOGAN` was set to `ON` CMake, continued it's program and set the test files to the `autogenerated_`. Or possibly it failed when trying to generated the files.

**Comment by evgueni-ovtchinnikov** [comment]
@paskino I was referring to Python and Matlab tests, which are hardwired to `autogenerated_*.h5`.

**Comment by evgueni-ovtchinnikov** [comment]
> could do, but then we need to update the files in data

@KrisThielemans My understanding is: if they do not want to re-generate, then they should be happy with whatever we have in `data`.

**Comment by KrisThielemans** [comment]
as long as what we have in `data` works... (and it didn't)

**Comment by evgueni-ovtchinnikov** [comment]
> as long as what we have in data works... (and it didn't)

not after my PR #1021

**Comment by KrisThielemans** [comment]
ok. in that case, I think your suggestion makes sense, but we'll need to add a GHA workflow that tests that functionality (should be easy: just set the extra CMake flag)

**Comment by evgueni-ovtchinnikov** [comment]
My suggestion was not to use `autogenerated_` prefix at all. 

If the user asks (via `cmake`) to re-generate, we run `ismrmrd_generate_cartesian_shepp_logan` and install generated data files, otherwise we install those from `data` subfolder.

As the two cases are mutually exclusive, there is no need to change names.

**Review by evgueni-ovtchinnikov — APPROVED** [review]

**Comment by johannesmayer** [comment]
This will currently fail the tests if `RUN_ISMRMRD_SHEPP_LOGAN` due due #1049 .
This is corrected in #1050

**Review by johannesmayer — APPROVED** [review]

File: src/xGadgetron/cGadgetron/tests/CMakeLists.txt [hunk 1]
```diff
@@ -58,7 +58,7 @@ if(RUN_ISMRMRD_SHEPP_LOGAN)
         COMMAND ${CMAKE_COMMAND} -E env LD_LIBRARY_PATH=${CMAKE_INSTALL_PREFIX}/lib:$ENV{LD_LIBRARY_PATH} ${EXE_ISMRMRD_SL} -w 64 -a 2 -o "${TESTDATA_TEMP_PREFIX}_2.h5"
         WORKING_DIRECTORY ${TESTDATA_OUTPUT_PATH})
   else()
-    message(ERROR "We could not find the ISMRMRD executable for the Shepp Logan Generation." )
+    message(FATAL_ERROR "We could not find the ismrmrd_generate_cartesian_shepp_logan executable. Either set the variable to the -DEXE_ISMRMRD_SL=<path_to>ismrmrd_generate_cartesian_shepp_logan executable or disable the generation of the Shepp Logan phantom with -DRUN_ISMRMRD_SHEPP_LOGAN=OFF" )
```
KrisThielemans, line 61:
```suggestion
    message(FATAL_ERROR "We could not find the ismrmrd_generate_cartesian_shepp_logan executable. Either set the corresponding variable to the EXE_ISMRMRD_SL=<path_to>ismrmrd_generate_cartesian_shepp_logan_executable or disable the generation of the Shepp Logan phantom wby setting RUN_ISMRMRD_SHEPP_LOGAN=OFF" )
```

PR #1050, opened 2022-02-04
Ensuring an empty target header before AcquisitionsInfo::deserialize()

### LINKED ISSUE #1049: If the Ismrmrd versions don't match reading of MR data fails
`AcquisitionsInfo::deserialize()` is called twice if ISMMRD version number does not match.
This results in the duplication of encoding spaces inside the header.

Since only one is expected the sorting of the raw data fails and the program crashes

johannesmayer:
Fixes #1049


KrisThielemans:
I have no clue. @evgueni-ovtchinnikov ?

evgueni-ovtchinnikov (approved):
@johannesmayer the title of this PR is a bit confusing - so far as I can see, you empty `header_` *before* deserializing

PR #1059, opened 2022-02-17
Fix appending .h5 instead of overwriting issue

### LINKED ISSUE #1055: MR tests write .h5 files and append to existing files
Our `MRAcquisitionData::write()` will always append to existing files. Some MR cpp tests write .h5 files to disc.

Upon repeated local running of the tests during the development of something new this leads to that at some point the files are too large and cryptic errors appear about .h5 files.

Should we stop writing them in the tests?

evgueni-ovtchinnikov:
fixes #1055

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,9 @@
 
 ## v3.x.x
 
-* a version.h is created by CMake and installed to access version minor and major from C++.
+* To avoid appending to an existing `.h5` file, writing methods now first check whether the file to which data is to be written already exists, and if so, delete it before writing.
+
+* A version.h is created by CMake and installed to access version minor and major from C++.
 
 * MR Geometry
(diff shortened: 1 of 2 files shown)


evgueni-ovtchinnikov:
@johannesmayer please test on acquisition data (tested on image data only)

johannesmayer:
For the tests it didn't build for some siemens-to-ismrmrd reason, at least on my VM the C++ tests don't write to the same file any more.

KrisThielemans:
I think I've fixed the siemens_to_ismrmrd problem (we were running its `master`). in any case unrelated.

KrisThielemans:
So, this isn't just a change to the tests, but actually any writing will no longer append. correct? That's fine for me, but I would prefer to document this in `CHANGES.md`. Then feel free to merge.

PR #1062, opened 2022-02-18
Updated changes for PR #1008

johannesmayer:


KrisThielemans:
I meant like "SIRF currently supports sequences that use 2D and 3D cartesian sampling. If the Gadgetron toolboxes were found during building, in addition ..."

PR #1064, opened 2022-02-25
allow specifying TOF mashing for STIR

### LINKED ISSUE #1056: allow setting TOF mashing factor
We currently cannot set TOF info from SIRF. https://github.com/SyneRBI/SIRF/blob/434a85e5505e7a9602c1f6bf65afb3ebb9c4dc56/src/xSTIR/pSTIR/STIR.py#L756

No wonder, as it isn't available in STIR master either (we added some dummy TOF functions, but not that one)! Creating the issue here such that we don't forget.

KrisThielemans:
It will only be functional if you compiled the STIR TOF branch of course. However, you need to have STIR dated 25Feb2022 (i.e. https://github.com/UCL/STIR/commit/77ec42303521e3a84d8053d7511d929e06003084 or more recent) even without TOF as that added some TOF dummy functions.

Fixes #1056


KrisThielemans:
this should work, but there is test and we do not have the bandwidth to add this now, so postponing...

KrisThielemans:
@evgueni-ovtchinnikov could you add a `get_tof_mashing_factor` to `STIR.AcquisitionData`, which just calls the one from `stir::ProjDataInfo` (present from STIR 5.0). Then this should be ready.

PR #1067, opened 2022-03-07
Stir spect matrix

samdporter:
Tidied code using Codacy recommendations

Fixed bugs in SPECTUBMatrix Class

Small edits in acquisition_model.py to make image data and projection data compatible

SIRF builds fine. acquisition_model.py runs fine.

PR #1068, opened 2022-03-07
Stir SPECT matrix example

samdporter:
Added simple SPECT reconstruction example.

Uses OSEM with noisy data including an attenuation image

Runs fine


KrisThielemans:
interesting... what happened here? a good topic for a chat.

In any case, let's not merge `master` in #1067 until all done

samdporter:
> interesting... what happened here? a good topic for a chat.
> 
> In any case, let's not merge `master` in #1067 until all done

samdporter:
I have no idea what happened here. I think probably as a result of me being lazy and trying to edit everything within github (lesson learned).
This branch is no longer required as everything is in STIR_SPECT_Matrix. 
Not what I originally intended, but it should be OK

PR #1072, opened 2022-03-16
update SIRF version in CMakeList

paskino:
updates the version of SIRF in the CMakeLists.txt which remained at 3.2


KrisThielemans:
I believe we should go for 3.2.0 now. We have new features. People might depend on it. They need to know as soon as they are in `master`. (Of course, once there are more features, just having a version number Incrementing once isn't enough. I know).

PR #1074, opened 2022-03-18
Final Codacy corrections

samdporter:
That should be everything now


KrisThielemans:
Not a great idea to create a PR using a branch that I've squash-merged already in general. Better remove that branch in such cases.

I'll squash merge this, should work I hope.

PR #1075, opened 2022-03-19
minor changes to reduce conversion compiler warnings

KrisThielemans:
Visual Studio gives quite a few warnings on conversions. After this there are a few less.


KrisThielemans:
Still quite a few left for `ANumRef`::operator/=(float) when the underlying type is an `int` but these might be harder (as I think they are genuinely unsafe conversions)

PR #1079, opened 2022-03-21
work around problem finding HDF5 with older CMake

KrisThielemans:
ISMRMRD and STIR can use HDF5, but ISMRMRD needs C and STIR needs C++. This creates a conflict on older CMake, see https://gitlab.kitware.com/cmake/cmake/-/issues/20909

Until recently STIR prevented a conflict, but this will change now, as it actually broke STIR import as STIRConfig.cmake was just looking for the C HDF5 libraries. See https://github.com/UCL/STIR/blob/476947105233d5aa0336e16f1a437a43bca47894/src/cmake/STIRConfig.cmake.in#L72


KrisThielemans:
STIR now updated: https://github.com/UCL/STIR/commit/0b684d519352c85630e5bfe404392d257bb25ced

PR #1080, opened 2022-03-22
ImageWrap Constructor/destructor exception fixes

KrisThielemans:
- declare constructors that can throw as `noexcept(false)`
- catch exceptions in destructor

These prevent calls of `std::terminate`. Of course, in our case they would never occur as it would mean something wrong with the type spec, but it is safer and removed ugly compiler warnings.


evgueni-ovtchinnikov (approved):
Frankly, I am not familiar with this kind of stuff, but I guess it would not hurt using it.

PR #1088, opened 2022-04-20
Ensure Python int arrays exchanged with C++ contain compatible (sizeof(int)-bytes) integers

evgueni-ovtchinnikov:
At least on one occasion (see #1028), a Python `int` array passed to C++ turned out to contain 64-bit integers, whereas C++ expected 32-bit integers. This PR introduces checking the size of C++ `int` and creating/converting Python `int` arrays to be passed to C++ accordingly.

### DIFF
--- a/src/common/Utilities.py
+++ b/src/common/Utilities.py
@@ -32,18 +32,18 @@ RE_PYEXT = re.compile(r"\.(py[co]?)$")
 
 
 def cpp_int_bits():
-    """Returns the number of bits in a C++ integer"""
+    """Returns the number of bits in a C++ integer."""
     return pyiutil.intBits()
 
 
 def cpp_int_dtype():
-    """Returns numpy dtype corresponding to a C++ int"""
+    """Returns numpy dtype corresponding to a C++ int."""
     dt = 'int%s' % cpp_int_bits()
     return numpy.dtype(dt)
 
 
 def cpp_int_array(v):
-    """Converts the input into numpy.ndarray compatible with C++ int array"""
+    """Converts the input into numpy.ndarray compatible with C++ int array."""
     dt = numpy.dtype('int%s' % cpp_int_bits())
     if not isinstance(v, numpy.ndarray):
         v = numpy.array(v, dtype=dt)
--- a/src/common/Utilities.py
+++ b/src/common/Utilities.py
@@ -32,18 +32,18 @@ RE_PYEXT = re.compile(r"\.(py[co]?)$")
 
 
 def cpp_int_bits():
-    '''Returns the number of bits in C++ integer'''
+    """Returns the number of bits in a C++ integer"""
     return pyiutil.intBits()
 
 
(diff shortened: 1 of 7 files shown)


KrisThielemans (changes requested):
good idea.

I think it'd look a bit cleaner if you'd replace
```python
        dt = 'int%s' % cpp_int_bits()
        dim = numpy.ndarray((MAX_IMG_DIMS,), dtype=numpy.dtype(dt))
```
with
```python
        dt = 'int%s' % cpp_int_bits()
        dim = numpy.ndarray((MAX_IMG_DIMS,), dtype=cpp_int_dtype())
```
with
```python
def cpp_int_dtype():
        return numpy.dtype('int%s' % pyiutil.intBits())
```
I think you then no longer need `cpp_int_bits`.

Also, please add a doxygen string for `intBits` and Python docstring for the new functions.

src/common/Utilities.py
```diff
@@ -31,6 +31,29 @@
 RE_PYEXT = re.compile(r"\.(py[co]?)$")
 
 
+def cpp_int_bits():
+    '''Returns the number of bits in C++ integer'''
```
KrisThielemans, line 35:
```suggestion
    '''Returns the number of bits in a C++ integer'''
```

src/common/Utilities.py
```diff
@@ -31,6 +31,29 @@
...
+    return pyiutil.intBits()
+
+
+def cpp_int_dtype():
+    '''Returns numpy dtype corresponding to C++ int'''
```
KrisThielemans, line 40:
```suggestion
    '''Returns numpy dtype corresponding to a C++ int'''
```

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -37,6 +37,7 @@
 from sirf.Utilities import show_2D_array, show_3D_array, error, check_status, \
      try_calling, assert_validity, assert_validities, label_and_name, \
      name_and_parameters, parse_arglist, \
+     cpp_int_bits, cpp_int_dtype, cpp_int_array, \
```
KrisThielemans, line 40:
```suggestion
     cpp_int_dtype, cpp_int_array, \
```

src/xSTIR/pSTIR/STIR.py
```diff
@@ -34,6 +34,7 @@
 
 from sirf.Utilities import show_2D_array, show_3D_array, error, check_status, \
      try_calling, assert_validity, \
+     cpp_int_bits, cpp_int_dtype, cpp_int_array, \
```
KrisThielemans, line 37:
```suggestion
     cpp_int_dtype, cpp_int_array, \
```

KrisThielemans (changes requested):

KrisThielemans (approved):
fast :-)

PR #1100, opened 2022-04-28
created PR template

paskino:
Creates a PR template to easily handle PR.

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -22,4 +22,4 @@ Please read and adhere to the [contribution guidelines](https://github.com/SyneR
 
 Please tick the following: 
 
-- [ ] I declare that the contributed code is under [Apache-2.0](https://spdx.org/licenses/Apache-2.0.html) license.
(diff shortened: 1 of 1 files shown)


KrisThielemans (approved):

KrisThielemans:
I've changed this from using "you" for the writer to read as "informative" to the reader.

@DANAJK @evgueni-ovtchinnikov @ckolbPTB @gfardell any final changes for this? (People will get this text when they create a PR)

DANAJK:
It might be better to have all contribution instructions in [CONTRIBUTING.md](https://github.com/SyneRBI/SIRF/blob/master/CONTRIBUTING.md) rather than duplicating and extending instructions here.

If we are having a problem with specific tasks not being done, then maybe it is necessary.

ckolbPTB:
I think it is a good idea to have a comprehensive to-do list here which people can simply tick off. The hope is that people read [CONTRIBUTING.md](https://github.com/SyneRBI/SIRF/blob/master/CONTRIBUTING.md) before they start on a PR but by the time they submit it, it is helpful to remind them on the important points.

PR #1103, opened 2022-05-10
Enabling PET reconstruction from list mode data

### LINKED ISSUE #134: enable PET listmode reconstruction
needs different objective function

evgueni-ovtchinnikov:
Fixes #134.


KrisThielemans:
looks good to me. Now the objective function...

it should be virtually the same as the one for projection data. Main difference is that it only accepts a projection matrix (not a projector pair).

evgueni-ovtchinnikov:
@NikEfth I increased my VM memory to 16384, and got past set up, but the job was killed (it appears, not long before it would finish):
```
sirfuser@vagrant:~$ cstir_test6
running test6.cpp...

WARNING: Interfile warning: 'default bin size (cm)' (0.208000) is expected to be 0.200000.


WARNING: Interfile warning: num_axial_blocks_per_bucket (35) is expected to be 7.


WARNING: Interfile warning: axial crystals per singles unit (8) is expected to be 56.


WARNING: Interfile warning: transaxial crystals per singles unit (8) is expected to be 32.


WARNING: Interfile warning: I have used all explicit settings for the scanner
	from the Interfile header, and remaining fields set from the
	UPENN_5rings model.


WARNING: Scanner UPENN_5rings: inconsistent transaxial block info

WARNING: Interfile parsing ended up with the following scanner:
Scanner parameters:= 
Scanner type := UPENN_5rings
Number of rings                          := 280
Number of detectors per ring             := 594
Inner ring diameter (cm)                 := 76.4
Average depth of interaction (cm)        := 0.7
Distance between rings (cm)              := 0.39655
Default bin size (cm)                    := 0.208
View offset (degrees)                    := 0
Maximum number of non-arc-corrected bins := 331
Default number of arc-corrected bins     := 331
Energy resolution := 0.109
Reference energy (in keV) := 511
Number of blocks per bucket in transaxial direction         := 4
Number of blocks per bucket in axial direction              := 35
Number of crystals per block in axial direction             := 8
Number of crystals per block in transaxial direction        := 8
Number of detector layers                                   := 1
Number of crystals per singles unit in axial direction      := 8
Number of crystals per singles unit in transaxial direction := 8
Scanner geometry (BlocksOnCylindrical/Cylindrical/Generic)  := Cylindrical
end scanner parameters:=


image dimensions: 559x200x200
Offset: (-208, -208, -0)
Spacing: (2.08, 2.08, 1.98275)
Size: (200, 200, 559)
Direction matrix: 
1, 0, 0
0, 1, 0
0, 0, -1

== ok

WARNING: WARNING: ProjMatrixByBinUsingRayTracing used for pixel size (in x,y) that is larger than the bin size.
Backprojecting with this matrix might have artefacts at views 0 and 90 degrees.


WARNING: WARNING: ProjMatrixByBinUsingRayTracing used for pixel size (in x,y) that is larger than the bin size.
Backprojecting with this matrix might have artefacts at views 0 and 90 degrees.

tangential LORs: 2
Setting cache path...
Cache path: /home/sirfuser/devel/buildVM/sources/SIRF/data/examples/TBPET/
PoissonLogLikelihoodWithLinearModelForMeanAndListModeData: Skipping input!
Setting scanner template...
Dummy LM file

WARNING: The default max ring difference is set based full Scanner geometry
Setting acquisition model...
Setting max ring diff. ...
Sensitivity images: /home/sirfuser/devel/buildVM/sources/SIRF/data/examples/TBPET/sens_%d.hv
Setting max cache size ...
Max cache: 1500000000
Setting objective function ...
Setting up the reconstruction, please wait ...

WARNING: WARNING: ProjMatrixByBinUsingRayTracing used for pixel size (in x,y) that is larger than the bin size.
Backprojecting with this matrix might have artefacts at views 0 and 90 degrees.


WARNING: WARNING: ProjMatrixByBinUsingRayTracing used for pixel size (in x,y) that is larger than the bin size.
Backprojecting with this matrix might have artefacts at views 0 and 90 degrees.


WARNING: WARNING: ProjMatrixByBinUsingRayTracing used for pixel size (in x,y) that is larger than the bin size.
Backprojecting with this matrix might have artefacts at views 0 and 90 degrees.


WARNING: We skip the check on balanced subsets and presume they are balanced!

WARNING: We skip the check on balanced subsets and presume they are balanced!
Reconstructor set up.
Starting loop with 10 threads
Starting loop with 10 threads
Starting loop with 10 threads
Starting loop with 10 threads
Starting loop with 10 threads
Starting loop with 10 threads
Starting loop with 10 threads
Starting loop with 10 threads
Starting loop with 10 threads
Killed
sirfuser@vagrant:~$ 
```
Would it be possible to shorten this test run time by quitting early?

NikEfth:
Yes, can you try with this file?

evgueni-ovtchinnikov:
@NikEfth Thanks a lot, it works much faster now and does not get killed!

NikEfth:
Sure, but the images you get should not make sense

evgueni-ovtchinnikov:
@NikEfth not important at this stage

src/xSTIR/pSTIR/STIR.py
```diff
@@ -856,7 +856,53 @@ def set_resolution_model(self, collimator_sigma_0_in_mm, collimator_slope_in_mm,
...
+        print("Nikos1111")
+        self.handle = pystir.cSTIR_objectFromFile('ListmodeData', filename)
+        check_status(self.handle)
+        self.read_only = True
+
+    def get_cache_path(self): 
+        return "Nikos"
+
+    def set_cache_path(self): 
+        print("NikosSEt")
```
KrisThielemans, line 900: this need changing

src/xSTIR/pSTIR/STIR.py
```diff
@@ -2414,6 +2460,64 @@ def set_acquisition_data(self, ad):
             self.handle, self.name, 'acquisition_data', ad.handle)
 
 
+class PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin(ObjectiveFunction):
+    """Class for a STIR type of Poisson loglikelihood object.
```
KrisThielemans, line 2464:
```suggestion
    """Class for a STIR type of Poisson loglikelihood object for listmode data.
```

src/xSTIR/pSTIR/STIR.py
```diff
@@ -3040,21 +3144,29 @@ def set_relaxation_parameter(self, value):
             self.handle, self.name, 'relaxation_parameter', value)
 
 
-def make_Poisson_loglikelihood(acq_data, likelihood_type='LinearModelForMean',
+#def make_Poisson_loglikelihood(acq_data, likelihood_type='LinearModelForMean',
+def make_Poisson_loglikelihood(acq_data=None, likelihood_type=None,
```
KrisThielemans, line 3148: not sure why we're changing this default here

src/xSTIR/pSTIR/STIR.py
```diff
@@ -3040,21 +3144,29 @@ def set_relaxation_parameter(self, value):
...
     model types.
     """
     # only this objective function is implemented for now
-    if likelihood_type == 'LinearModelForMean':
+    #if likelihood_type == 'LinearModelForMean':
+    if likelihood_type is None:
         obj_fun = PoissonLogLikelihoodWithLinearModelForMeanAndProjData()
-        obj_fun.set_acquisition_data(acq_data)
+        if acq_data is not None:
+            obj_fun.set_acquisition_data(acq_data)
```
KrisThielemans, line 3160: not sure why we're changing this code here. Do we really need to? What happens if `acq_data` is `None`? It'd presumably crash. (Only the listmode case can handle this)

evgueni-ovtchinnikov: @KrisThielemans so far as I can see, Nikos' list mode objective function does not need acquisition data, so `acq_data` must be optional. We need to discuss this with Nikos.

src/xSTIR/cSTIR/tests/test6.cpp
```diff
@@ -0,0 +1,129 @@
...
+See the License for the specific language governing permissions and
+limitations under the License.
+
+*/
+
+/*!
+\file
+\ingroup PET
+
+\author Nikos Efthimiou
```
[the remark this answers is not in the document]
evgueni-ovtchinnikov: Nikos did

src/xSTIR/pSTIR/STIR.py
KrisThielemans: Agreed. @NikEfth's listmode function can either use listmode data or the corresponding cache.

However, these lines are for the "old" case with projection data, which do need it.

src/xSTIR/pSTIR/STIR.py
```diff
@@ -893,11 +894,11 @@ def read_from_file(self, filename):
...
 
-    def get_cache_path(self): 
-        return "Nikos"
+    def get_cache_path(self):
+        return self.cache_path
 
-    def set_cache_path(self): 
-        print("NikosSEt")
+    def set_cache_path(self, path):
+        self.cache_path = path
```
KrisThielemans, line 901: wouldn't this need a call to the C++ function?

src/xSTIR/pSTIR/STIR.py
evgueni-ovtchinnikov: ok, will make this function throw in the non-listmode case

KrisThielemans: presumably the old code did already, but if not, best to put it in, agreed.

I still think we can revert the default to what it as, such that we have minimal code change, but as you prefer.

src/xSTIR/pSTIR/STIR.py
evgueni-ovtchinnikov: so far as I can see, `cache_path` is needed by Nikos' objective function, so my guess is that the latter will obtain it from the `LismodeData` object passed either via `acq_data` argument of `make_Poisson` function or via its method `set_listmode_data` or suchlike - @NikEfth am I getting it right?

NikEfth: This function should get the cache path, as we set it in the test6. 
It was meant as means to test if  I can correctly set the path from python but didn't finish it. 

If we set the path correctly, we can remove the get function, otherwise, get_cache_path() is the function we should return.

evgueni-ovtchinnikov:
The first of the GHA builds fails with lots of messages of the kind
```
/home/runner/work/SIRF/SIRF/src/xSTIR/cSTIR/include/sirf/STIR/stir_x.h: In member function ‘void sirf::xSTIR_PoissonLLhLinModMeanListDataProjMatBin3DF::set_acquisition_data(std::shared_ptr<sirf::PETAcquisitionData>)’:
/home/runner/work/SIRF/SIRF/src/xSTIR/cSTIR/include/sirf/STIR/stir_x.h:1046:13: error: ‘set_proj_data_info’ was not declared in this scope
             set_proj_data_info(*sptr->data());
             ^~~~~~~~~~~~~~~~~~
```
I do not get such messages on my VM that uses Nikos' STIR branch, and since it has been merged onto STIR master, I suspect GHA uses some old STIR version.

In the second one `PET_TESTS_CPLUSPLUS_6` fails because it looks for the data in `SIRF/data/examples`, where it should not be placed I believe - could it not be downloaded during the build into `install/share` folder instead?

evgueni-ovtchinnikov:
@NikEfth yes, the simplest solution is to remove `cstir_test6` from the tests run by GHA but keep in in `install/bin`. Also, in `run_test6.cpp`, the path to data should be specified entirely by the user via `argv[1]` without using `SIRF_PATH` as the test data should sit outside of `SIRF` folder.

src/xSTIR/cSTIR/tests/CMakeLists.txt
```diff
@@ -25,6 +25,13 @@
...
 
+  add_executable(cstir_test6  test6.cpp run_test6.cpp ${STIR_REGISTRIES})
+  target_link_libraries(cstir_test6 csirf cstir ${STIR_LIBRARIES})
+  INSTALL(TARGETS cstir_test6 DESTINATION bin)
+
 ADD_TEST(NAME PET_TESTS_CPLUSPLUS_1 COMMAND cstir_tests WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
 
 ADD_TEST(NAME PET_TESTS_CPLUSPLUS_4 COMMAND cstir_test4 WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
+
+ADD_TEST(NAME PET_TESTS_CPLUSPLUS_6 COMMAND cstir_test6 WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR})
```
KrisThielemans, line 36: comment this out until data situation is sorted

KrisThielemans:
> The first of the GHA builds fails with lots of messages of the kind
> ...

true, the first job runs with STIR release_5.0. Would it be easy to do things like
```
#if STIR_VERSION < 500100
  throw some error
#else
  obj_fun.set_cache_max_size(dataFromHandle<int>(hv));
#endif
```
This would only have to be done for the new members added by @NikEfth. On the other hand, that might be pretty much all `set/get` functions.  So maybe we should put (nearly) everything in the new listmode class support in an #ifdef. I suppose it then becomes ugly on the Python side. sigh.

Let me know if not doable. I can release STIR 5.1 probably tomorrow in any case.

src/xSTIR/cSTIR/cstir_p.cpp
```diff
@@ -660,6 +663,50 @@ sirf::cSTIR_setPoissonLogLikelihoodWithLinearModelForMeanAndProjDataParameter
...
+void*
+sirf::cSTIR_setPoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBinParameter
+(DataHandle* hp, const char* name, const DataHandle* hv)
+{
+    xSTIR_PoissonLLhLinModMeanListDataProjMatBin3DF&
+        obj_fun = objectFromHandle
+        <xSTIR_PoissonLLhLinModMeanListDataProjMatBin3DF>
+        (hp);
+    std::string arg(name);
+    const char* prefix = arg.substr(0, 10).c_str();
```
KrisThielemans, line 675: why 0-10?

src/xSTIR/cSTIR/cstir_p.cpp
```diff
@@ -660,6 +663,50 @@ sirf::cSTIR_setPoissonLogLikelihoodWithLinearModelForMeanAndProjDataParameter
...
+    std::string cp("cache_path");
+    if (sirf::iequals(prefix, cp)) {
+        std::string s(charDataFromDataHandle(hv));
+        auto found = s.find(",");
```
KrisThielemans, line 679: why are we looking for a comma?

src/xSTIR/cSTIR/include/sirf/STIR/stir_data_containers.h
```diff
@@ -135,17 +135,31 @@ namespace sirf {
...
 	};
 
+	/*!
+	\ingroup PET
+	\brief Abstract base class for PET scanner data
+	*/
+	class PETScanData : public DataContainer {
+	public:
+		virtual ~PETScanData() {}
+		virtual stir::shared_ptr<stir::ExamData> data_sptr() = 0;
```
KrisThielemans, line 145: not needed

src/xSTIR/pSTIR/STIR.py
```diff
@@ -3040,21 +3145,32 @@ def set_relaxation_parameter(self, value):
...
+        if acq_data is not None: 
+            obj_fun.set_acquisition_data(acq_data)
     else:
-        raise error(
-            'only PoissonLogLikelihoodWithLinearModelForMeanAndProjData ' +
-            'is currently implemented in SIRF')
+        raise error('Poisson_loglikelihood of type ' + likelihood_type + \
+                    ' is not implemented')
+#            'only PoissonLogLikelihoodWithLinearModelForMeanAndProjData ' +
+#            'is currently implemented in SIRF')
```
KrisThielemans, line 3173:
```suggestion
```

KrisThielemans (changes requested):

src/xSTIR/pSTIR/STIR.py
```diff
@@ -856,7 +856,54 @@ def set_resolution_model(self, collimator_sigma_0_in_mm, collimator_slope_in_mm,
...
+        print("Nikos1111")
+        self.handle = pystir.cSTIR_objectFromFile('ListmodeData', filename)
+        check_status(self.handle)
+        self.read_only = True
+
+    def get_cache_path(self):
+        return self.cache_path
+
+    def set_cache_path(self, path):
+        self.cache_path = path
```
KrisThielemans, line 901:
```suggestion
```

KrisThielemans:
@evgueni-ovtchinnikov this no longer compiles after the merge as `DataContainer` now defines pure virtuals for the algebraic operations. However, they don't make sense for list-mode data.

Even members like `bits()` don't really (`items()` could make some sense, although there is a question what it would mean as list-mode data generally contains different types of "records" (e.g. prompts, delayed, singles, external triggers, time stamps, etc), but is generally unknown without traversing the whole list, which is a very expensive operation.

The proposed hierarchy here was
```C++
	class STIRScanData : public DataContainer {
	public:
		virtual ~STIRScanData() {}
		virtual stir::shared_ptr<stir::ExamData> data_sptr() const = 0;
	};
class STIRAcquisitionData : public STIRScanData {};

class ListmodeData : public DataContainer {};
```
I suppose we could not derived `STIRScanData` from `DataContainer`, but conceptually that sounds a bit weird.

I think we need to discuss the `DataContainer` concept class therefore.

KrisThielemans:
Also, there are a large number of Codacy issues due to the "old" style of the `exec` in the python scripts. Can you fix those?

KrisThielemans:
This worked on the C++ front. However, Python tests fail with segfault when calling `STIR.AcquisitionData.apxby`. I believe this is because of problem casting when going from handles to objects. I've tried to cut this down locally by making `STIRScanData` as small as possible, and I can still reproduce it. Here are some snippets:
```c++
class STIRScanData {
 public:
 virtual ~STIRScanData() {}
};

class STIRAcquisitionData : public STIRScanData, public DataContainer {...}

extern "C" void* cSIRF_axpby(
	const void* ptr_a, const void* ptr_x, const void* ptr_b, const void* ptr_y) {
try {
	auto const& x =	objectFromHandle<DataContainer>(ptr_x);
       ...
}
```
where `objectFromHandle` does a (C-style) cast via `CAST_PTR(std::shared_ptr<Object>, ptr_sptr, ptr);`

The problem seems to be that if `STIRAcquisitionData` is derived from both `STIRScanData` and `DataContainer`, this C-style cast fails (i.e. points to the wrong bit of memory, and therefore causing a segfault).

The good news is that I can get rid of this segfault by removing `STIRScanData` from `STIRAcquisitionData` and Python algebra tests get through.

**However**, then I don't know what to do for the reconstruction. Code looks like this
```c++
sirf::cSTIR_setReconstructionParameter
(DataHandle* hp, const char* name, const DataHandle* hv)
{
	Reconstruction<Image3DF>& recon =
		objectFromHandle< Reconstruction<Image3DF> >(hp);
	if (sirf::iequals(name, "input_data")) {
          // what to do here?
          //SPTR_FROM_HANDLE(STIRScanData, sptr_ad, hv);
          SPTR_FROM_HANDLE(STIRAcquisitionData, sptr_ad, hv);
		recon.set_input_data(sptr_ad->data_sptr());
	}
	else ...
}
```
`stir::Reconstruction::set_input_data` takes a `stir::ExamData`, which I tried to wrap via `STIRScanData`. I can no longer do that if `ListmodeData` and `STIRAcquisitionData` are not derived from the same SIRF class I guess.

@evgueni-ovtchinnikov is there a way to discover the type of the object that the handle is for?

KrisThielemans:
> **However**, then I don't know what to do for the reconstruction.

@evgueni-ovtchinnikov @paskino Introducing an almost empty class on top of `DataContainer` indeed allows me to use C++ RTTI, so this problem is resolved like this
https://github.com/SyneRBI/SIRF/blob/61e692d7bced3593e867ace9271fbd3878e99a8d/src/xSTIR/cSTIR/cstir_p.cpp#L1056-L1064

So I think we can go ahead with this approach.

I haven't checked the Python side yet. We did introduce a `ScanData` class in Python, but have now no correspondence in C++ anymore. But maybe we don't need it.

KrisThielemans:
things to do:
- Python test
- remove `osem_lm_reconstruction.py` example
- update 'listmode_reconstruction.py` example to use `ListmodeData` object as opposed to filename
- change name of objective function

KrisThielemans:
`listmode_reconstruction.py` example now works (i.e. runs through and produces an image. Not 100% sure if it is correct. Could be that the acquisition model isn't passed through completely.

`ListmodeToSinograms.set_input()` now accepts an `ListmodeData` object as well.

examples/Python/PET/reconstruct_from_listmode.py
```diff
@@ -23,7 +23,7 @@
   -C <cnts>, --counts=<cnts>   account for delay between injection and acquisition start by shifting interval to start when counts exceed given threshold.
   --visualisations             show visualisations
   --nifti                      save output as nifti
-  --gpu                        use gpu
+  --gpu                        use gpu (actually means: use NiftyPET projectors)
```
KrisThielemans, line 26: Please keep removed. The STIR listmode recon only supports projmatrixbybin at the moment.

KrisThielemans:
Best to remind all use_gpu lines from the demo, as it isn't supported.

KrisThielemans (changes requested):
Actually, the whole osem_lm_reconstruction.py needs to be removed. I think everything in it is in  listmode_reconstruction.py, which is more complete. Some things on the osem demo are currently not supported anyway.

KrisThielemans:
I think that later we should move the preparation of randoms, scatter etc to SIRF itself, i.e. taking most of what you wrote for the challenge. However, let's do that in a separate PR, and merge this one asap.

examples/Python/PET/reconstruct_from_listmode.py
evgueni-ovtchinnikov: do you mean go back to just `use gpu`?

KrisThielemans:
I guess this is a problem on your VM. You should probably give it more memory.

It's not clear where this happens. If it's during scatter estimation (as  the above implies), reducing number of listmode events is not going to change anything, as the scatter estimate is a (full resolution)  projdata.

In any case, the number of events used here is very small.

KrisThielemans:
I think that later we should move the preparation of randoms, scatter etc to SIRF itself, i.e. taking most of what you wrote for the challenge. However, let's do that in a separate PR, and merge this one asap.

evgueni-ovtchinnikov:
> I guess this is a problem on your VM. You should probably give it more memory.

True. I have increased VM memory from 6GB to 12GB, and `listmode_reconstruction.py` now runs ok (takes a lot of time though).

Permission to merge?

KrisThielemans:
Great. Did you remove the osem_lm_reconstruction.py? Otherwise good to go.

KrisThielemans:
Oh. Don't forget to add something to the doc and changes.md

PR #1120, opened 2022-06-24
Implement set/get modality methods for SIRF ImageData

### LINKED ISSUE #1078: Silencing WARNING (setting SIRF ImageData modality)
I am creating an image data object and filling it like so:

```
im = pet.ImageData()
im.initialise(arr.shape, tuple(res))
im.fill(arr)
```
Where arr is the numpy array of data and res is the physical resolution. The data template is taken simply from the SIRF examples. When running my algorithm I get the following warning, which I would like silence:

**WARNING: forward_project. Imaging modality unknown for either the image or the projection data or both.**

I have tried `sirf.STIR.set_verbosity(False)` but to no avail. Perhaps there is a way of assigning my own modality to the ImageData object.

Thanks,
Imraj

evgueni-ovtchinnikov:
Fixes #1078.


src/xSTIR/cSTIR/include/sirf/STIR/stir_data_containers.h
```diff
@@ -893,12 +890,24 @@ namespace sirf {
...
 			return 1;
 		}
-        /// Write to file
+
+		std::string modality() const
+		{
+			ExamInfo ex_info = data().get_exam_info();
+			return ex_info.imaging_modality.get_name();
+		}
+		void set_modality(const char* mod)
```
KrisThielemans, line 903: I prefer `const std::string&`. `char*` gets converted to `string`, but not the other way around. Of course, this is in many other places as well, but we can start here.

(If you need to keep `char *`, much better is `const char * const`)

KrisThielemans (changes requested):
looks good. simpler as well! Great

src/xSTIR/cSTIR/tests/test1.cpp
```diff
@@ -118,6 +119,14 @@ int test1()
 			std::cout << "== failed \n";
 		fail = fail || !ok;
 
+		// show and change modality demo
+		std::string mod = image_data.modality();
+		std::cout << "modality: " << mod << '\n';
+		image_data.set_modality("SPECT");
+		std::cout << "modality: " << image_data.modality() << '\n';
```
KrisThielemans, line 126: can we convert this to a test?

src/xSTIR/pSTIR/STIR.py
```diff
@@ -323,6 +323,12 @@ def same_object(self):
         """See DataContainer method."""
         return ImageData()
 
+    def modality(self):
```
KrisThielemans, line 326: docstring

src/xSTIR/pSTIR/STIR.py
```diff
@@ -323,6 +323,12 @@ def same_object(self):
...
+        return parms.char_par(self.handle, 'ImageData', 'modality')
+
+    def set_modality(self, mod):
```
KrisThielemans, line 329: docstring with allowed values

KrisThielemans (changes requested):
I see you're using "SPECT" but STIR uses DICOM conventions, which I think is a good idea. (even though it doesn't document it!). See https://github.com/UCL/STIR/blob/b7352113eba3c28087be930d1d511d08745361e6/src/include/stir/ImagingModality.h#L84-L89 for allowed values.

KrisThielemans:
> I see you're using "SPECT" but STIR uses DICOM conventions, which I think is a good idea. (even though it doesn't document it!). 

https://github.com/UCL/STIR/pull/1053 documents, but also standardises. constructing with "SPECT" therefore will no longer return "SPECT". Best to write the test using "NM".

PR #1125, opened 2022-07-01
Trying to fix dicom output for undersampled data. 

### LINKED ISSUE #1122: MRImageData.write not working for dicoms
The writer for the MR image data does not connect properly to Gadgetron to write Dicom files.
If the `FullySampledReconstructor` is the reconstruction chain then 
If you use the `CartesianGRAPPAReconstructor` then it will not connect to the Gadgetron server and not write anything.

The script producing the error is below
```
path_in = '/some/path/to/data/'
path_out '/some/other/path/'

file_in = 'some_cartesian_cine_grappa_data.h5'
rawdata = pMR.AcquisitionData(path_in + '/' + file_in)

rawdata = pMR.preprocess_acquisition_data(rawdata)
rawdata.sort()

recon = pMR.FullySampledReconstructor()
#recon = pMR.CartesianGRAPPAReconstructor() # this will not write dicoms

recon.set_input(rawdata)
recon.process()
img_data = recon.get_output()


# some scaling is necessary

img_data = img_data.abs()
img_values = img_data.as_array()
img_values = (img_values - np.min(img_values))/ (np.max(img_values) - np.min(img_values))
img_data = img_data.fill(img_values)

img_data *= 2^16
file_out = file_in.replace('.h5', '.dcm')
img_data.write(path_out + '/' + file_out)
```

evgueni-ovtchinnikov:
Fixes #1122.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,6 @@
 # ChangeLog
 
-## v 3.3.1
+## v3.3.1
 
 * writing MR images and their attributes to `.dcm` and `.xml` files fixed.
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 # ChangeLog
 
+## v 3.3.1
+
+* writing MR images and their attributes to `.dcm` and `.xml` files fixed.
+
+
 ## v3.3.0
 
 * SPECTUBMatrix for (simple) usage in SPECT added.
--- a/examples/Python/MR/fully_sampled_recon.py
+++ b/examples/Python/MR/fully_sampled_recon.py
@@ -80,7 +80,7 @@ def main():
     recon.process()
     
     # retrieve reconstruced image data
-    image_data = recon.get_output()
+    image_data = recon.get_output().real()
     image_data.write(args['--output'])
 
     if show_plot:
--- a/examples/Python/MR/grappa_basic.py
+++ b/examples/Python/MR/grappa_basic.py
@@ -103,6 +103,7 @@ def main():
 
     # retrieve reconstruced image and G-factor data
     image_data = recon.get_output('image')
+    #print(image_data.dimensions())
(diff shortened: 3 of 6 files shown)


evgueni-ovtchinnikov:
@johannesmayer @KrisThielemans the trouble with the dicom output is apparently caused by Gadgetron people switching some time ago from `xml` to `ISMRMRD::MetaContainer`s format for MR images attributes. This does not affect our fully sampled tests/demos as the images these tests/demos create do not have attributes.

I have for now applied quick fix, setting the size of the attribute to zero in `GadgetronClientConnector::send_ismrmrd_image`. The dicom files are now created for undersampled data too, but I could not manage to verify whether these files are any good - @johannesmayer could you?

I will try to implement a more proper fix after I return from my two-weeks holiday on 18 July.

evgueni-ovtchinnikov:
The first sentence of my previous comment turned out to be wrong - the actual culprit was the fact that the attributes of the image created by `ImageWrap::real` were undefined - now fixed by adding just one missing line in `ImageWrap::real`:
```
			im.setAttributeString(attributes());
```
I thought I would need the same for the images algebra, where new images are also created, but luckily they are created by copy-constructing, which takes care about attributes. Ideally, the attributes of operands should better be checked for compatibility, but this I believe deserves a separate issue/PR by somebody much more competent than myself - @DANAJK? @ashgillman?

evgueni-ovtchinnikov:
@johannesmayer @KrisThielemans I now have fixed the dicom output properly, with attributes saved by Gadgetron to `.xml` files parallel to `.dcm` files with images - ok to merge?

KrisThielemans:
great. fine for me. Do add something in CHANGES.md before merging please.

evgueni-ovtchinnikov:
> Do add something in CHANGES.md before merging please.

Had to put it in v3.3.1 as v3.3.0 has been tagged already.

PR #1142, opened 2022-10-20
[GHA] test with gcc 8 and 9

### LINKED ISSUE #1140: GitHub actions are failing due to missing gcc8
https://github.com/SyneRBI/SIRF/actions/runs/3189378929/jobs/5203118530#step:5:77

I think I've fixed this in the SuperBuild. @paskino maybe you can have a look?

KrisThielemans:
port some changes from the SIRF-SuperBuild.

Fixes #1140, as we no longer assume gcc-8 is installed.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -20,11 +20,31 @@ jobs:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
-        # the agent machine operating systems
-        os: [ubuntu-latest]
-        DEVEL_BUILD: ["OFF", "ON"]
-        EXTRA_BUILD_FLAGS: ["-DUSE_ITK=ON"]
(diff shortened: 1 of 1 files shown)


KrisThielemans:
fails with
```
Error: .github/workflows/build-test.yml (Line: 94, Col: 30): A sequence was not expected
Error: .github/workflows/build-test.yml (Line: 95, Col: [2](https://github.com/SyneRBI/SIRF/actions/runs/3291458693/jobs/5425633867#step:6:2)9): A sequence was not expected
```
but it seems to be valid YAML. @paskino any idea?

KrisThielemans:
https://rhysd.github.io/actionlint/ gave some (cryptic) errors on the use of `${{ matrix.blabla }}` so I've got rid of then `env` entirely

KrisThielemans:
The Gadgetron version that we use fails to compile with gcc 10, so disabling that

PR #1158, opened 2022-12-09
Dropped TO_BE_IGNORED in MRAcquisitionData::read

### LINKED ISSUE #1156: SIRF Reading & Writing and gadgetron_ismrmrd_client collide
I encountered a problem executing the following steps:
- reading an ISMRMRD raw file (from patient data) with SIRF and then writing it back to file.
- calling the `gadgetron_ismrmrd_client`  on that file to reconstruct
- (I only checked the output for a DICOM writer in the XML configuration file for the gadget chain)

all reconstructions yield zeros.

I have found the solution:
https://github.com/SyneRBI/SIRF/blob/93f280fbfc95f914642d1a532a65882d83938eb5/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h#L62

to `false`  for all acquisitions will solve this problem.

evgueni-ovtchinnikov:

Dropped `TO_BE_IGNORED` in `MRAcquisitionData::read`.

## Testing performed

All MR tests passed.

## Related issues

fixes #1156

## Contribution Notes

Please tick the following: 

### DIFF
--- a/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
+++ b/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
@@ -158,10 +158,7 @@ MRAcquisitionData::read( const std::string& filename_ismrmrd_with_ext )
 			d.readAcquisition( i_acqu, acq);
 			mtx.unlock();
 
-			if( TO_BE_IGNORED(acq) )
-				continue;
-			else
-				this->append_acquisition( acq );
+			this->append_acquisition( acq );
 		}
         this->sort_by_time();
 		if( verbose )


johannesmayer:
I have found that this will crash the `dimensions()` function as well.
In particular the line 

https://github.com/SyneRBI/SIRF/blob/c06c53f47c5bd68dad5d4822c00c9d91ecfc518f/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp#L203

does not throw an error.
For patientdata where one noise sample is contained the dimension function returns `(1,1,1)` and breaks other functions such as `get_ISMRMRD_info()` that depend on it, without throwing any errors.

johannesmayer:
Should we maybe put the `TO_BE_IGNORED` macro into the getter function for acquisitions? this way they stay in the dataset but the code will never access them.

evgueni-ovtchinnikov:
> Should we maybe put the TO_BE_IGNORED macro into the getter function for acquisitions?

Brilliant idea @johannesmayer - I wish I thought about it when introducing the getter for acquisitions, would save a lot of  uses of `TO_BE_INGORED`.

We need to add an extra argument `all` though, defaulting to `false`, to allow getting all acquisitions.

johannesmayer:
I think it might still be dangerous if it is not made sure to check all the dimensions() functions analogous to #1160 .

evgueni-ovtchinnikov:
> I think it might still be dangerous if it is not made sure to check all the dimensions() functions analogous to https://github.com/SyneRBI/SIRF/issues/1160 .

I thought you created PR #1161 for that, did you not?

johannesmayer:
That's true.

KrisThielemans:
> we would get
> 
> ```
> 	if (!get_acquisition(i, a)) {
> 		...
> 	}
> ```
> 
> the meaning of which would also be somewhat less obvious.

it might be somewhat less obvious, but it is less verbose and also less error-prone (the compiler will warn when the return value isn't used, while it won't warn when `TO_BE_IGNORED` is not used). It's also easier to debug as most debuggers struggle with macros.

KrisThielemans:
By the way, should these PRs be merged before 3.4 release? They seem somewhat drastic to me.

evgueni-ovtchinnikov:
> By the way, should these PRs be merged before 3.4 release? They seem somewhat drastic to me.

@KrisThielemans IMHO corrections in both PRs are pretty small and obvious and affect 2 files only. I suggest we merge them now as they are, whereas reducing the use of `TO_BE_IGNORED`, which is indeed rather drastic, should better be done after the release 3.4.

KrisThielemans:
ok. go ahead with what you think is best

evgueni-ovtchinnikov:
> That's true. But I would merge https://github.com/SyneRBI/SIRF/pull/1161 first.

@johannesmayer ok, please respond to my comments and merge then. Or, if you do not have time for it right now, please let me know, and I will simply add your corrections to my PR, as I want this to be sorted before release 3.4.

KrisThielemans:
@evgueni-ovtchinnikov as maintainer, you can push to any branch fro @johannesmayer from which he generated a PR (unless he disabled it), so best way seems to be to put your corrections in his PR.

PR #1161, opened 2022-12-13
Fix mrdata dimensions

### LINKED ISSUE #1160: AcquisitionData.dimensions() does not catch errors
The function AcquisionData.dimensions() does not catch errors from C++

In line 
https://github.com/SyneRBI/SIRF/blob/c06c53f47c5bd68dad5d4822c00c9d91ecfc518f/src/xGadgetron/pGadgetron/Gadgetron.py#L965
the call to the C-function should be contained in a `try_calling()` otherwise errors that occur on C++ level are not caught.

johannesmayer:
Checks dimensions of k-space only for acquisitions that are not ignored.
Catches runtime errors that are caused by C++ code in sirf.Gadgetron.AcquisitionData.dimension()` function

## Testing performed
Used code on file which previously caused wrong dimensions. Yielded correct dimensions afterwards.

## Related issues

Fixes #1160 

## Contribution Notes

Please tick the following: 

### DIFF
--- a/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
+++ b/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
@@ -193,11 +193,17 @@ MRAcquisitionData::get_acquisitions_dimensions(size_t ptr_dim) const
     int* dim = (int*)ptr_dim;
     ISMRMRD::Acquisition acq;
     get_acquisition(0, acq);
+    int counter_first_valid_acq = 1;
+    while(TO_BE_IGNORED(acq) && counter_first_valid_acq < this->number())
+    {
+        get_acquisition(counter_first_valid_acq, acq);
+        counter_first_valid_acq++;
+    }
 
     int ns = acq.number_of_samples();
     int nc = acq.active_channels();
 
-    for(int i=1; i<na; ++i)
+    for(int i=counter_first_valid_acq; i<na; ++i)
     {
         get_acquisition(i, acq);
         ASSERT(acq.number_of_samples() == ns, "One of your acquisitions has a different number of samples. Please make sure the dimensions are consistent.");
--- a/src/xGadgetron/pGadgetron/Gadgetron.py
+++ b/src/xGadgetron/pGadgetron/Gadgetron.py
@@ -962,8 +962,8 @@ class AcquisitionData(DataContainer):
         if self.number() < 1:
             return numpy.zeros((MAX_ACQ_DIMENSIONS,), dtype=cpp_int_dtype())
         dim = numpy.ones((MAX_ACQ_DIMENSIONS,), dtype=cpp_int_dtype())
-        hv = pygadgetron.cGT_getAcquisitionDataDimensions\
-             (self.handle, dim.ctypes.data)
+        hv = try_calling(pygadgetron.cGT_getAcquisitionDataDimensions\
+             (self.handle, dim.ctypes.data))
         pyiutil.deleteDataHandle(hv)
         dim[2] = numpy.prod(dim[2:])
         return tuple(dim[2::-1])


src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -962,8 +962,8 @@ def dimensions(self):
         if self.number() < 1:
             return numpy.zeros((MAX_ACQ_DIMENSIONS,), dtype=cpp_int_dtype())
         dim = numpy.ones((MAX_ACQ_DIMENSIONS,), dtype=cpp_int_dtype())
-        hv = pygadgetron.cGT_getAcquisitionDataDimensions\
-             (self.handle, dim.ctypes.data)
+        hv = try_calling(pygadgetron.cGT_getAcquisitionDataDimensions\
```
evgueni-ovtchinnikov, line 965: `try_calling` has no return value (so the next line is no longer needed) - see `src/common/Utilities.py`

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
```diff
@@ -193,11 +193,17 @@ MRAcquisitionData::get_acquisitions_dimensions(size_t ptr_dim) const
...
+    {
+        get_acquisition(counter_first_valid_acq, acq);
+        counter_first_valid_acq++;
+    }
 
     int ns = acq.number_of_samples();
     int nc = acq.active_channels();
 
-    for(int i=1; i<na; ++i)
+    for(int i=counter_first_valid_acq; i<na; ++i)
```
evgueni-ovtchinnikov, line 206: same edits for `get_trajectory_dimensions` and `get_kspace_dimensions`?

src/xGadgetron/pGadgetron/Gadgetron.py
evgueni-ovtchinnikov: to be sure, by the next line I mean line 967 (`try_calling` takes care about the handles returned by C interface functions)

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
evgueni-ovtchinnikov: Actually, the situation is more complicated, I am afraid. If some acquisitions are to be ignored, `dim[2]` needs to be reduced accordingly in order for `fill` and `as-array` to work with correct dimensions, and **all** computations with acquisition data **must** do `TO_BE_IGNORED` checks, which makes it rather pointless to keep ignored acquisitions at all. And if we do not keep them, then `TO_BE_IGNORED` would only be needed when reading from file, and I believe we should issue warning message for every ignored acquisition to make the users aware that the data SIRF reads is not identical to the data in the file.

This all should be well thought through and discussed, so I suggest we postpone this PR and PR #1158 until after Release 3.4.

evgueni-ovtchinnikov: It turns out that `gadgetron_ismrmrd_client` using `Generic_Cartesian_Grappa.xml` without conversion to unsigned shorts produces expected images.

Apparently, the pixel values of the reconstructed images are too small to be represented by non-zero unsigned shorts, which suggests Gadgetron scaling problems in the absence of noise calibration data. I will create a Gadgetron issue for this bug when we switch to using up-to-date Gadgetron.

Meanwhile I suggest adding second `int`/`bool` argument `all` to `MRAcquisitionData::read`, defaulting to `false`, that would allow the user to disable ignoring acquisitions, and the same argument to `AcquisitionsVector` constructor-from-file etc. I volunteer to do that and also fix dimensions bugs (in fact, it looks like ignoring acquisitions in `MRAcquisitionData::read` was merely my quick fix for those bugs, so a proper fix is on me).

@johannesmayer are you joining our tomorrow zoom?

evgueni-ovtchinnikov:
taken care of by #1174, closing...

PR #1162, opened 2022-12-15
[GHA] fix gcc8 test

KrisThielemans:
force gcc-8 test to ubuntu 20.04 as it is no longer in ubuntu 22.04


KrisThielemans:
@evgueni-ovtchinnikov this fixes the first job on GHA.

evgueni-ovtchinnikov:
@KrisThielemans Thanks a lot!

PR #1168, opened 2023-01-18
Sorted out various issues with PET demos

evgueni-ovtchinnikov:

* added missing `--non-interactive` option and skipping data displays in non-interactive mode
* added missing default output to `acf.hs` in `acquisition_sensitivity_from_attenuation.py`
* removed skipping listmode scripts by `run_all.py`
* sorted demos in `run_all.py` (pending more general solution to the execution order problem)
* corrected naming mismatch for some demos output files used as input for some other demos

## Contribution Notes

Please tick the following: 

### DIFF
--- a/examples/Python/PET/STIR_acquisition_model_using_raytracing.py
+++ b/examples/Python/PET/STIR_acquisition_model_using_raytracing.py
@@ -1,7 +1,10 @@
 '''Demo for setting advanced parameters in sirf.STIR.AcquisitionModelUsingRayTracingMatrix
 
 Usage:
-  STIR_acquisition_model_using_raytracing.py
+  STIR_acquisition_model_using_raytracing [options]
+
+Options:
+  --non-interactive           do not show plots
 
 This just sets some parameters and prints settings. Check other demos on how to use it.
 '''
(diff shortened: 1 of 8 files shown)


KrisThielemans (approved):
I haven't run this, but all looks sensible. Feel free to merge

evgueni-ovtchinnikov:
I had to change the scanning time interval in `reconstruct_from_listmode.py` from `(0,50)` to `(0,100)`, otherwise `scatter_estimation.py` complained that it expected `(0,100)` but got `(0,50)` - could not figure out how the former demo affects the latter (the only thing I can see is that the latter is executed after the former, but it does not help me much).

If this is ok, I will merge.

KrisThielemans:
This is ok. I guess we were trying to save some time in not repeating stuff. Indeed, if we calculate/plot scatter from data with a different duration, things wouldn't be good, so it checks for that.

Would be useful to add a comment in the relevant demo saying that they need to run the other first (it might be there already of course), and that interval needs to be the same.

then go ahead

evgueni-ovtchinnikov:
> Would be useful to add a comment in the relevant demo saying that they need to run the other first (it might be there already of course)

Yes it is there in `scatter_estimation.py`, so merging.

PR #1176, opened 2023-02-15
Re-implementing Python DataContainer methods in C++

### LINKED ISSUE #283: Implement DataContainer methods in C++
As documented in the [UserGuide](https://github.com/CCPPETMR/SIRF/blob/add_to_sirf_classes/doc/UserGuide.md#datacontainer) and being implemented in *Python only* in this [PR]
(https://github.com/CCPPETMR/SIRF/pull/237#issuecomment-449058718)

This issue is to see whether we can implement this in C++.

`sirf.DataContainer` has the method `copy` as an alias to `clone`. 
A number of methods are currently implemented on CCPi and not on SIRF `DataContainers`:
1. (Pixelwise) binary operations:
    1. `add(self, other , out=None, *args, **kwargs)`
    1. `subtract(self, other, out=None , *args, **kwargs):`
    1. `multiply(self, other , out=None, *args, **kwargs)` present with different signature `multiply(self, other)`
    1. `divide(self, other , out=None ,*args, **kwargs)` present with different signature `divide(self, other)`
    1. `power(self, other , out=None ,*args, **kwargs)`
    1. `maximum(self, other , out=None ,*args, **kwargs)`
1. all inline algebra is missing, achievable by using the above methods.
1. (Pixelwise) unary operations:
    1. `abs(self, out=None, *args,  **kwargs)`
    1. `sign(self, out=None, *args,  **kwargs)`
    1. `sqrt(self, out=None, *args,  **kwargs)`
1. reductions
    1. `sum(out=None, *args,  **kwargs)`

evgueni-ovtchinnikov:

A large number of `DataContainer` methods are currently implemented in Python with numerous duplications of data by `as_array`/`fill` copies. This PR aims at reducing unnecessary copying by moving as much computation as possible to C++.

## Related issues

Closes #283

## Contribution Notes

Please tick the following: 


src/common/csirf.cpp
```diff
@@ -297,6 +297,40 @@ cSIRF_ratio(const void* ptr_x, const void* ptr_y)
 	CATCH;
 }
 
+extern "C"
+void*
+cSIRF_maximum(const void* ptr_x, const void* ptr_y)
+{
+	try {
+		DataContainer& x =
+			objectFromHandle<DataContainer >(ptr_x);
```
KrisThielemans, line 306:
```suggestion
		auto& x = objectFromHandle<DataContainer >(ptr_x);
```
is less verbose maybe. In fact, at least for some of these you should probably do
```
		auto const& x = objectFromHandle<DataContainer >(ptr_x);
```

paskino:
@evgueni-ovtchinnikov I merged #1180 removing the Ubuntu 20.04 build. If you merge master into here you should get a green tick from the other builds on the GHA

evgueni-ovtchinnikov:
@KrisThielemans One of the builds keeps failing with

`/usr/bin/ld: /home/runner/install/lib/libIO.a(GEHDF5ListmodeInputFileFormat.cxx.o)`: in function `stir::GE::RDF_HDF5::GEHDF5ListmodeInputFileFormat::read_from_file(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const`:
`GEHDF5ListmodeInputFileFormat.cxx:(.text+0x83)`: undefined reference to `stir::GE::RDF_HDF5::CListModeDataGEHDF5::CListModeDataGEHDF5(std::__cxx11::basic_string<char, std::char_traits<char>`, `std::allocator<char> > const&`)

Looks like STIR I/O issue, does it not?

KrisThielemans:
> @KrisThielemans One of the builds keeps failing with
> 
> <snip>
> Looks like STIR I/O issue, does it not?

yes, unfortunately. This seems to have been there quite a while. It's the `DEVEL_BUILD=on` version, so using STIR master. Not sure why it's there, especially as STIR itself is building fine. I cannot fix this now, so I suppose we have to ignore this here for the moment. Apologies.

src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -536,7 +536,7 @@ float NiftiImageData<dataType>::get_norm(const NiftiImageData<dataType>& other)
         if (!std::isnan(this->operator()(i)+other(i)))
             result += double(pow( this->operator()(i) - other(i), 2));
```
KrisThielemans, line 537: Possibly unrelated to this PR, but can `dataType` be `std::complex<>`? Maybe needs `std::abs()`^2 instead.

src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -1742,11 +1742,34 @@ void NiftiImageData<dataType>::dot(const DataContainer& a_x, void* ptr) const
 {
     const NiftiImageData<dataType>& x = dynamic_cast<const NiftiImageData<dataType>&>(a_x);
     ASSERT(_nifti_image->nvox == x._nifti_image->nvox, "dot operands size mismatch");
-    double s = 0.0;
-    for (unsigned i=0; i<this->_nifti_image->nvox; ++i)
-        s += double(_data[i] * x._data[i]);
+    float s = 0.0;
```
KrisThielemans, line 1745: not sure why you make this `float` now. Will lead to loss of precision.

Also wouldn't handle complex data type (but `dot` then probably needs a conjugate anyway, so I guess it might not be used then?)

src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -1742,11 +1742,34 @@ void NiftiImageData<dataType>::dot(const DataContainer& a_x, void* ptr) const
...
+    for (unsigned i = 0; i < this->_nifti_image->nvox; ++i)
+        s += _data[i] * x._data[i];
+    float* ptr_s = static_cast<float*>(ptr);
```
KrisThielemans, line 1748: I'm not sure, but this seems like a very dangerous cast to me. Are we sure it will always be a `float *`? If so, why not pass as a `float *`? Or let it return `float`. This would need a change in the base class of course, so not sure if this is feasible now

src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -1742,11 +1742,34 @@ void NiftiImageData<dataType>::dot(const DataContainer& a_x, void* ptr) const
...
+    *ptr_s = s;
+}
+
+template<class dataType>
+void NiftiImageData<dataType>::sum(void* ptr) const
+{
+    float s = 0.0;
```
KrisThielemans, line 1755: again `float`. Far better to use `double` (if you don't need `complex`)

src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -1742,11 +1742,34 @@ void NiftiImageData<dataType>::dot(const DataContainer& a_x, void* ptr) const
...
+    for (unsigned i = 0; i < this->_nifti_image->nvox; ++i)
+        s += _data[i];
+    float* ptr_s = static_cast<float*>(ptr);
+    *ptr_s = s;
+}
+
+template<class dataType>
+void NiftiImageData<dataType>::max(void* ptr) const
+{
+    float s = 0.0;
```
KrisThielemans, line 1765: here `float` doesn't imply rounding, but `dataType` should work. Of course, it then implies that `ptr` is  a `dataType*`

src/Registration/pReg/tests/tests_data_container_algebra.py
```diff
@@ -0,0 +1,42 @@
+# -*- coding: utf-8 -*-
+"""sirf.STIR algebra tests
+v{version}
+
+Usage:
+  tests_six [--help | options]
```
KrisThielemans, line 6:
```suggestion
  tests_data_container_algebra [--help | options]
```

src/common/include/sirf/common/DataContainer.h
```diff
@@ -52,21 +53,55 @@ namespace sirf {
 		/// calculates the dot product of this container with another one
 		virtual void dot(const DataContainer& dc, void* ptr) const = 0;
 
+		/// calculates the sum of this container elements
```
KrisThielemans, line 56:
```suggestion
		/// calculates the sum of all container elements
```

src/common/include/sirf/common/DataContainer.h
```diff
@@ -52,21 +53,55 @@ namespace sirf {
 		/// calculates the dot product of this container with another one
```
KrisThielemans, line 53: will have to document `ptr` needs to be a `float *` (but I really don't like it...)

KrisThielemans (changes requested):
some comments. I didn't go through the whole thing.

src/Registration/cReg/NiftiImageData.cpp
evgueni-ovtchinnikov: Please see my previous reply.

src/Registration/cReg/NiftiImageData.cpp
evgueni-ovtchinnikov: Let me try to explain again in more detail.

I needed C interface to abstract `DataContainer` class. The problem is, it would not know the type of the data it works with, and C does not have templates. My solution to this problem is as follows. Firstly, I assume that data is either `float` or `comples_float_t` (i.e. `std::complex<float>`). Whenever a value needs to be computed (e.g. `dot` or `sum`), a zero complex float value is allocated, and the address of that complex value is passed to the interface function as `void*`. If the computed value is actually real, it will occupy the first 4 bytes of the allocated complex value. The Python interface, having received a complex value, checks the imaginary part and returns real value if it is zero and complex value otherwise.

src/Registration/pReg/tests/tests_data_container_algebra.py
evgueni-ovtchinnikov: thanks for spotting, corrected

src/Registration/cReg/NiftiImageData.cpp
KrisThielemans: sure, but that's a C interface problem. I'm talking about the C++ interface. (such things as (unsafe) casting should be handled in the C interface in my opinion.

evgueni-ovtchinnikov: C interface does not know the data type.

src/common/include/sirf/common/DataContainer.h
evgueni-ovtchinnikov: no, it can be also complex float

src/Registration/cReg/NiftiImageData.cpp
KrisThielemans: I'm a bit lost. sorry. Where in C++ do you actually handle complex in these functions?

KrisThielemans: ok. I also see that the C++ `DataContainer` is not templated in `DataType`, so it cannot have virtual functions that depend on `DataType`. A common problem with base classes, so this needs a longer term solution, but let's go ahead as-is now.

Possible longer term solutions:
- template `DataContainer` if possible
- put a class in between `RealDataContainer` and `ComplexDataContainer`, which have the intuitive signatures, as well as the generic ones, which do all the ugly casting, such that the derived classes don't need to.

Can you create an issue about this with some comments?

evgueni-ovtchinnikov:
I have re-designed `DataContainer` methods `dot`, `sum` and `max`. Here is how it looks for `sum` now:

`SIRF.py`:
```
    def sum(self):
        z = numpy.zeros((2,), dtype=numpy.float32)
        try_calling(pysirf.cSIRF_compute_sum(self.handle, z.ctypes.data))
        if z[1] == 0:
            return z[0]
        return z[0] + 1j*z[1]
```
`csirf.cpp`:
```
void*
cSIRF_compute_sum(const void* ptr_x, void* ptr_z)
{
	try {
		auto const& x = objectFromHandle<DataContainer>(ptr_x);
		x.sum(ptr_z);
		return new DataHandle;
	}
	CATCH;
}
```
(so, no complex variables anywhere in this interface any more)

`gadgetron_data_containers.cpp`:
```
void
MRAcquisitionData::sum(void* ptr) const
{
    int n = number();
    complex_float_t z = 0;
    ISMRMRD::Acquisition a;
    for (int i = 0; i < n;) {
        if (!get_acquisition(i, a)) {
            i++;
            continue;
        }
        z += MRAcquisitionData::sum(a);
        i++;
    }
    complex_float_t* ptr_z = static_cast<complex_float_t*>(ptr);
    *ptr_z = z;
}
```
(with the same `MRAcquisitionData::sum(const ISMRMRD::Acquisition& acq_a)` as before)

`stir_data_containers.cpp`:
```
void
STIRAcquisitionData::sum(void* ptr) const
{
	int n = get_max_segment_num();
	double t = 0;
	for (int s = 0; s <= n; ++s)
	{
		SegmentBySinogram<float> seg = get_segment_by_sinogram(s);
		SegmentBySinogram<float>::full_iterator seg_iter;
		for (seg_iter = seg.begin_all(); seg_iter != seg.end_all();)
			t += *seg_iter++;
		if (s != 0) {
			seg = get_segment_by_sinogram(-s);
			for (seg_iter = seg.begin_all(); seg_iter != seg.end_all();)
				t += *seg_iter++;
		}
	}
	float* ptr_t = static_cast<float*>(ptr);
	*ptr_t = (float)t;
}
```
If this is ok, then I believe I am done with this PR.

src/common/include/sirf/common/DataContainer.h
evgueni-ovtchinnikov: Added a comment on the actual values of `void*` in `DataContainer` methods.

PR #1166, opened 2023-01-04
Fixing issues raised in #1163

evgueni-ovtchinnikov:
## Checklist before requesting a review

- [ ] I have performed a self-review of my code
- [ ] I have added docstrings/doxygen in line with the guidance in the developer guide
- [ ] I have implemented unit tests that cover any new or modified functionality
- [ ] The code builds and runs on my machine
- [ ] CHANGES.md has been updated with any functionality change

## Contribution Notes

Please read and adhere to the [contribution guidelines](https://github.com/SyneRBI/SIRF/blob/master/CONTRIBUTING.md).

Please tick the following: 

 - [ ] The content of this Pull Request (the Contribution) is intentionally submitted for inclusion in SIRF (the Work) under the terms and conditions of the [Apache-2.0](https://www.apache.org/licenses/LICENSE-2.0) License.

PR #1194, opened 2023-05-11
Synergistic IO fix

### LINKED ISSUE #622: compilation failure with DISABLE_GADGETRON=ON due to syn_utilities
syn_utilities.cxx fails to compile in that case as it doesn't know about  include gadgetron_data_containers.h etc
https://github.com/SyneRBI/SIRF/blob/685469981611d1beab84bc98df409e706e9e4300/src/common/syn_utilities.cpp#L32-L35

I'm a bit confused about all this. 
- If it's synergistic, then it should sit in `Synergistic`.
- In fact, this file include `Syn/utilities.h`, which defines the class `ImageDataWrap` but
   - a class declaration should ideally sit in a file with the name of the class.
   - the class implementation and declaration should be in the same folder structure
- seems that we need to have `ImageWrap` existing all the time (so in `common` I guess), but use preprocessor defines to exclude lines in `syn_utilities.cpp` that won't compile if an engine is not available.
-  The `CMakeListst.txt` seems to imply that `syn_utilities.ccp` should only be compiled when all engines are present, which seems to be the opposite to what was intended
https://github.com/SyneRBI/SIRF/blob/685469981611d1beab84bc98df409e706e9e4300/src/common/CMakeLists.txt#L30-L45

KrisThielemans:
Fixes #622  (I hope)

Currently sits on top of #1166, sorry. but I'll rebase once that's merged to `master`. i first wanted to see if this would work.


KrisThielemans:
now rebased on top of master

KrisThielemans:
When using STIR release, we get
```
[ 30%] Building CXX object src/common/CMakeFiles/_pysirf.dir/home/runner/install/share/stir/src/buildblock_registries.cxx.o
/home/runner/install/share/stir/src/buildblock_registries.cxx:23:10: fatal error: stir/SeparableCartesianMetzImageFilter.h: No such file or directory
   23 | #include "stir/SeparableCartesianMetzImageFilter.h"
```
while with STIR `master`
```
ImportError: /home/runner/install/python/sirf/_pysirf.so: undefined symbol: _ZTIN4stir8ExamDataE
```
which might just be the constructor

KrisThielemans:
Hopefully the above problem is fixed now (it is on my machine with STIR master).

However, I still have a Python problem such as
```
ImportError: /home/sirfuser/devel/install/python/sirf/_pyreg.so: undefined symbol: deleteTextPrinter
```
@evgueni-ovtchinnikov `deleteTextPrinter` is declared in [cReg.h](https://github.com/KrisThielemans/SIRF/blob/synergistic_IO_fix/src/Registration/cReg/include/sirf/Reg/cReg.h#L148), but it is not implemented in `cReg*cpp`. Is it using the implementation in [cstir_tw.cpp](https://github.com/KrisThielemans/SIRF/blob/synergistic_IO_fix/src/xSTIR/cSTIR/cstir_tw.cpp#L89)? If so, `creg` depends on `cstir`, which I then should make explicit in the CMake file.

evgueni-ovtchinnikov:
> Is it using the implementation in [cstir_tw.cpp](https://github.com/KrisThielemans/SIRF/blob/synergistic_IO_fix/src/xSTIR/cSTIR/cstir_tw.cpp#L89)?

Apparently it is - when I do "Go To Definition" on the prototype in `cReg.h`, Visual Studio shows me the one in `cstir_tw.cpp`.

KrisThielemans:
@evgueni-ovtchinnikov this is pretty much finished now. I've tried to give descriptive log messages, so maybe that helps in the review. Most important change is that `csirf` is now split into `csirf` (C++ lib) and `cinterface-sirf` (for Python/MATLAB), and same for `cReg`. I did this for 2 reasons:
- it's cleaner
- it also avoids the circular dependency of `csirf` (actually `p/msirf`) needing `cSyn` (for `ImageDataWrapper`), which needs `csirf`.

There are a few things that we could still do here:
- do the same split `cinterface-stir` and `cinterface-gadgetron`
- move `iutilities.*`  (containing C functions for `DataHandle`) to `common` (and add it to `cinterface-sirf` I guess). This is more logical now that we have `common`. We can then remove the `iutilities` library (also in Python), but  I suppose we'd have to keep it as an empty library until 4.0.
- rename `csirf` (again) to `common`. It is the more logical choice.

These things would be nice, but we could also do them later. Would be cleanest to do the first bullet though (and it's very simple, so I can still do that).

evgueni-ovtchinnikov:
Main comment: I very much welcome suggested changes (thought about them myself, but never attempted for lack of understanding of CMake intricacies), but they look too dramatic to me for 3.5 - should this not go to 4.0?

I need more time to study your PR, for now just one minor comment that cannot wait: frankly, I find `cinterface-sirf` rather ugly. I would rather use `csirf` for C interface library and `cppsirf` for C++ library.

PR #1195, opened 2023-05-26
Implemented safer C++ DataContainer data type handling, fixes #1193

### LINKED ISSUE #1193: Longer term solutions to DataContainer data type issue needed.
Some `DataContainer` methods compute a value that is complex for Gadgetron containers and real for other containers, and so, the variable that would store such a value cannot be explicitly declared in this abstract base class. Currently, this issue is quick-fixed by passing a `void*` to these methods that is cast to `float*` or `complex_float_t*` in the respective methods of the derived classes.

Possible longer term fixes:
- template `DataContainer` if possible.

evgueni-ovtchinnikov:

Current C++ `DataContainer` classes use static casts of `void*` pointers, which may be unsafe (cf. #1193).

This PR implements an alternative template-based design whereby static casts are only used on C interface level, which is not supposed to be used by SIRF users. 

## Related issues

Fixes #1193

## Contribution Notes

Please tick the following: 


src/Registration/cReg/NiftyResampler.cpp
```diff
@@ -240,7 +240,7 @@ static void set_post_resample_outputs(std::shared_ptr<DataContainer> &out_to_ret
         NumberType::Type output_num_type = (*output_to_return_sptr->begin()).get_typeID();
         if (output_num_type != NumberType::CXFLOAT)
             throw std::runtime_error("NiftyResampler: Only complex type currently supported is complex float");
-        ImageData<dataType> ::Iterator &it_out = output_to_return_sptr->begin();
+        typename ImageData<dataType> ::Iterator &it_out = output_to_return_sptr->begin();
```
KrisThielemans, line 243: much simpler to say
```suggestion
        auto &it_out = output_to_return_sptr->begin();
```
but are you sure it needs to be a reference? It's going to change, and therefore affect `output_to_return_sptr->begin()`, which seems terribly dangerous. (This is actually also true for `it_real` etc)

evgueni-ovtchinnikov: I cannot be sure about anything here - this is Richard's code, I simply had to add `typename` for SIRF to build (now it builds but synergistic tests fail).

Is `auto` clever enough to use `typename`?

src/common/include/sirf/common/DataContainer.h
```diff
@@ -37,11 +37,239 @@ which rely on the same features of the items.
...
+
+		/// \c *this = elementwise sum of elementwise \c x*a and \c y*b
+		void xapyb(
+			const DataContainer& a_x, const DataContainer& a_a,
+			const DataContainer& a_y, const void* ptr_b)
+		{
+			xapyb(a_y, ptr_b, a_x, a_a);
+		}
+
+		static T product(T x, T y)
```
KrisThielemans, line 196: I think that all these `static` function have to be `protected` (or even `private`). they're not part of the public interface really

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
```diff
@@ -1571,31 +1577,31 @@ GadgetronImageData::multiply(const DataContainer& a_x, const DataContainer& a_y)
 {
 	SIRF_DYNAMIC_CAST(const GadgetronImageData, x, a_x);
 	SIRF_DYNAMIC_CAST(const GadgetronImageData, y, a_y);
-    binary_op(x, y, DataContainer::product<complex_float_t>);
+    binary_op(x, y, DataContainerTempl<complex_float_t>::product);
```
KrisThielemans, line 1580: I think all these would be a lot cleaner if we'd have
```
typedef GadgetronImageData self_type;
```
as then we can do
```
    SIRF_DYNAMIC_CAST(const  self_type, x, a_x);
    SIRF_DYNAMIC_CAST(const self_type, y, a_y);
    binary_op(x, y, self_type::product);
```
such that the code looks class-independent

KrisThielemans (commented):
some probably mostly random comments but hopefully helpful in this review.

I'm not sure if templating `ImageData` is a good idea. It certainly breaks backwards compatibility in a huge way. Possibly this could be done by having an intermediate `ImageDataTemplate` like you did for DataContainer`. However, it might not work, as I guess we'd then need to derive `ImageData` from `DataContainer`, but `ImageDataTemplate` from `DataContainerTemplate`. This can be done in C++, but has the potential for some confusion.

src/Registration/cReg/NiftyResampler.cpp
evgueni-ovtchinnikov: virtual method's return types must be covariant, must they not?

src/Registration/cReg/cReg_p.cpp
```diff
@@ -115,15 +115,15 @@ sirf::cReg_NiftiImageDataParameter(const DataHandle* handle, const char* name)
 void*
 sirf::cReg_setRegistrationParameter(void* hp, const char* name, const void* hv)
 {
-    std::shared_ptr<const ImageData> im_sptr;
+    std::shared_ptr<const ImageData<float> > im_sptr;
 
     Registration<float>& s = objectFromHandle<Registration<float> >(hp);
     if (strcmp(name, "reference_image") == 0) {
-        getObjectSptrFromHandle<const ImageData>(hv, im_sptr);
+        getObjectSptrFromHandle<const ImageData<float> >(hv, im_sptr);
```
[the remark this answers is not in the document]
evgueni-ovtchinnikov: My understanding is that in his C interface Richard has `float` for `dataType`.

src/Registration/cReg/NiftyResampler.cpp
KrisThielemans: not sure what you mean. Virtual methods return types have indeed to be either identical or returning a derived version of something related to the base class (I haven't checked exact definition of "covariant return types", see maybe https://stackoverflow.com/questions/25813435/what-exactly-are-covariant-return-types-in-c).

However, that doesn't really apply here as far as I can see. I was merely suggestion to say
```c++
auto it_out = output_to_return_sptr->begin();
```
(if that compiles anyway)

evgueni-ovtchinnikov: AFAIK covariant returns must be either pointers or references

KrisThielemans: agreed. But as it's a reference, it might be possible to just say `auto`. Not sure...

evgueni-ovtchinnikov: from the link you quoted: "Covariance only happens when you are returning a pointer or a reference to a class, and the classes are related by inheritance."

src/Registration/cReg/NiftyResampler.cpp
```diff
@@ -42,18 +42,26 @@ using namespace sirf;
 using namespace detail;
 
 template<class dataType>
-static void convert_ImageData_to_ComplexNiftiImageData(ComplexNiftiImageData<dataType> &output, const std::shared_ptr<const ImageData> input_sptr)
+static void convert_ImageData_to_ComplexNiftiImageData(ComplexNiftiImageData<dataType> &output, const std::shared_ptr<const DataContainer> input_sptr)
```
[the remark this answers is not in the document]
evgueni-ovtchinnikov: Here we need a base class for both real and complex `ImageData` - `DataContainer` is the only one we have. Not pretty, but I have not found an alternative.

Regarding shared pointers - it looks like the use of them is not justified in this and likely many other places, but I suggest we create a separate issue on that.

src/Registration/cReg/cReg_p.cpp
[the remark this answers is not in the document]
KrisThielemans: ok. that's what the original code was saying anyway, so I'll resolve this

src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp
evgueni-ovtchinnikov: @KrisThielemans works nicely, thanks! Will apply to all container methods after I got rid of all `void*` (operations involving scalars still have them).

src/Registration/cReg/NiftyResampler.cpp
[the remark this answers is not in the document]
KrisThielemans: > Here we need a base class for both real and complex `ImageData` - `DataContainer` is the only one we have. Not pretty, but I have not found an alternative.
> 

question is why you need the base class. For instance, as `Resampler` is templated in `dataType`, things like
```
 virtual void set_reference_image(const std::shared_ptr<const DataContainer > reference_image_sptr);

    /// Set floating image. This is the image that would be the floating if you were doing a forward transformation.
    virtual void set_floating_image(const std::shared_ptr<const DataContainer > floating_image_sptr);
```
could all have `shared_ptr<ImageData<dataType>>`. Maybe that helps?

This static function could then easily be templated as well.

> Regarding shared pointers - it looks like the use of them is not justified in this and likely many other places, but I suggest we create a separate issue on that.

sure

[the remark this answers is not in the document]
KrisThielemans: in discussions, @evgueni-ovtchinnikov  explained that currently `dataType` is always `float`. Complex images are handled independent. So we need `DataContainer` here at the moment.

evgueni-ovtchinnikov:
Left unattended for a long time, resulting in weird conflicts with master. Taken over by PR #1210.

PR #1204, opened 2023-07-20
add OSSPS set_relaxation_gamma

### LINKED ISSUE #1203: OSSPSReconstructor should expose set_relaxation_gamma, and have get_* functions
We have `relaxation_parameter` but not `relaxation_gamma`. Also, In Python, there are only `set` functions, but no `get` (this is actually the case in `OSMAPOSL` as well, e.g. for `maximum_relative_change`, but that probably needs a separate issue)

KrisThielemans:
Fixes #1203.

@evgueni-ovtchinnikov I made a start with this. Would you be able to finish it off, and add a quick test?

### DIFF
--- a/doc/UserGuide.md
+++ b/doc/UserGuide.md
@@ -520,6 +520,7 @@ Class for reconstructor objects using Ordered Subsets Separable Paraboloidal Sur
 
     OSSPSReconstructor       Constructor. Creates new OSSPS reconstructor object.
     set_relaxation_parameter Sets relaxation parameter.
+    set_relaxation_gamma Sets relaxation gamma parameter.
 
 ##### FullySampledReconstructor (MR) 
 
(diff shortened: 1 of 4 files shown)


evgueni-ovtchinnikov:
testing: what value of `gamma` would be realistic and what effect to expect?

KrisThielemans:
https://stir.sourceforge.net/documentation/doxy/html/classstir_1_1OSSPSReconstruction.html says alpha~1 and and gamma ~ .1. Those are the defaults in STIR, so I guess SIRF. Higher gamma means slower convergence (as step-size goes down), but more subset fluctuations.

In any case, I think the test could simply be to `set` and then `get`.

Bu the way, there's an `upper_bound` that would also be useful to expose.

KrisThielemans (commented):
need to update CHANGES.md

KrisThielemans (commented):
looks good to me. ready to merge I guess!

PR #1207, opened 2023-07-21
A more flexible alternative to TO_BE_IGNORED macro

### LINKED ISSUE #1200: Reading in MR data with _all=True
When [reading in an MR raw data file](https://github.com/SyneRBI/SIRF/blob/a31f2d23c76ad43b36611a81e3b3e9f66ba296f6/src/xGadgetron/pGadgetron/Gadgetron.py#L847) using `_all=True` and then calling `dimensions()` or `number()` then different results are returned:
`number()` returns the number of all readouts, also noise calibration data
`dimensions()` returns only the readouts which are selected by [get_acquisition()](https://github.com/SyneRBI/SIRF/blob/a31f2d23c76ad43b36611a81e3b3e9f66ba296f6/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h#L731)

In addition
`get_ISMRMRD_info('flags', range(AcquisitionData.number()))` and `get_ISMRMRD_info('flags')` lead to different results.

I understand that the definitions of which readouts to use based on the flags is necessary to be able to sort the data into a numpy array. Nevertheless, I find this behaviour confusing and might lead to many problems further down the processing pipeline. 

I am not sure yet how best to fix this, probably requires further discussions...

evgueni-ovtchinnikov:

Introduced new class `IgnoreMask` that allows the users to set their own acquisition selection rules.

## Related issues

Fixes #1200.

- [x] I have added docstrings/doxygen in line with the guidance in the developer guide

## Contribution Notes

Please tick the following: 


src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -514,8 +566,11 @@ namespace sirf {
 
 		virtual gadgetron::shared_ptr<ISMRMRD::Acquisition>
 			get_acquisition_sptr(unsigned int num) = 0;
-		virtual int get_acquisition(unsigned int num, ISMRMRD::Acquisition& acq) const = 0;
-		virtual void set_acquisition(unsigned int num, ISMRMRD::Acquisition& acq) = 0;
+		virtual int get_acquisition(unsigned int,
```
ckolbPTB, line 569: in `get_data` and `set_data` you pass `ignore_mask` to `get_acquisitions` but not anywhere else. Should it not be done everywhere where `get_acquisitions` is called? 

An alternative could be to change the code such that the ignore mask is used when reading in the data from file but afterwards get_acquisition always returns everything in AcquisitionsData and the flags are disregarded. If the users wants a subset for a particular set of flags they can use `get_subset`.

KrisThielemans: I like the subset idea...

evgueni-ovtchinnikov: > in get_data and set_data you pass ignore_mask to get_acquisitions but not anywhere else. Should it not be done everywhere where get_acquisitions is called?

I would rather not have it in algebraic methods lest MR algebra would require separate interface as we do not use masks on PET and Reg sides.

> An alternative could be to change the code such that the ignore mask is used when reading in the data from file but afterwards get_acquisition always returns everything in AcquisitionsData and the flags are disregarded.

Some acquisition data include noise calibration data, which we do not use at present but might need later, but it should be ignored by algebraic operations.

> If the users wants a subset for a particular set of flags they can use `get_subset`.

In the presence of calibration data, the number of acquisitions in the subset of data needed for reconstruction might be just 1 acquisition less than the whole data, so creating a subset with `get_subset` would essentially duplicate a huge amount of data.

ckolbPTB: The problem I have is that I have data with say 1000 acquisitions. 

If I use `pMR.AcquisitionData(praw+fdat, False, ignored = pMR.IgnoreMask(19))` then they are all read in correctly.

If I then call `print('Py dimensions ', acq_data.dimensions())` I get something like (500, 6, 128). I.e. 500 acquisitions are missed. This is because when `dimensions()` is called the algorithm goes through all acquisitions with `get_acquisition(i, acq)` to find out if the acquisition is valid. `get_acquisition` uses the default IgnoreMask() which is different to `IgnoreMask(19)` and hence ignores 500 acquisitions.

evgueni-ovtchinnikov: @ckolbPTB it looks like you wanted to ignore acquisitions with flag `ISMRMRD_ACQ_IS_NOISE_MEASUREMENT = 19`, did you not? If so, then I have to apologise for not documenting `IgnoreMask` object properly, relying instead on self-explanatory names of variables and methods.

To ignore flag 19 one needs two steps:
```
ignored = IgnoreMask(0) # nothing to ignore - all 64 bits are zeros
ignored.ignore(19) # ignore acquisitions with flag ISMRMRD_ACQ_IS_NOISE_MEASUREMENT (nonzero bit 19)
```

ckolbPTB: I changed my code based on your suggestions but it still does not work. I still only get 500 acquisitions instead of 1000.

To me I always see a problem if a different mask is used when reading in the data compared to calling `get_acquisition(i, acq)` later in the code. I understand the problem with the noise measurements because they need to be ignored for any image reconstruction process later, but there are lots of other acquisition flags which the user would want to have in the reconstruction pipeline and they are not covered by the default `IgnoreMask` used in  `get_acquisition(i, acq)`. I think we need to discuss this in more detail in the next software meeting.

evgueni-ovtchinnikov: It looks like my suggestion was a bit too laconic, I should have added the third line:
```
ignored = IgnoreMask(0) # nothing to ignore - all 64 bits are zeros
ignored.ignore(19) # ignore acquisitions with flag ISMRMRD_ACQ_IS_NOISE_MEASUREMENT (nonzero bit 19)
pMR.AcquisitionData(praw+fdat, False, ignored=ignored) # use your handmade mask to ignore noise measurement acquisition only
```
or perhaps just one line `pMR.AcquisitionData(praw+fdat, False, ignored=pMR.IgnoreMask(0).ignore(19))` would work too.

What I mislooked though was that all this would have no effect on the dimensions because at present `AcquisitionData.dimensions` does not get ignored mask - I need to add it to the argument list (and in a few other places too). 

Furthermore, I started to suspect that I might have misunderstood the whole idea of these flags. The only flag that may cause ignoring an acquisition that I understand is  `ISMRMRD_ACQ_IS_NOISE_MEASUREMENT`. It occurred to me just now that e.g. flags  `ISMRMRD_ACQ_USER1` etc. that in my current design would cause ignoring the acquisition actually should not have such effect, serving some other purposes.

evgueni-ovtchinnikov: @ckolbPTB as agreed on last Thu tcon, I changed the default value of `IgnoreMask.mask` to `1 << 18` - please check if this solves the problems you reported, so I can merge this PR.

ckolbPTB: It does work now for Paul's data but I don't think this solution will work for data in general. The main problem I see is that we are using different IgnoreMasks throughout the code. 

When reading AcquisitionData from file we can specify an IgnoreMask. That is fine. 
Later in the code when e.g. get_acquisition_dimension() is called, get_acquisition() is called (https://github.com/SyneRBI/SIRF/blob/b6bdc997df50978807175648c2145d2141810d9e/src/xGadgetron/cGadgetron/gadgetron_data_containers.cpp#L192C14-L192C29) without specifying an IgnoreMask. This means get_acquisitions() uses the default IgnoreMask. If the user has specified a specific IgnoreMask when reading in the data (e.g. ignoring also some navigator data in addition to the nosie data) this is not used here and the code will probably crash. 

The only way I can see is to:
- make IgnoreMask an attribute of the AcquisitionData class
- allow the user to modify IgnoreMask to adapt it to the current steps in the reconstruction pipeline
- use AcquisitionData.IgnoreMask in e.g. get_acquisition()

evgueni-ovtchinnikov: @ckolbPTB thanks, implemented

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -129,6 +115,67 @@ namespace sirf {
         mutable bool have_header_;
 	};
 
+	/*!
+	\ingroup Gadgetron Data Containers
+	\brief Class enabling ignoring certain 'irregular' acquisitions.
```
KrisThielemans, line 120: some more doc here

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -129,6 +115,67 @@ namespace sirf {
...
+		}
+		unsigned long int bits() const
+		{
+			return ignore_;
+		}
+		bool ignored(unsigned long int bits) const
+		{
+			return bits & ignore_;
+		}
+		void show_bits() const
```
KrisThielemans, line 159: I guess this is for debugging, but generally better to return a string. (can use `std::stringstream` to construct it), or pass a `ostream` as input

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
```diff
@@ -618,72 +669,75 @@ namespace sirf {
...
-    	*/
+		/*!
+		\brief Reader for ISMRMRD::Acquisition from ISMRMRD file.
+		* filename_ismrmrd_with_ext: filename of ISMRMRD rawdata file with .h5 extension.
+		* In case the ISMRMRD::Dataset constructor throws an std::runtime_error the reader catches it,
+		* displays the message and throws it again.
+		* To avoid reading noise samples and other calibration data, IgnoreMask may be employed
+		* to exclude potentially incompatible input.
+		*/
 		void read(const std::string& filename_ismrmrd_with_ext, int all = 0);
```
KrisThielemans, line 731: what is the `all` parameter? not documented

examples/Python/MR/noncartesian_cg_sense.py
```diff
@@ -137,7 +137,7 @@ def main():
 
     # acquisition data will be read from an HDF file input_file
     # AcquisitionData.set_storage_scheme('memory')
-    acq_data = mr.AcquisitionData(input_file)
+    acq_data = mr.AcquisitionData(input_file, False)
```
KrisThielemans, line 140: Is it now necessary to add the 2nd argument? what does it do?

KrisThielemans (changes requested):
some comments from someone who doesn't know too much of what's going on! (Obviously, don't forget to update CHANGES.md)

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
evgueni-ovtchinnikov: done

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
evgueni-ovtchinnikov: documented

src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h
evgueni-ovtchinnikov: added

KrisThielemans:
Gadgetron error seems to be
```
ERROR [Core.cpp:18] [Connection Main Thread/Connection Input Thread] ERROR: map::at
```
@ckolbPTB any idea?

evgueni-ovtchinnikov:
> Thanks Evgueni - looks fine to me. If you have some more energy could you add a get_ignore_mask() function for AcquisitionData on Python-level?

@ckolbPTB Done. Also provided setter for convolution kernel size (for smoothing out CSM).

KrisThielemans:
@evgueni-ovtchinnikov can you replace
```c++
unsigned long int* ptr_ignored = (unsigned long int*)ptr;
```
with
```c++
auto ptr_ignored = static_cast<unsigned long int*>(ptr);
```
to satisfy Codacy?

Also merge `master` to resolve conflict in `CHANGES.md` and expand it.

Best to do the C++ changes as a last commit in the push, such that CI will run (if there are only changes in .md files in the last commit, CI will not run currently).

evgueni-ovtchinnikov:
> @evgueni-ovtchinnikov can you replace 
> ```c++
> unsigned long int* ptr_ignored = (unsigned long int*)ptr;
> ```
> with
> ```c++
> auto ptr_ignored = static_cast<unsigned long int*>(ptr);
> ```
> to satisfy Codacy?

@KrisThielemans  Tried but got
```
 error: invalid ‘static_cast’ from type ‘size_t’ {aka ‘long unsigned int’} to type ‘long unsigned int*’

  572 |                 auto ptr_ignored = static_cast<unsigned long int*>(ptr);
```
so had to stick to C-style cast :(

KrisThielemans:
oops. `reinterpret_cast`. sorry

evgueni-ovtchinnikov:
> oops. `reinterpret_cast`. sorry

@KrisThielemans Thanks! (I forgot about `reinterpret_cast` too.)

KrisThielemans:
success!

ckolbPTB:
I tested it successfully on Paul's data and also tried the different smoothness and kernel sizes. All seems to work fine. One last thing: could you add some comments about what the parameters do for `class CoilSensitivityData(ImageData)`. E.g. I don't think it is obvious that the parameter `smoothness` actually means number of iterations. Otherwise fine to merge from my side.

KrisThielemans:
thanks @ckolbPTB. I'm assuming that the extra `False` argument to `mr.AcquisitionData` is still being sorted out. Also, I do think that the current line in `CHANGES.md` is not sufficient. (As a user, I'd like to know what improvements...). Also possibly point to an example.

great stuff!

evgueni-ovtchinnikov:
> success!

Not quite yet, I just found out why the GHA tests started failing all of a sudden, and why I could not reproduce these failures on my VM.

About a month ago I updated Gadgetron on another VM to commit `0670db847aca8df629b2005c1b23512dda84f175` (10 Mar 2023), and preprocessing failed on acquisition data with noise calibration acquisitions. Setting argument `all_` of the `AquisitionData` constructor to `False` effectively removed such acquisitions, and the tests succeeded. I blamed my updating VM via `SIRF-SuperBuild` rather than `update_VM.sh`, and soon forgot about this trouble, switching back to VMs with old Gadgetron.

What I actually found yesterday was that it was just the removal of `False` from the argument list of `mr.AcquisitionData` (which assigned the default value `True` to `all_`) that triggered tests failures. This suggests that GHA use some more recent version of Gadgetron than the VMs I normally use for running tests, which is why I could not reproduce the failures - until I tried the VM with new Gadgetron.

I also found that some tests succeed if `NoiseAdjustGadget` was removed, suggesting that this gadget may have a bug.  I need to reproduce failures by running Gadgetron client and create an issue on `github/gadgetron`. Until Gadgetron team finds time for fixing it, we will have to use `all_=False`, I am afraid.

KrisThielemans:
Thanks @evgueni-ovtchinnikov This goes beyond what I understand and have time for, but it puzzles me that you say that GHA "suddenly started failing".  Some comments in an attempt to be useful (but likely  fail!)
- Our Gadgetron version hasn't changed in a while now. I suppose the "sudden" was when you removed the `False` argument? (Possibly you could check in the history when this error started to appear.)
-  Current `master` doesn't fail

Would defaulting to `False` be an alternative? This could allow merging, and then sorting the actual problem out later? (I have no idea what the argument does, you see 😄 )

evgueni-ovtchinnikov:
> * Our Gadgetron version hasn't changed in a while now. I suppose the "sudden" was when you removed the `False` argument? (Possibly you could check in the history when this error started to appear.)

Yes it was, sorry for not being clear about it.

> Would defaulting to False be an alternative? This could allow merging, and then sorting the actual problem out later? (I have no idea what the argument does, you see 😄 )

The second argument `all_` forces all acquisitions in the input file to be used, ignoring the value of the third argument. Defaulting to `False` (as it is done on `master`) would do only as a temporary measure to avoid tests failure.

KrisThielemans:
> The second argument `all_` forces all acquisitions in the input file to be used, ignoring the value of the third argument.

Sorry, I don't quite understand. Using a 2nd `bool` argument that has as function to ignore the 3rd argument (mask) seems unnecessary to me. Either there is a 3rd argument (in which case it should be used) or there isn't. But it is likely not that simple, as the problem occurs when using `False` and no 3rd argument. At first sight, this makes no sense: there is no mask, so setting `all_=False` should make no difference. This seems to point to `ignore_mask_` not being initialised properly in some cases,

 For instance, I note that in `AcquisitionsVector::clone_impl` the mask is not copied. That seems wrong/dangerous. (By the way, as `ignore_mask_` is a member of `MRAcquisitionData`, it seems better to let `MRAcquisitionData` have a `clone_impl_` which has then to be called by the derived classes, such that they will automatically copy whatever is needed).

(You could also try and run through `valgrind` on the VM to check if all variables are initialised).

Looking a bit further at the code, I think `all_` should be removed. If a mask is set, it should be used always, as @ckolbPTB suggested. For instance, while [AcquisitionsVector::get_acquisition](https://github.com/SyneRBI/SIRF/blob/26d1211eefec409a4eeb1d9f129cd03feae37bc5/src/xGadgetron/cGadgetron/include/sirf/Gadgetron/gadgetron_data_containers.h#L799) always uses the mask, but the `read` function allows circumventing it, that can only lead to trouble.

evgueni-ovtchinnikov:
@KrisThielemans thanks for pointing at un-initialised `ignore_mask_` - now fixed (I hope).

I kept `all_` after introducing `IgnoreMask` just for backward compatibility, though the likelihood of somebody having a script that requires this argument is low and fixing is not difficult, so I will get rid of it soon.

KrisThielemans:
possibly it needs yet another merge with `master` before the actions will run.

PR #1208, opened 2023-08-21
fixes STIR TOF AcquisitionData algebra

KrisThielemans:
inserted TOF loops. However, this needs the STIR_TOF preprocessor variable to be defined, which is still to be merged on the STIR TOF branch.

### DIFF
--- a/src/xSTIR/cSTIR/stir_data_containers.cpp
+++ b/src/xSTIR/cSTIR/stir_data_containers.cpp
@@ -23,6 +23,8 @@ limitations under the License.
 #include "stir/KeyParser.h"
 #include "stir/is_null_ptr.h"
 #include "stir/zoom.h"
+#include "stir/CartesianCoordinate3D.h"
+#include "stir/numerics/norm.h"
 
 using namespace stir;
 using namespace sirf;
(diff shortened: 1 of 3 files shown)


src/xSTIR/cSTIR/stir_data_containers.cpp
```diff
@@ -32,20 +32,29 @@ using namespace sirf;
 std::string STIRAcquisitionData::_storage_scheme;
 std::shared_ptr<STIRAcquisitionData> STIRAcquisitionData::_template;
 
+#ifdef STIR_TOF
+#define TOF_LOOP  for (int k=data()->get_min_tof_pos_num(); k<=data()->get_max_tof_pos_num(); ++k)
+#define TOF_ARG , k
+#else
+#define TOF_LOOP
```
evgueni-ovtchinnikov, line 39: suggestion:
```
#define TOF_LOOP(k) for(int k=...)
#define TOF_ARG(k) , k
#else
#define TOF_LOOP(k)
#define TOF_ARG(k)
```

evgueni-ovtchinnikov (commented):
Will this be backward compatible? If not, should this not be merged after `data-cont-templ`?

KrisThielemans:
It should be once I'm done (had some stupid errors!). All changes should be within `#ifdef STIR_TOF`. If you don't build with it, everything is identical.

KrisThielemans:
@evgueni-ovtchinnikov currently some of the tests fail due to numerical rounding error. Printing stuff like
```python
 print('norm:', s, t, eps, abs(t-s), eps*abs(t))
```
after https://github.com/SyneRBI/SIRF/blob/1bb1b111a80f48df5e491a8ee9201715d4e9db42/src/common/Utilities.py#L600

I get 
```
+++ test 0 failed: expected 1, got False
norm: 47163.3984375 47161.4 1e-05 2.0 0.471613984375
+++ test 1 passed
+++ test 2 passed
+++ test 3 failed: expected 1, got False
dot: 68452680.0 68447064.0 1e-05 5616.0 684.47064
+++ test 4 passed
+++ test 5 passed
```
etc (2 more tests are failing, not sure yet which).

Your implementations look ok though, they use `double` for the temp variable (see e.g. [here](https://github.com/SyneRBI/SIRF/blob/1bb1b111a80f48df5e491a8ee9201715d4e9db42/src/xSTIR/cSTIR/stir_data_containers.cpp#L36)). I guess we could increase our tolerance a bit?

Aside from that, it's annoying that these tests don't give any feedback about failure and I had to add the print by hand. Easy to replace with check_if_equal_within_tolerance`?

KrisThielemans:
note that the above failure is on some TOF data (I haven't run on non-TOF yet, but Actions will tell us)

KrisThielemans:
@evgueni-ovtchinnikov that does indeed look better but it didn't make any practical difference. It could indeed be that the `numpu.linalg.norm`  isn't as smart as we are 😄 . In any case, the difference is well below our precision.

I've now
- modified the tests to
   - relax it for `norm` and `dot`
   - use `check_if_..._tolerance` for clearer output if they fail.
- changed `norm` to do what you suggest, but I also rewrote the loop, such that it's far shorter. Finally, I used `stir::norm_squared` to save 2 lines (sadly had to modify an unrelated line using `stir::norm` due to a STIR bug). This could be done for the others as well, but I have run out of time.

I'm done with this therefore and would like to merge. Can you have a look please?

evgueni-ovtchinnikov (approved):

KrisThielemans:
@nicolejurjew, this is now merged to SIRF master

PR #1210, opened 2023-08-25
Safer C++ DataContainer data type handling updated

### LINKED ISSUE #1193: Longer term solutions to DataContainer data type issue needed.

evgueni-ovtchinnikov:

This PR takes over from PR #1195, which was left unattended for a long time, resulting in weird conflicts with master.

## Testing performed

All tests run by `ctest` succeeded.

## Related issues

Fixes #1193.

## Contribution Notes

Please tick the following: 


src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -64,27 +64,64 @@ static void check_folder_exists_if_not_create(const std::string &path)
...
+}
+
+template<class dataType>
+NiftiImageData<dataType>& NiftiImageData<dataType>::operator=(const ImageData<float>& to_copy)
 {
     // Check for self-assignment
-    if (this != &to_copy) {
+    if (static_cast<void*>(this) != static_cast<const void*>(&to_copy)) {
+        // Try to cast to NiftiImageData.
+        const NiftiImageData<dataType> * const nii_ptr = dynamic_cast<const NiftiImageData<dataType> * const >(&to_copy);
```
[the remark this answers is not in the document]
evgueni-ovtchinnikov: no idea, this is Richard's stuff, and it looks like he wanted to cover for lots of data types - from `bool` to `long double` (see `NiftiImageData.h`)

at any rate, I doubt we would ever need any `dataType` other than `float`

to remind you: `dataType` cannot be complex - complex images are treated as pairs of real ones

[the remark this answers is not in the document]
KrisThielemans: Checking `copy_nifti_image` it contains https://github.com/SyneRBI/SIRF/blob/60bda8a52414a77172825594ddc78c484321122e/src/Registration/cReg/NiftiImageData.cpp#L683
therefore, it needs to have the same data-types, see these lines are correct.

Note that it can be simplified to
```
auto nii_ptr = dynamic_cast<const NiftiImageData<dataType> * const >(&to_copy)
```

src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -64,27 +64,64 @@ static void check_folder_exists_if_not_create(const std::string &path)
...
+                throw std::runtime_error("Trying to copy an uninitialised image.");
+            copy_nifti_image(_nifti_image,nii_ptr->_nifti_image);
+            this->_data = static_cast<float*>(_nifti_image->data);
+            this->_original_datatype = nii_ptr->_original_datatype;
+            set_up_geom_info();
+        }
+        else {
+            this->_nifti_image = NiftiImageData<float>::create_from_geom_info(*to_copy.get_geom_info_sptr());
+            // Always float
+            this->set_up_data(NIFTI_TYPE_FLOAT32);
```
KrisThielemans, line 108: this looks incorrect to me. `*this` should always have `dataType`, otherwise templating this makes no sense. So we need to set-up with the corresponding `NIFTI...`. (Easiest to do by having a helper class with template specialisations for supported `dataType`s).

src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -102,7 +139,11 @@ NiftiImageData<dataType>& NiftiImageData<dataType>::operator=(const ImageData& t
             // Always float
             this->set_up_data(NIFTI_TYPE_FLOAT32);
             // Finally, copy the data
-            this->copy(to_copy.begin(), this->begin(), this->end());
+            auto &it_src = to_copy.begin();
+            auto &it_dst = this->begin();
+            for (; it_src != to_copy.end(); ++it_src, ++it_dst)
+                *it_dst = (*it_src).complex_float().real();
```
KrisThielemans, line 145: this is truncating `complex` to `float`. 2 remarks:
- This would be amazingly confusing if `dataType` is complex. I know you're saying this can currently not happen, but do we guarantee it anywhere?
- If `dataType` is `float` or `double`, then this in my opinion shouldn't be in an assignment operator.

src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -1670,10 +1712,7 @@ dataType
 NiftiImageData<dataType>::
 get_inner_product(const NiftiImageData &other) const
 {
-	dataType s;
-	this->dot(other, &s);
-	return s;
-//    return std::inner_product(this->begin(),this->end(),other.begin(),dataType(0));
+    return (dataType)this->dot(other);
```
KrisThielemans, line 1715:
```suggestion
    return static_cast<dataType>(this->dot(other));
```
otherwise the brackets are rather confusing

KrisThielemans (changes requested):

src/Registration/cReg/NiftiImageData.cpp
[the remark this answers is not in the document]
evgueni-ovtchinnikov: > therefore, it needs to have the same data-types

so far as I can see, this is to be taken care of by `nifti_copy_nim_info`

> Note that it can be simplified

thanks, adopted

src/Registration/cReg/NiftiImageData.cpp
evgueni-ovtchinnikov: No, `dataType` cannot be complex - see 12 data types dealt with by `change_datatype`.

As the destination can only store `float` data while the source has `complex_float_t` data, some kind of truncation (either `abs` or `real`) is unavoidable I am afraid.

src/Registration/cReg/NiftiImageData.cpp
evgueni-ovtchinnikov: thanks, adopted

PR #1226, opened 2023-12-11
Fixed a bug in NiftiImageData.cpp causing segfault

### LINKED ISSUE #1225: NiftyResampler norm crashes on MR data with multiple repetitions
With https://github.com/SyneRBI/SIRF/pull/1186  `norm` on the `NiftyResampler` was introduced. While running the [MCIR reconstruction](https://github.com/paskino/SIRF-Contribs/blob/MR_MCIR_2023/src/notebooks/simulating_motion_2.ipynb) I call it and python crashes.

A minimal example to reproduce it is

```python

import sirf.Reg as pReg
import sirf.Gadgetron as pMR
import numpy as np
import os

from cil.utilities.display import show2D

from sirf.Utilities import  examples_data_path
acq_data = pMR.AcquisitionData(os.path.join
            (examples_data_path('MR'),'simulated_MR_2D_cartesian.h5')
        )
# ad1.sort()
preprocessed_data = pMR.preprocess_acquisition_data(acq_data)
recon = pMR.FullySampledReconstructor()
recon.set_input(preprocessed_data)
recon.process()
im_ms = recon.get_output()

max_shift = 20.


t_x=max_shift
t_y=0
tm = pReg.AffineTransformation(np.array(\
        [[1, 0, 0, t_x], \
            [0, 1, 0, t_y], \
            [0, 0, 1, 0  ], \
            [0, 0, 0, 1  ]]))


im_ms1 = im_ms * 1.1

# Create resampler
mf_resampler = pReg.NiftyResample()
# mf_resampler[ind] = NiftyResampleWithNormCache()
mf_resampler.set_reference_image(im_ms)
mf_resampler.set_floating_image(im_ms1)
mf_resampler.add_transformation(tm)
mf_resampler.set_padding_value(0)
mf_resampler.set_interpolation_type_to_linear()

mf_resampler.norm()
```

I am running this on SIRF from the https://github.com/SyneRBI/SIRF/tree/ignore-acq branch

evgueni-ovtchinnikov:

Fixed a bug in `NiftiImageData.cpp` that caused segfault.

## Testing performed

Tested on the code provided by @paskino in #1225.

## Related issues

Fixes #1225.

## Contribution Notes

Please tick the following: 

### DIFF
--- a/src/Registration/cReg/NiftiImageData.cpp
+++ b/src/Registration/cReg/NiftiImageData.cpp
@@ -220,7 +220,7 @@ void NiftiImageData<dataType>::construct_NiftiImageData_from_complex_im_real_com
 
     auto &it_in = in_sptr->begin();
     auto &it_out = out_sptr->begin();
-    for (; it_in!=in_sptr->end(); ++it_in, ++it_out)
+    for (; it_out != out_sptr->end(); ++it_in, ++it_out)
         *it_out = (*it_in).complex_float().real();
 }
 
@@ -235,7 +235,7 @@ void NiftiImageData<dataType>::construct_NiftiImageData_from_complex_im_imag_com
 
     auto &it_in = in_sptr->begin();
     auto &it_out = out_sptr->begin();
-    for (; it_in!=in_sptr->end(); ++it_in, ++it_out)
+    for (; it_out != out_sptr->end(); ++it_in, ++it_out)
         *it_out = (*it_in).complex_float().imag();
 }
 


src/Registration/cReg/NiftiImageData.cpp
```diff
@@ -220,7 +220,7 @@ void NiftiImageData<dataType>::construct_NiftiImageData_from_complex_im_real_com
 
     auto &it_in = in_sptr->begin();
     auto &it_out = out_sptr->begin();
-    for (; it_in!=in_sptr->end(); ++it_in, ++it_out)
+    for (; it_out != out_sptr->end(); ++it_in, ++it_out)
```
KrisThielemans, line 223: not sure why this is. Aren't they the same size?

evgueni-ovtchinnikov: no, the input image dimensions are (2, 256, 256) and the output (1, 256, 256) because the first dimension of a `NiftiImageData` object is z-dimension, which is 1 for both input and output images, but the input image is reconstructed from MR acquisition data with two repetitions.

KrisThielemans: Are we (silently) constructing a single volume from a repeated volume? That sounds wrong. I would suggest that we have a chechk that dimensions are identical (e.g. via `get_geometrical_info()` ) and if not throw an error.

Where is the call then that you're trying to fix?

evgueni-ovtchinnikov: Throwing an error to avoid segfault is not very helpful, I believe - I would rather issue a warning that the complex data contains more than one image, and only the first one is going to be used. @paskino which way would you prefer?

> Where is the call then that you're trying to fix?

In `Reg.NiftiImageData.construct_from_complex_image`, if I understand your question correctly

KrisThielemans: I'm trying to understand if this is a "conceptual" problem (i.e. what should we do with weird data) or a bug (i.e. we know how to handle the data, but the code doesn't do it).

What data do we stick into `Reg.NiftiImageData.construct_from_complex_image` where it throws this error? Is it a normal complex image, and If it is complex data with 2 repetitions, then I would argue that the output should be a real image with 2 repetitions, not silently throw away the 2nd repetition.

Unfortunately, this might be getting this into our problem of encoding MRI dimensions. From your description, I understand that data of a single slice with 2 repetitions is encoded as a 3D NiftiImageData. That in itself would be wrong (it should be a 4D image, with a single slice). And as I stated above, the `construct_from` functions really should check that all args have the same geometry. If they don't, an error should be thrown as it is invalid usage. Otherwise other crashes are unavoidable. 

I might misunderstand this completely of course...

KrisThielemans:
@evgueni-ovtchinnikov you will have to merge `master` here, as I've updated STIR `master`, and that needed small SIRF changes. So, `BUILD_DEVEL=ON` tests will fail.

KrisThielemans:
so... the tests worked. Can't say I understand that as I needed #1228. oh well.

PR #1230, opened 2024-01-15
fix STIR HDF5

KrisThielemans:
We're using `#ifdef HAVE_HDF5` but it should be `STIR_built_with_HDF5`. On my local machine, `HAVE_HDF5` is defined, although I cannot find where, so it was enabled, but on GHA it probably isn't.

I'll test that first. Then replace.


KrisThielemans:
`HAVE_HDF5` is set in `STIR/config.h`. `STIR_built_with_HDF5` is a CMake variable.

So this was all ok. Sorry

Issue #1249, opened 2024-05-09
time frame handling in listmode reconstruction


KrisThielemans:
- We do not have a `set_time_interval` yet for the list-mode objective function. It looks like this can be done in STIR by taking the objective function and setting its (public) `frame_defs` member. See also https://github.com/UCL/STIR/issues/1417
- Our demo sets a time interval for the creation of the scatter etc [here](https://github.com/SyneRBI/SIRF/blob/2c66faff3bc0f12c864cfc2a2931eba5ade60ba0/examples/Python/PET/listmode_reconstruction.py#L145), but as listmode reconstruction will be using all events, there is going to be a mismatch, so this is a bug.

@evgueni-ovtchinnikov easy to add a `set_time_interval` method to the listmode objective function?

evgueni-ovtchinnikov:
I understand it is not in STIR yet? Once you make it available in STIR, should not be difficult to access it from SIRF.

KrisThielemans:
It doesn't need STIR modifications luckily.

Essentially a copy of `ListmodeToSinograms::set_time_interval`, i.e. adding `set_time_interval` to `xSTIR_PoissonLLhLinModMeanListDataProjMatBin3DF` (as [here](e.g. https://github.com/SyneRBI/SIRF/blob/2c66faff3bc0f12c864cfc2a2931eba5ade60ba0/src/xSTIR/cSTIR/include/sirf/STIR/stir_x.h#L141-L148 )) and exposing it in exactly the same way

evgueni-ovtchinnikov:
fixed by #1253

Issue #1251, opened 2024-05-09
list-mode obj_fun.get_subset_sensitivity(0) returns None


KrisThielemans:
Example is at https://github.com/gschramm/SIRF-Exercises/blob/a581dc1e72d3cb1ff156ad88c2db770398a28c17/notebooks/Deep_Learning_listmode_PET/01_SIRF_listmode_recon.py#L176

It works fine for the "sinogram" objective function.

Found by @gschramm.

KrisThielemans:
@evgueni-ovtchinnikov wrote in https://github.com/SyneRBI/SIRF/pull/1253#issuecomment-2104760586
> which is due to the fact that `PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin` does not override empty `ObjectiveFunction.get_subset_sensitivity()`.

Can you elaborate a bit? `get_subset_sensitivity` is [implemented in `PoissonLogLikelihoodWithLinearModelForMean`](https://github.com/UCL/STIR/blob/4e8d40eb46f8d767f3821325acd1ff302e6dc7f0/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMean.cxx#L122). It also used in [`PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin::actual_compute_subset_gradient_without_penalty`](https://github.com/UCL/STIR/blob/4e8d40eb46f8d767f3821325acd1ff302e6dc7f0/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin.cxx#L868). It is of course entirely possible that there is a bug somewhere, but I'd like to understand what you saw.

evgueni-ovtchinnikov:
In STIR.py we have
```
class PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin(ObjectiveFunction):
```
From what you write it looks like this class should have been derived from `PoissonLogLikelihoodWithLinearModelForMean`, should it not?

evgueni-ovtchinnikov:
yes we do. I have just pushed the corrected STIR.py - let George try it.

gschramm:
@evgueni-ovtchinnikov @KrisThielemans was the update pushed to the master branch of SIRF?

Should I re-run the complete Super-Build with `DEVEL_BUILD=ON`?

I am have some time for tests this late afternoon / tomorrow morning.

[...]

PR #1253, opened 2024-05-10
Computing Hessian products with image data objects and related stuff

### LINKED ISSUE #1244: add accumulate_Hessian_times_input to STIR objective function and prior
STIR has functionality to multiplying the Hessian with an image for [priors](https://github.com/UCL/STIR/blob/8ced2d73933420457e0bc76074964b7d4ff00f0c/src/include/stir/recon_buildblock/GeneralisedPrior.h#L84) and [objective functions](https://github.com/UCL/STIR/blob/8ced2d73933420457e0bc76074964b7d4ff00f0c/src/include/stir/recon_buildblock/GeneralisedObjectiveFunction.h#L206-L209).

In STIR it's called `accumulate_Hessian_times_input` but "input" is a bit confusing (it does H . input), so suggestions welcome (@epapoutsellis @gschramm @Imraj-Singh @paskino).

This would have to be added to `sirf.STIR.ObjectiveFunction`. I think the C++ version can be as-is, but the Python one should be
```python
ret = accumulate_Hessian_times_input(current_estimate, input, optional_subset, out=optional)
```
without subset, it should do all of them I think (or at least, it should do the same as `gradient`).

evgueni-ovtchinnikov:
## Related issues

Fixes #1244, #1249

## Contribution Notes

Please tick the following: 


src/xSTIR/cSTIR/cstir.cpp
```diff
@@ -417,6 +417,23 @@ void* cSTIR_objectFromFile(const char* name, const char* filename)
 	CATCH;
 }
 
+typedef xSTIR_PoissonLLhLinModMeanListDataProjMatBin3DF LMObjFun;
+
+extern "C"
+void* cSTIR_objFunListModeSetInterval(void* ptr_f, size_t ptr_data)
```
KrisThielemans, line 423: Not really for a question for this PR, but why `size_t ptr_data` and not `void * ptr_data`? On some systems, they are not the same size, and this could therefore create trouble.

src/xSTIR/cSTIR/cstir.cpp
```diff
@@ -417,6 +417,23 @@ void* cSTIR_objectFromFile(const char* name, const char* filename)
...
+{
+	try {
+		LMObjFun& objFun = objectFromHandle<LMObjFun>(ptr_f);
```
KrisThielemans, line 426: can use `auto`

src/xSTIR/cSTIR/cstir.cpp
```diff
@@ -1249,6 +1266,36 @@ cSTIR_computeObjectiveFunctionGradientNotDivided(void* ptr_f, void* ptr_i, int s
 	CATCH;
 }
 
+extern "C"
+void*
+cSTIR_objectiveFunctionAccumulateHessianTimesInput
+    (void* ptr_fun, void* ptr_est, void* ptr_inp, int subset, void* ptr_out)
+{
+	try {
+		ObjectiveFunction3DF& fun = objectFromHandle<ObjectiveFunction3DF>(ptr_fun);
```
KrisThielemans, line 1275: I'd prefer `auto` in most/all of these.

src/xSTIR/cSTIR/cstir.cpp
```diff
@@ -1249,6 +1266,36 @@ cSTIR_computeObjectiveFunctionGradientNotDivided(void* ptr_f, void* ptr_i, int s
...
+			fun.accumulate_sub_Hessian_times_input(output, curr_est, input, subset);
+		else {
+			int nsub = fun.get_num_subsets();
+			output.fill(0.0);
+			shared_ptr<STIRImageData> sptr_sub(new STIRImageData(input));
+			Image3DF& sub = sptr_sub->data();
+			for (int s = 0; s < nsub; s++) {
+				fun.accumulate_sub_Hessian_times_input(sub, curr_est, input, subset);
+				output += sub;
+			}
```
KrisThielemans, line 1292: no need for the sum by hand. `accumulate*` automatically sums, so you can just do
```suggestion
			for (int s = 0; s < nsub; s++) {
				fun.accumulate_sub_Hessian_times_input(output, curr_est, input, subset);
```

src/xSTIR/cSTIR/cstir.cpp
```diff
@@ -1249,6 +1266,36 @@ cSTIR_computeObjectiveFunctionGradientNotDivided(void* ptr_f, void* ptr_i, int s
...
+		STIRImageData& inp = objectFromHandle<STIRImageData>(ptr_inp);
+		STIRImageData& out = objectFromHandle<STIRImageData>(ptr_out);
+		Image3DF& curr_est = est.data();
+		Image3DF& input    = inp.data();
+		Image3DF& output   = out.data();
+		if (subset >= 0)
+			fun.accumulate_sub_Hessian_times_input(output, curr_est, input, subset);
+		else {
+			int nsub = fun.get_num_subsets();
+			output.fill(0.0);
```
KrisThielemans, line 1286: if we're accumulating in SIRF (which I don't think we should), then this `fill` is wrong. On the other hand, if we're not accumulating, then you'll need it also in the case `subset>=0`

KrisThielemans (changes requested):

KrisThielemans:
Thanks, this already looks usable (with a caveat for the "accumulation"). Is this in a state that @gschramm can already try?

src/xSTIR/cSTIR/cstir.cpp
evgueni-ovtchinnikov: I vaguely remember having some SWIG trouble with `void*` but no longer remember what it was. Perhaps with the latest SWIG `void*` is now ok.

src/xSTIR/cSTIR/cstir.cpp
evgueni-ovtchinnikov: done

src/xSTIR/cSTIR/cstir.cpp
evgueni-ovtchinnikov: with `auto` got error
```
error: cannot declare variable ‘fun’ to be of abstract type ‘stir::GeneralisedObjectiveFunction<stir::DiscretisedDensity<3, float> >’
```

src/xSTIR/cSTIR/cstir.cpp
evgueni-ovtchinnikov: thanks!

src/xSTIR/cSTIR/cstir.cpp
evgueni-ovtchinnikov: moved the fill above `if` for now - will get rid of it when we switch to `multiply_with_Hessian`

evgueni-ovtchinnikov:
We still have #1251, which is due to the fact that `PoissonLogLikelihoodWithLinearModelForMeanAndListModeDataWithProjMatrixByBin` does not override empty `ObjectiveFunction.get_subset_sensitivity()`. I am afraid I cannot be of much help here.

src/xSTIR/cSTIR/cstir.cpp
KrisThielemans: interesting. I guess `auto&` could work

gschramm:
Just implemented a short test script for the **PoissonLogL** `.accumulate_Hessian_times_input()` [here](https://github.com/gschramm/SIRF-Exercises/blob/a0ef76efc839d0ad764f130de35da2d0d9b8a462/notebooks/Deep_Learning_listmode_PET/test.py#L238).

```python
current_estimate = ref_recon.copy()
input_img = acq_data.create_uniform_image(value=1, xy=nxny)
np.random.seed(0)
input_img.fill(np.random.rand(*input_img.shape) * (obj_fun.get_subset_sensitivity(0).as_array() > 0) * current_estimate.max())

hess_out_img = acq_data.create_uniform_image(value=1.0, xy=nxny)

obj_fun.accumulate_Hessian_times_input(current_estimate, input_img, subset=0, out=hess_out_img)
hess_out_img2 = obj_fun.accumulate_Hessian_times_input(current_estimate, input_img, subset=0)

fig, ax = plt.subplots(1, 4, figsize=(16, 4), tight_layout=True)
ax[0].imshow(current_estimate.as_array()[71, :, :], cmap = 'Greys')
ax[1].imshow(input_img.as_array()[71, :, :], cmap = 'Greys')
ax[2].imshow(hess_out_img.as_array()[71, :, :], cmap = 'Greys')
ax[3].imshow(hess_out_img2.as_array()[71, :, :], cmap = 'Greys')
ax[0].set_title('current estimate', fontsize = 'medium')
ax[1].set_title('input', fontsize = 'medium')
ax[2].set_title('hess_out_img', fontsize = 'medium')
ax[3].set_title('hess_out_img2', fontsize = 'medium')
fig.show()
```

Unfortunately, the output of both calls (without and with using the out kwarg), don't seem to work. 
`hess_out_img` is still an image full of ones and `hess_out_img2` seems to be the same as `input_img`.

Figure_1

KrisThielemans:
This is a test with protection data, correct?

gschramm:
Indeed. The `obj_fun` is 
```
obj_fun = sirf.STIR.make_Poisson_loglikelihood(acq_data)
```

KrisThielemans:
@gschramm could you try again?

gschramm:
Looks much better now. Tmr morning I will test whether we images are what we expect from the formulas.

Screenshot 2024-05-12 at 22 51 27

KrisThielemans:
Could work for listmode as well

gschramm:
@KrisThielemans I just compared against manually computing the "Hessian multiply" calling the fwd / back projections in SIRF. The result is very close in the foreground (where the current estimate is >> 0), but not super close in the background (where the current estimate is close to 0). Any ideas why that is?

```python
hess_out_img = obj_fun.accumulate_Hessian_times_input(current_estimate, input_img, subset=0)

# %%
# calcuate the the Hessian multiply "manually"

acq_model.set_up(acq_data, initial_image)
acq_model.num_subsets = num_subsets
acq_model.subset_num = 0

# get the linear (Ax) part of the Ax + b affine acq. model
lin_acq_model = acq_model.get_linear_acquisition_model()
lin_acq_model.num_subsets = num_subsets
lin_acq_model.subset_num = 0

# for the Hessian "multiply" we need the linear part of the acquisition model applied to the input image
input_img_fwd = lin_acq_model.forward(input_img)
current_estimate_fwd = acq_model.forward(current_estimate)
h = -acq_model.backward(acq_data*input_img_fwd / (current_estimate_fwd*current_estimate_fwd))
``` 

**The Hessian outputs are shown using a "narrow" color window (top) and a wide color window (bottom) to show the differences in foreground and background.**

Screenshot 2024-05-13 at 09 22 15

KrisThielemans:
hmmm. We try to cancel singularities in the division
https://github.com/UCL/STIR/blob/feb6d85eadb392f5b8278d3b97ae2ee67ca439d9/src/recon_buildblock/PoissonLogLikelihoodWithLinearModelForMeanAndProjData.cxx#L1128 with STIR's [divide_and_truncate function](https://github.com/UCL/STIR/blob/master/src/buildblock/recon_array_functions.cxx#L222-L250). However, possibly that goes a bit wrong, as it was designed for (measured / estimated), while here it is used for (fwd * measured / estimated^2), so maybe thresholds are a bit off.

Of course, ideally there wouldn't be any singularities. Is your background strictly positive?

gschramm:
1. The background term has indeed a few 0s (Siemens mMR data which I guess has virtual 0-sens. LORs).
2. In the meantime I also tested the listmode objective hessian. (i) currently, it returns the negative of what we want (output is positive, but should be negative). (ii) The "listmode hessian" is very close to my "manual sinogram" hessian (see below - I plotted the "minus the LM hessian output").

Screenshot 2024-05-13 at 11 47 33

KrisThielemans:
that's encouraging.

can you try your manual with some handling of 0/0 or so? A bit surprising it didn't generate NaNs then I guess. Easiest is to add an eps to the denominator.

I'll flip the sign of the LM Hessian. 😊

KrisThielemans:
ah well, this looks like a STIR issue then for the Hessian with projection data. Best to open an issue there.

KrisThielemans:
@evgueni-ovtchinnikov can you then add `multiply_with_Hessian`? (Can just as well leave the accumulate version in I guess)

KrisThielemans:
@gschramm what computation times are you getting?

evgueni-ovtchinnikov:
`multiply_with_Hessian`: @KrisThielemans which version/branch of STIR should I use? I have `rel_6.0.0` on my latest VMs.

src/xSTIR/cSTIR/cstir.cpp
evgueni-ovtchinnikov: BTW SWIG actually still does not like passing the numpy array data pointer as `void*`  - tried with `cSTIR_getImageData`, got this:
```
  File "/home/sirfuser/devel/buildVM/sources/SIRF/examples/Python/PET/acquisition_data.py", line 172, in <module>
    main()
  File "/home/sirfuser/devel/buildVM/sources/SIRF/examples/Python/PET/acquisition_data.py", line 141, in main
    image_array = image.as_array()
  File "/home/sirfuser/devel/install/python/sirf/STIR.py", line 611, in as_array
    try_calling(pystir.cSTIR_getImageData(self.handle, array.ctypes.data))
  File "/home/sirfuser/devel/install/python/sirf/pystir.py", line 306, in cSTIR_getImageData
    return _pystir.cSTIR_getImageData(ptr, ptr_data)
TypeError: in method 'cSTIR_getImageData', argument 2 of type 'void *'
sirfuser@vagrant:~/devel/buildVM/sources/SIRF/examples/Python/PET$ 
```
same story with `float*`:
```
  File "/home/sirfuser/devel/buildVM/sources/SIRF/examples/Python/PET/acquisition_data.py", line 172, in <module>
    main()
  File "/home/sirfuser/devel/buildVM/sources/SIRF/examples/Python/PET/acquisition_data.py", line 141, in main
    image_array = image.as_array()
  File "/home/sirfuser/devel/install/python/sirf/STIR.py", line 611, in as_array
    try_calling(pystir.cSTIR_getImageData(self.handle, array.ctypes.data))
  File "/home/sirfuser/devel/install/python/sirf/pystir.py", line 306, in cSTIR_getImageData
    return _pystir.cSTIR_getImageData(ptr, ptr_data)
TypeError: in method 'cSTIR_getImageData', argument 2 of type 'float *'
```

gschramm:
> @gschramm what computation times are you getting?

Using 21 subsets, mMR LM file with 4,332,816 counts and `acq_model = sirf.STIR.AcquisitionModelUsingRayTracingMatrix()`:

```
In [5]: %timeit hess_out_img = obj_fun.accumulate_Hessian_times_input(current_estimate, input_img, subset=0)
2.7 s ± 45.9 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

In [6]: %timeit hess_out_img_lm = lm_obj_fun.accumulate_Hessian_times_input(current_estimate, input_img, subset=0)
9.43 s ± 249 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

KrisThielemans:
> `multiply_with_Hessian`: @KrisThielemans which version/branch of STIR should I use? I have `rel_6.0.0` on my latest VMs.

you can use 6.0.0. Just add an extra member (that calls `accumulate`) and remove the `fill` in `accumulate.

KrisThielemans:
Thanks @gschramm would you mind adding the timings in the STIR PR? Ideally also add gradient timings. I'd be interested as well in seeing what it does if `num_subsets` decreases (i.e. is the listmode calculations slowing us down, or the reading). Of course, you could also enable the `lm_obj_fun` caching, but this is going beyond what you care about. Thanks a lot for all your help.

KrisThielemans:
1) No `parallelproj` yet. There's a difficulty that with the proj-matrix, I re-use the matrix-row for all calculations. That won't be possible with `parallproj`. So, we'll need new functions. Not sure if I can still do that.

2) In STIR, we can,  but it looks like this isn't enabled yet in SIRF. @evgueni-ovtchinnikov do we expose `PoissonLogLikelihoodWithLinearModelForMean::set_subset_sensitivity_filenames`? It should call
```c++
  //! set filename pattern to read (or write) the subset sensitivities
  /*! set to a zero-length string to avoid reading/writing a file
  Could be e.g. "subsens_%d.hv"
  boost::format is used with the pattern (which means you can use it like sprintf)

  Calls error() if the pattern is invalid.
 */
  void set_subsensitivity_filenames(const std::string&);
```
I see
https://github.com/SyneRBI/SIRF/blob/1d53570cff96acb33734ac1da15b94bc43373d92/src/xSTIR/cSTIR/cstir_p.cpp#L933-L934 but
https://github.com/SyneRBI/SIRF/blob/1d53570cff96acb33734ac1da15b94bc43373d92/src/xSTIR/cSTIR/cstir_p.cpp#L933
In any case, we need the subset version ATM.

If we have this, the usage is a bit awkward: first run as normal. then use `set_recompute_sensitivity(false)`, in which case it will go and find them.

gschramm:
@KrisThielemans Let's leave LM acq.model with Parallelproj for later

gschramm:
@KrisThielemans @evgueni-ovtchinnikov is `accumulate_Hessian_times_input` the final name? (I remember discussions on the naming) If possible, I'd like to finish my notebooks today.

src/xSTIR/cSTIR/cstir.cpp
KrisThielemans: hmmm. weird. Here's an example that does that https://stackoverflow.com/a/37308401. Anyway, let's leave that for later!

KrisThielemans:
@evgueni-ovtchinnikov is working on implementing https://github.com/SyneRBI/SIRF/issues/1244#issuecomment-2103346336. Hopefully this will be done soon. @evgueni-ovtchinnikov if you have trouble, please ask for help.

KrisThielemans:
@evgueni-ovtchinnikov here's the STIR test I talked about. https://github.com/UCL/STIR/blob/90cff236be35628f447ded4d98a046d5c4e3b316/src/include/stir/recon_buildblock/test/ObjectiveFunctionTests.h#L169
Best to implement the name first though.

KrisThielemans:
@evgueni-ovtchinnikov you seem to be pretty close. Can you summarise current situation?

evgueni-ovtchinnikov:
@KrisThielemans @gschramm I believe this PR can be merged - any objections?

evgueni-ovtchinnikov:
builds failed at Coveralls step - any idea anyone why all of a sudden?

src/xSTIR/cSTIR/cstir.cpp
```diff
@@ -417,12 +417,28 @@ void* cSTIR_objectFromFile(const char* name, const char* filename)
...
+typedef xSTIR_PoissonLLhLinModMeanListDataProjMatBin3DF LMObjFun;
+
+extern "C"
+void* cSTIR_objFunListModeSetInterval(void* ptr_f, size_t ptr_data)
+{
+	try {
+		auto objFun = objectFromHandle<LMObjFun>(ptr_f);
+		// Does not compile!!!
+	    //xSTIR_PoissonLLhLinModMeanListDataProjMatBin3DF& objFun =
+		//	objectFromHandle<xSTIR_PoissonLLhLinModMeanListDataProjMatBin3D>(ptr_f);
```
KrisThielemans, line 429:
```suggestion
		auto& objFun = objectFromHandle<LMObjFun>(ptr_f);
```

src/xSTIR/cSTIR/cstir.cpp
```diff
@@ -417,12 +417,28 @@ void* cSTIR_objectFromFile(const char* name, const char* filename)
...
+	CATCH;
+}
+
 extern "C"
 void* cSTIR_setListmodeToSinogramsInterval(void* ptr_lm2s, size_t ptr_data)
 {
 	try {
-		ListmodeToSinograms& lm2s = 
-			objectFromHandle<ListmodeToSinograms>(ptr_lm2s);
+		auto lm2s = objectFromHandle<ListmodeToSinograms>(ptr_lm2s);
```
KrisThielemans, line 441:
```suggestion
		auto& lm2s = objectFromHandle<ListmodeToSinograms>(ptr_lm2s);
```

src/xSTIR/cSTIR/cstir.cpp
```diff
@@ -1249,6 +1265,51 @@ cSTIR_computeObjectiveFunctionGradientNotDivided(void* ptr_f, void* ptr_i, int s
...
+{
+	try {
+		auto& fun = objectFromHandle<ObjectiveFunction3DF>(ptr_fun);
+		auto& est = objectFromHandle<STIRImageData>(ptr_est);
+		auto& inp = objectFromHandle<STIRImageData>(ptr_inp);
+		auto& out = objectFromHandle<STIRImageData>(ptr_out);
+		auto& curr_est = est.data();
+		auto& input    = inp.data();
+		auto& output   = out.data();
+		output.fill(0.0);
```
KrisThielemans, line 1281:
```suggestion
```

src/xSTIR/pSTIR/STIR.py
```diff
@@ -2333,6 +2333,15 @@ def set_up(self, image):
         """Sets up."""
         try_calling(pystir.cSTIR_setupPrior(self.handle, image.handle))
 
+    def accumulate_Hessian_times_input(self, current_estimate, input_, out=None):
+        """Computes the multiplication of the Hessian with a vector and adds it to output.
+        """
+        if out is None or out.handle is None:
+            out = input_.clone()
```
KrisThielemans, line 2340:
```suggestion
            out = input_.get_uniform_copy(0)
```

src/xSTIR/pSTIR/STIR.py
```diff
@@ -2732,6 +2741,41 @@ def get_subset_gradient(self, image, subset, out=None):
         """
         return self.gradient(image, subset, out)
 
+    def accumulate_Hessian_times_input(self, current_estimate, input_, subset=-1, out=None):
+        """Computes the multiplication of the Hessian with a vector and adds it to output.
+        """
+        if out is None or out.handle is None:
+            out = input_.clone()
```
KrisThielemans, line 2748:
```suggestion
            out = input_.get_uniform_copy(0)
```

examples/Python/PET/osem_reconstruction.py
```diff
@@ -128,6 +128,8 @@ def main():
     print('setting up, please wait...')
     recon.set_up(image)
 
+    obj_fun.test_Hessian(image, subset=0)
+
```
KrisThielemans, line 132:
```suggestion
```

KrisThielemans:
Also, please comment out the Coveralls lines
https://github.com/SyneRBI/SIRF/blob/d54c4550d1533132c6a281b6cf2159c8fd22f45c/.github/workflows/build-test.yml#L133C4-L147C32
You probably have to preserve indentation. Or leave that for a separate PR

KrisThielemans:
@evgueni-ovtchinnikov `test_Hessian` isn't a great test, as it just compares the norms, while it should compare the difference. 

I'm modifying it and putting it in test_ObjectiveFunction.py
```python
  def test_Hessian(self, subset=-1, eps=1e-3):
        """Checks that grad(x + dx) - grad(x) is close to H(x)*dx
        """
        x = self.image
        dx = x.clone()
        dx *= eps/dx.norm()
        dx += eps/2
        y = x + dx
        gx = self.obj_fun.gradient(x, subset)
        gy = self.obj_fun.gradient(y, subset)
        dg = gy - gx
        Hdx = self.obj_fun.multiply_with_Hessian(x, dx, subset)
        norm = dg.norm()
        q = (dg - Hdx).norm()/dg.norm()
        print('norm of grad(x + dx) - grad(x): %f' % dg.norm())
        print('norm of H(x)*dx: %f' % Hdx.norm())
        print('relative difference: %f' % q)
        numpy.testing.assert_array_equal(q,0)
```
However, I had to add a constant term to the acq_model as otherwise the test fails (essentially because the Hessian is ill-defined otherwise). I'll commit this soon.

KrisThielemans:
@evgueni-ovtchinnikov I'm now adding a similar test to the priors. (Had to delete a `subset` in the call). However, I'm confused by results. Does `get_gradient(x)` modify the result? Looks like it for priors.

evgueni-ovtchinnikov:
@KrisThielemans 
> Does get_gradient(x) modify the result? Looks like it for priors.

No, it cannot, `get_gradient(image)` calls (eventually) STIR's prior `compute_gradient`, where the image argument is const.

KrisThielemans:
got distracted... Sorry,

Id di push the test for the LogLikelihood. Works for me.

KrisThielemans:
> @KrisThielemans
> 
> > Does get_gradient(x) modify the result? Looks like it for priors.
> 
> what result?

ignore that. The loop in the test is a bit hard to understand... Update coming soon.

KrisThielemans:
ok. Should be all done now. Can you add a line to CHANGES.md and merge?

PR #1263, opened 2024-06-06
Revert "Moving SyneRBI-Challenge nema-data utilities to SIRF (#1241)"

KrisThielemans:
This reverts commit b90383e72001d994efd7d0600e5d52394c415a36.


evgueni-ovtchinnikov (approved):
I do not mind reverting if what I have done is not good enough, but I am afraid I am completely lost now. If randoms cannot be computed for SPECT data, we need a method that computes prompts only, do we not? BTW is there a way to find out whether one has PET or SPECT data?

KrisThielemans:
>  If randoms cannot be computed for SPECT data, we need a method that computes prompts only, do we not?

sure, but we have one already, i.e. the `set_input`, `process()`, `get_output()`.

> BTW is there a way to find out whether one has PET or SPECT data?

You have to get into `stir::ExamInfo::imaging_modality`. We don't expose any of that to SIRF, and will have to.

PR #1299, opened 2025-01-20
Error raised in AcquisitionSensitivityModel.[un]normalise methods applied to a read-only object

### LINKED ISSUE #1298: STIR.AcquisitionSensitivityModel.forward etc can crash python
`stir.BinNormalisation.apply` etc can throw exceptions (e.g. when not being able to write to file in the example below). They are currenty lot caught
https://github.com/SyneRBI/SIRF/blob/b2053845e25f87149d1a7406846110e4bf66817e/src/xSTIR/cSTIR/stir_x.cpp#L415-L434
leading to python just exciting:
```
>>> m=sirf.STIR.AcquisitionData('SIRFmultfactors_withatn_3.hs')
>>> asm.set_up(p)
>>> xx=asm.unnormalise(p)

WARNING: write_data: error after writing to stream.


ERROR: ProjDataFromStream::set_viewgram: viewgram (view=0, segment=-11, timing_pos=-5) corrupted due to problems with writing or the scale factor (out of disk space?)
terminate called after throwing an instance of 'std::runtime_error'
  what():  ProjDataFromStream::set_viewgram: viewgram (view=0, segment=-11, timing_pos=-5) corrupted due to problems with writing or the scale factor (out of disk space?)
Aborted
```
Looks like we need to `TRY...CATCH` stuff around these calls.

evgueni-ovtchinnikov:

Error raised in `AcquisitionSensitivityModel.[un]normalise` methods if applied to a read-only object.

## Testing performed

Yes

## Related issues

Quick fix for issue 1298

## Contribution Notes

Please tick the following: 

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,8 @@
 
 * SIRF/STIR
   - `ScatterEstimation` has extra methods that allow setting masks for the tail-fitting
+  - Error raised in `AcquisitionSensitivityModel.[un]normalise` methods applied to a read-only object.
+
 
 
 ## v3.8.1
--- a/src/xSTIR/pSTIR/STIR.py
+++ b/src/xSTIR/pSTIR/STIR.py
@@ -1685,6 +1685,9 @@ class AcquisitionSensitivityModel(object):
         if self.handle is None:
             raise AssertionError()
         assert_validity(ad, AcquisitionData)
+        if ad.read_only:
+            raise error(
+                'Cannot normalise a read-only object, consider using method invert instead')
         try_calling(pystir.cSTIR_applyAcquisitionSensitivityModel(
             self.handle, ad.handle, 'normalise'))
 
@@ -1697,6 +1700,9 @@ class AcquisitionSensitivityModel(object):
         if self.handle is None:
             raise AssertionError()
         assert_validity(ad, AcquisitionData)
+        if ad.read_only:
+            raise error(
+                'Cannot normalise a read-only object, consider using method forward instead')
         try_calling(pystir.cSTIR_applyAcquisitionSensitivityModel(
             self.handle, ad.handle, 'unnormalise'))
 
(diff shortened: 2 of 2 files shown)


KrisThielemans (commented):
I'm a bit confused about the catching of a STIR error. As you saw, python just exited due to uncaught exception presumably. I'm not sure what `try_calling` does, but it looks like it didn't catch it.

Calling STIR functions could always get into `stir.error`, so I think it'd be safest to make sure we catch that somehow. Easy?

KrisThielemans:
The build error is due to https://github.com/gadgetron/gadgetron/issues/1277. They haven't looked at my PR yet.

I've downgraded to GHA to use Ubuntu 22.04 in #1302, so merging `master` here would work-around this problem.

KrisThielemans:
> So far as I can see, methods `undo` and `apply` of `BinNormalisation` failed when applied to data that was supposed to be read-only but did not throw an exception, otherwise this exception would be caught by `try` in `cSTIR_applyAcquisitionSensitivityModel` and reported by `try_calling` (which just checks the execution status recorded in the `DataHandle` object returned by every C-interface function, in this case by `cSTIR_applyAcquisitionSensitivityModel`).

I don't know what's going on there, but as mentioned in https://github.com/SyneRBI/SIRF/issues/1298#issue-2795502300, STIR does throw an exception (see the error `terminate called after throwing an instance of 'std::runtime_error'`). So somehow SIRF didn't catch those.

KrisThielemans:
The modifications to use the `read_only` attribute are good of course.

KrisThielemans:
Great. @evgueni-ovtchinnikov don't forget to create a new issue for the `CATCH` problem

evgueni-ovtchinnikov:
I suggest re-using #1298, which I have just renamed "`STIR.AcquisitionSensitivityModel.[un]normalise` does not catch `std::runtime_error` thrown by STIR causing SIRF crash"

PR #1301, opened 2025-01-20
Fix check on number of arguments in sirf_inverse_transform.cpp

### LINKED ISSUE #1300: sirf_inverse_transform requires 4th argument, even though it should be optional

KrisThielemans:
Fixes #1300


KrisThielemans:
Compilation error in Gadgetron due to https://github.com/gadgetron/gadgetron/issues/1277.  Looks like we will need to downgrade GHA to Ubunti 22.04 for the moment.

PR #1346, opened 2025-10-27
minor improvements to STIR Hessian code (require STIR 6.3)

KrisThielemans:
- call `accumulate_Hessian_times_input` in the no-subset case, as this will now be a bit faster (it's also a tiny bit less code)
- give correct name of parameters in cstir.h to avoid confusion
- add test for objective function with prior
- clarify RDP exclusion of test case


KrisThielemans:
> I now expect the DEVEL_BUILD=OFF job to fail, but ON to succeed.

indeed

KrisThielemans:
@evgueni-ovtchinnikov @paskino please update the SIRF-SuperBuild to use `DEFAULT_STIR_TAG` to current STIR master

PR #1358, opened 2025-11-17
Extends SIRF data algebra possible operands to numpy arrays

### LINKED ISSUE #1357: backwards incompatibility: algebraic operations with numpy arrays/scalars on the "left" return a numpy array
In the previous version of SIRF (e.g. used for PETRIC1), things like `sirf_image * numpy_array` worked, but now it raises an error. This breaks for instance the MaGeZ implementation. https://github.com/SyneRBI/PETRIC2/issues/12.

@evgueni-ovtchinnikov is it possible to restore this?

evgueni-ovtchinnikov:
## Related issues

fixes #1357 (see instructions on the correct usage of extended functionality in User Guide section "SIRF data algebra")

- [x] I have implemented unit tests that cover any new or modified functionality (added in `Utilities.data_container_algebra_tests`)

## Contribution Notes

Please tick the following: 

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,7 +3,7 @@
 ## v3.9.1
 
 * Python interface
-  - Restored functionality for algebraic operations mixing STIR.ImageData and numpy arrays.
+  - Restored functionality for algebraic operations mixing STIR.ImageData and numpy arrays. (Note that sirf objects need to be on the "left" of the operation.)
 
 ## v3.9.0
 


evgueni-ovtchinnikov:
Actually, `get_uniform_copy` cannot be any faster as it calls `clone` to make a copy and then fills it with the specified value:
```
    def get_uniform_copy(self, value=1.0):
        '''Initialises an instance of DataContainer based on the template'''
        y = self.clone() # SIRF uses `DiscretisedDensity.clone()` to make a copy of `self` data
        y.fill(value)
        return y
```

src/xGadgetron/pGadgetron/Gadgetron.py
```diff
@@ -437,6 +437,56 @@ def get_ISMRMRD_info(self, par):
...
 
+    def dot(self, other):
+        '''
+        Returns the dot product of the container data with another container
+        data or numpy array viewed as vectors.
+        other: ImageData or numpy array or scalar
+        '''
+        if not (issubclass(type(other), type(self))):
+            other = self.clone().fill(other)
+        return super(ImageData, self).dot(other)
```
KrisThielemans: according to doc, since Python 3.0, you can write
```suggestion
        return super().dot(other)
```
which is a nicer, and in fact seems to say that we can write this only once somewhere, as opposed to for Gadgetron/STIR/Reg etc. No idea how to do that though.

PR #1381, opened 2026-03-13
 Hackathon Project 8: Expose CUDA-managed STIR image and acquisition queries in sirf.STIR

Dimitra-Kyriakopoulou:
This PR is PART of Hackathon Project 8: update STIR data to CUDA managed pointers, then adapt `sirf.STIR` data-containers and expose the underlying CUDA managed pointer to Python. This PR contains the full SIRF side of that work across Stage 1, Stage 2, and Stage 3.
- Stage 1 covers CUDA-managed STIR image queries and Python exposure for the image path. 
- Stage 2 extends the same path to acquisition/projection data. 
- Stage 3 uses these SIRF changes together with the companion STIR changes in the end-to-end managed-memory reconstruction proof. 

The STIR image part has already been merged upstream in [`UCL/STIR#1693`](https://github.com/UCL/STIR/pull/1693) (Stage 1) and [`UCL/STIR#1694`](https://github.com/UCL/STIR/pull/1694) (Stage 2 and Stage 3). The runnable benchmark/proof repository for the whole Project 8 path, including Stage 1, Stage 2, and Stage 3, is `Dimitra-Kyriakopoulou/hackathon_project_08_stir_cuda_managed_pointers`.

  - Added image-level CUDA-managed-memory queries in `cSTIR`.
  - Added acquisition-level CUDA-managed-memory queries in `cSTIR`.
  - Added Python exposure in `sirf.STIR` for both managed-image and managed-acquisition paths.
  - The Python-facing query interface now exposes, for the relevant managed STIR objects:
      - `is_cuda_managed`
      - `cuda_address`
      - `__cuda_array_interface__`
  - For the image path, this means Python can detect whether the image is CUDA-managed, retrieve its underlying address, and obtain a CUDA-array-style view.
  - For the acquisition path, the same query/exposure pattern is now available for managed acquisition data.
  - Added the acquisition-side query exposure in `cstir_p.cpp`.
  - Kept the Stage 1 image-query path and extended the same design through
    Stage 2 and Stage 3, rather than introducing a separate parallel interface.
  - Propagated `STIR_WITH_CUDA` into `cstir` and linked `CUDA::cudart`, so the CUDA-aware query path is actually compiled in the relevant target.
  - Scope now covers the full SIRF Project 8 path needed for Stage 1, Stage 2, and Stage 3.

## Testing performed
  - Validated the C++ SIRF image probe end to end.
  - Validated Python import of `sirf.STIR` with the managed-image factory.
  - Checked the Stage 1 image-side Python properties on a real managed image:
      - `is_cuda_managed`
      - `cuda_address`
      - `__cuda_array_interface__`
  - Ran the Stage 1 phantom-clone benchmark; the managed image matched the source phantom exactly.
  - Validated the Stage 2 deterministic acquisition benchmark.
  - Checked the Stage 2 acquisition-side Python properties on real managed acquisition data:
      - `supports_array_view = true`
      - `is_cuda_managed = true`
      - `cuda_address = nonzero`
  - Validated the Stage 3 reconstruction proof:
      - managed acquisition input remained managed
      - managed image output remained managed
      - the CUDA address of the managed output did not change across the reconstruction update
      - managed and normal reconstructions matched to machine precision
  - The runnable proof-of-use benchmarks for Stage 1, Stage 2, and Stage 3 are kept separately in `Dimitra-Kyriakopoulou/hackathon_project_08_stir_cuda_managed_pointers`.

- [x ] The code builds and runs on my machine

## Contribution Notes

Please tick the following: 

 - [x ] The content of this Pull Request (the Contribution) is intentionally submitted for inclusion in SIRF (the Work) under the terms and conditions of the [Apache-2.0](https://www.apache.org/licenses/LICENSE-2.0) License.


.github/workflows/build-test.yml
```diff
@@ -2,7 +2,7 @@ name: Build-test CI
 
 on:
   push:
-    branches: [ master ]
+    branches: [ master, Hackathon_project8_II ]
```
KrisThielemans: will need to undo

src/xSTIR/cSTIR/CMakeLists.txt
```diff
@@ -42,6 +42,11 @@ target_include_directories(cstir PUBLIC
     "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}>$<INSTALL_INTERFACE:include>")
 target_include_directories(cstir PUBLIC "${STIR_INCLUDE_DIRS}")
 
+if (STIR_WITH_CUDA)
```
KrisThielemans: we'll need a `SIRF_WITH_CUDA` instead

src/xSTIR/cSTIR/stir_data_containers.cpp
```diff
@@ -25,12 +25,50 @@ limitations under the License.
...
+#if defined(__has_include)
+#if __has_include(<cuda_runtime_api.h>)
+#include <cuda_runtime_api.h>
+#define SIRF_HAS_CUDA_RUNTIME_API 1
+#endif
+#else
+#include <cuda_runtime_api.h>
+#define SIRF_HAS_CUDA_RUNTIME_API 1
+#endif
+#endif
```
KrisThielemans: we should define `HAS_CUDA_RUNTIME_API` at the CMake stage

src/xSTIR/pSTIR/STIR.py
```diff
@@ -478,6 +478,28 @@ def __array_interface__(self):
...
+    @property
+    def cuda_address(self):
+        """Returns the CUDA-visible image pointer for managed-memory-backed images."""
+        if not self.is_cuda_managed:
+            raise error('image data is not backed by CUDA managed memory')
+        return parms.size_t_par(self.handle, 'ImageData', 'cuda_address')
+
+    @property
+    def __cuda_array_interface__(self):
+        """As per https://numba.readthedocs.io/en/stable/cuda/cuda_array_interface.html"""
```
KrisThielemans: apparently that site is deprecated. 
```suggestion
        """As per https://nvidia.github.io/numba-cuda/user/cuda_array_interface.html"""
```

KrisThielemans (changes requested):
I think we need to move some things to `DataContainer`, in particular . Not so sure why we didn't do that yet with `address()` itself.

I suggest to rename `is_cuda_managed` to `supports_cuda_array_view` to be consistent with existing code. We should have a virtual `DataContainer` version that just returns `false`.

I'm not so sure we need `cuda_address()` as it's really the same as `address()` but with a check. If we do have it, it should be in `DataContainer` (calling virtual `DataContainer::address()`), and it shouldn't be anywhere else.

KrisThielemans (changes requested):
Thanks!

Let's move the `pointer` functions to `common/utilities.h` (sorry).

Also, all `address()` functions in the derived classes will now need `override`. And add something to `CHANGES.md` along the lines of "add `supports_cuda_array_view()` and related functions to expose CUDA arrays directly to Python for DataContainers that support it."

src/xSTIR/cSTIR/CMakeLists.txt
```diff
@@ -42,6 +48,20 @@ target_include_directories(cstir PUBLIC
...
+  if (_cudart_include_dirs)
+    set(CMAKE_REQUIRED_INCLUDES ${_cudart_include_dirs})
+  endif()
+  check_include_file_cxx(cuda_runtime_api.h HAS_CUDA_RUNTIME_API)
+  set(CMAKE_REQUIRED_INCLUDES "${_saved_required_includes}")
+  if (HAS_CUDA_RUNTIME_API)
+    target_compile_definitions(cstir PUBLIC HAS_CUDA_RUNTIME_API)
+    target_link_libraries(cstir CUDA::cudart)
+  endif()
+endif()
```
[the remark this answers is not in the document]
Dimitra-Kyriakopoulou: I simplified this by dropping the extra header check and using TARGET CUDA::cudart directly.

src/Registration/cReg/include/sirf/Reg/NiftiImageData.h
```diff
@@ -1,6 +1,7 @@
 /*
 SyneRBI Synergistic Image Reconstruction Framework (SIRF)
-Copyright 2017 - 2020 University College London
+Copyright 2017 - 2020, 2026 University College London
+Copyright 2026 Biomedical Research Foundation, Academy of Athens
```
KrisThielemans: I don't think it makes sense to add copyright info and authorship for just adding `override`. (also in other places)

src/xSTIR/cSTIR/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 #========================================================================
-# Author: Kris Thielemans, Richard Brown
+# Author: Kris Thielemans
+# Author: Richard Brown
+# Author: Dimitra Kyriakopoulou
```
KrisThielemans: looks like we should undo these changes

src/xSTIR/cSTIR/stir_data_containers.cpp
```diff
@@ -1,7 +1,8 @@
 /*
 SyneRBI Synergistic Image Reconstruction Framework (SIRF)
 Copyright 2017 - 2023 Rutherford Appleton Laboratory STFC
-Copyright 2018 - 2023 University College London
+Copyright 2018 - 2023, 2026 University College London
+Copyright 2026 Biomedical Research Foundation, Academy of Athens
```
KrisThielemans: undo

CHANGES.md
```diff
@@ -4,6 +4,7 @@
 
 * SIRF/STIR
   - The implementation of the creation of `sirf.STIR.ImageData` from `sirf.STIR.AcquisitionData` has been revised to ensure compatibility of `ImageData` dimensions and voxel sizes with `AcquisitionData`.
+  - Added `supports_cuda_array_view()` and related `DataContainer`/Python support, including `__cuda_array_interface__`, to expose CUDA-backed arrays directly for supported containers such as STIR image and acquisition data.
```
KrisThielemans:
```suggestion
  - Added `supports_cuda_array_view()` and related `DataContainer`/Python support, including `__cuda_array_interface__`, to expose CUDA-backed arrays directly for supported containers
```
we have to cut STIR support, as it hasn't been merged yet

KrisThielemans (changes requested):
A little bit of clean-up. We have reduced the number of changes, which is always good!

Dimitra-Kyriakopoulou:
Dear Professor @KrisThielemans,
THANK YOU WHOLEHEARTEDLY!!!

I faced two separate CI failures locally.

  1. `ImageWrap::address() const override`

  I applied the review request that all derived `address()` functions should get `override` too broadly by adding it in
  `gadgetron_image_wrap.h`. However, `ImageWrap` is not derived from `DataContainer`/`ImageData`. I fixed this by
  removing `override` there.

  2. `MR_PROCESS_TESTDATA` link error with `GEHDF5Wrapper` / `H5::...`

  Undoing the temporary CI restrictions restored the full build in `.github/workflows/build-test.yml`. This caused CI to
  build the Gadgetron/Synergistic path again, including `MR_PROCESS_TESTDATA`, which is why this problem became visible.

  The likely direct trigger in my implementation was that, in the follow-up review changes, I moved the generic
  `supports_cuda_array_view()` implementation out of `DataContainer.h` and into `csirf.cpp`. That implementation choice
  appears to have pulled a broader STIR/Synergistic link path into that Gadgetron test.

  I am therefore keeping `supports_cuda_array_view()` generic at the `DataContainer` level, but moving its
  implementation back inline into `DataContainer.h` instead of defining it out of line in `csirf.cpp`.

I will be looking forward to your further evaluation!

**THANK YOU WHOLEHEARTEDLY FOR ALL YOUR INVALUABLE HELP!!!**
Dimitra

src/common/include/sirf/common/Common.h
```diff
@@ -0,0 +1,63 @@
...
+limitations under the License.
+
+*/
+
+/*!
+\file
+\ingroup Common
+
+\author Dimitra Kyriakopoulou
+\author Kris Thielemans
```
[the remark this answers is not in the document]
Dimitra-Kyriakopoulou: This proved related to the second CI issue (please see post below). The problem there was that I had placed the generic `DataContainer::supports_cuda_array_view()` implementation in `csirf.cpp`. I now moved it to `src/common/DataContainer.cpp` and removed it from `src/common/csirf.cpp`, so it no longer lives in the C interface file and `DataContainer.h` still does not need to include `utilities.h`.

src/common/DataContainer.cpp
```diff
@@ -0,0 +1,34 @@
...
+#include "sirf/common/DataContainer.h"
+#include "sirf/common/utilities.h"
+
+namespace sirf {
+
+bool
+DataContainer::supports_cuda_array_view() const
+{
+	return this->supports_array_view() &&
+		pointer_supports_cuda_array_view(reinterpret_cast<const void*>(this->address()));
```
KrisThielemans, line 31: I don't think we need the cast

```suggestion
		pointer_supports_cuda_array_view(this->address());
```

src/common/csirf.cpp
```diff
@@ -50,6 +51,13 @@ unknownObject(const char* obj, const char* name, const char* file, int line)
 	return handle;
 }
 
+bool
+DataContainer::supports_cuda_array_view() const
+{
+	return this->supports_array_view() &&
+		pointer_supports_cuda_array_view(reinterpret_cast<const void*>(this->address()));
+}
+
```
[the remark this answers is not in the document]
KrisThielemans:
```suggestion
```

src/common/DataContainer.cpp
KrisThielemans: We do need the cast. Sorry. I'll revert.

KrisThielemans:
I've messed this up in various ways. I will remove my commits and start again. Apologies

KrisThielemans:
Fixed (I hope). I've force-pushed a replacement for my 2 commits, didn't touch the rest.

src/common/CMakeLists.txt
```diff
@@ -31,6 +31,11 @@ target_include_directories(csirf PUBLIC
   "$<BUILD_INTERFACE:${BUILD_INCLUDE_DIRECTORY}>$<INSTALL_INTERFACE:include>"
   )
 
+if (TARGET CUDA::cudart)
+  target_compile_definitions(csirf PRIVATE HAS_CUDA_RUNTIME_API)
+  target_link_libraries(csirf PRIVATE CUDA::cudart)
```
KrisThielemans, line 36: This is probably a bit problematic as it is used in `utilities.h`, not in `csirf`.

Better to move the implementation into utilities.cxx

KrisThielemans (approved):
This looks good to me. Codacy issues are due to existing coding style.

@evgueni-ovtchinnikov please check.
@casperdcl can we add a CI with the CUDA API? I think we can even run all tests there at the moment (as `pointer_supports_cuda_array_view` will not be called by anyone)

KrisThielemans:
Corrections needed after all. Pushing later.

KrisThielemans:
GitHub got confused and was running the previous commit.

KrisThielemans:
Finally green again. @casperdcl I've put CUDA toolkit in (but skip all tests). I think that'll do for now, as it might be hard to find out which tests are going to fail (I think more or less only anything with Parallelproj or STIR CUDA priors).

I suggest we merge this soon.

Dimitra-Kyriakopoulou:
Dear Professor @KrisThielemans,

**THANK YOU WHOLEHEARTEDLY for the meticulous and technically demanding work you put into giving this CUDA path a much stronger, safer, and more maintainable foundation!!!**

**I)** I traced your July changes through `DataContainer`, the C interface, the Python bindings, and the CUDA CI path. What impressed me most is that the changes protect the delicate contracts between these layers - where CUDA capability is decided, when it is valid to request a data address, how a single pointer identity is preserved, and how temporary C handles are owned and checked from Python.

On the C++ side, one of the strongest parts of the revision is that the CUDA-array-view capability is now handled at the correct abstraction level. `DataContainer::supports_cuda_array_view()` belongs to the container logic, while `csirf.cpp` remains an interface to that logic rather than carrying a second implementation. I particularly appreciate this decision because duplicated capability semantics across layers could later become inconsistent.

The ordering in

```cpp
supports_array_view() &&
pointer_supports_cuda_array_view(address())
```

is a particularly subtle and important safeguard. `supports_array_view()` is the precondition that makes calling `address()` legitimate. Because the expression short-circuits, containers without a valid contiguous array view are never asked for an address that may be invalid or meaningless for them. Then `cuda_address()` validates the CUDA capability and returns the same canonical `address()`, preserving one authoritative pointer identity across the C++, C, and Python interfaces.

Moving `pointer_supports_cuda_array_view()` out of the public utility header and into `utilities.cpp` is another strong architectural improvement. It keeps `cudaPointerGetAttributes()`, the CUDA runtime headers, the compile-time CUDA conditions, and the CUDA-version-specific pointer handling private to the implementation. CUDA therefore remains optional at the compilation boundary, not only at runtime, while the non-CUDA path still has a clean and explicit `false` result. I greatly appreciate how carefully this preserves the generic SIRF interface without weakening the CUDA functionality.

I was especially impressed by the Python update because it eliminates a potentially fragile cross-language ownership pattern. The `supports_array_view`, `supports_cuda_array_view`, `address`, and `cuda_address` properties now use the newer `Handle` path. This centralises status checking, typed extraction, temporary `DataHandle` destruction, and error propagation instead of repeating that ownership protocol manually in every property. The result is not only shorter code, but substantially safer and more maintainable bindings.

The CUDA CI matrix entry is another especially valuable part of the work. It now exercises the CUDA-specific configuration, compilation, and linking paths in a dedicated CUDA-enabled environment. This gives direct protection against regressions inside the `HAS_CUDA_RUNTIME_API` branch or in the `CUDA::cudart` linkage. The compiler- and CUDA-aware ccache key is also an important safeguard against accidentally reusing cached objects produced under incompatible toolchain configurations.

**II)** Because CUDA runtime validation was still open, I ran a separate focused check in Google Colab using a T4 GPU. CUDA is not available in my local WSL environment (and my three-month Google Cloud free trial had already expired), so Colab was used specifically to exercise the CUDA-managed-memory path at runtime.

The Colab validation used local source archives only: the SIRF PR 1381 head, the Project 8 probe scripts, and patched STIR source with the relevant companion STIR changes. It confirmed an NVIDIA GPU and CUDA compiler, built patched STIR with CUDA enabled, built SIRF PR 1381 against that STIR installation, and ran the managed-memory image, acquisition, and reconstruction probes.

This was a focused runtime validation, not a replacement for the complete upstream CI suite. The Colab run used a reduced SIRF/STIR configuration and small local test-harness adjustments needed to run the Project 8 probes in that environment.

The results were:

* `nvidia_smi_exit_code: 0`
* `nvcc_exit_code: 0`
* STIR build/install: `PASS`
* SIRF build: `PASS`
* managed image comparison: RMSE `0.0`, maximum absolute error `0.0`, correlation `1.0`, `supports_cuda_array_view: true`
* managed acquisition comparison: RMSE `0.0`, maximum absolute error `0.0`, correlation `1.0`, `supports_cuda_array_view: true`
* managed reconstruction comparison: RMSE `6.908700278529523e-10`, maximum absolute error `5.587935447692871e-09`, correlation `0.9999999999999934`
* managed reconstruction output kept the same CUDA address before and after reconstruction: `output_address_unchanged: true`

Within the scope of this focused validation, the patched STIR/SIRF stack builds successfully in a CUDA-capable environment, and the managed image, acquisition-data, and reconstruction paths behave correctly on a real NVIDIA GPU.

What I value most is how these changes secure the core invariants of the CUDA-array-view path: one place for capability semantics, a guarded contract for address access, one canonical pointer identity across the interface layers, one consistent Python handle-lifetime mechanism, and one explicit CUDA build contract in CI. **This is a critical and exceptionally valuable contribution to the reliability and architecture of this part of the codebase, and I am deeply grateful for the thought, care, and technical judgement you put into it!!!**

**THANK YOU WHOLEHEARTEDLY!!!**

.github/workflows/build-test.yml
```diff
@@ -112,38 +125,41 @@ jobs:
...
-        parallel-finished: true
+  # disabled as currently not functional
+  # coveralls_finish:
+  #  needs: build
+  #  runs-on: ubuntu-latest
+  #  steps:
+  #  - name: Coveralls Finished
+  #    uses: AndreMiras/coveralls-python-action@develop
+  #    with:
+  #      parallel-finished: true
```
casperdcl, line 165:
```suggestion
```

casperdcl (approved):

KrisThielemans (approved):

PR #1390, opened 2026-03-31
handle conveniences

casperdcl:
- [x] `HANDLE` convenience singleton (can't do `Handle.new` since classproperties are deprecated)
- [x] `functools.wraps` backend functions in `Handle.c*`
- [x] for discussion: what is the purpose of (unused) `check_stack`/`inspect.stack()`?

Follow-up to #1388

### DIFF
--- a/src/common/Utilities.py
+++ b/src/common/Utilities.py
@@ -113,6 +113,9 @@ class Handle:
         return obj
 
 
+HANDLE = Handle(None, -1) # for calling backend functions which do not require a handle argument
+
+
 def cpp_int_bits():
     """Returns the number of bits in a C++ integer."""
     return pyiutil.intBits()
@@ -152,7 +155,7 @@ def examples_data_path(data_type):
     Returns the path to PET/MR/Registration data used by SIRF/examples demos.
     data_type: either 'PET' or 'MR' or 'Registration'
     '''
-    path = Handle(None, -1).cSIRF_examples_data_path(data_type)
+    path = HANDLE.cSIRF_examples_data_path(data_type)
     return str(path)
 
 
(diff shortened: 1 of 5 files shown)


KrisThielemans (approved):
ok, but now might be a good time to add a little bit of documentation on the `Handle` class (and indeed the new `HANDLE`. It is an implementation detail (should we call it `_Handle` in Python tradition?), but developers do need to know how this works :-)

src/common/Utilities.py
```diff
@@ -36,7 +37,7 @@
 class Handle:
     def __init__(self, handle, check_stack: int | None = None):
         if (check_stack is None or check_stack >= 0) and pyiutil.executionStatus(handle) != 0:
-            check_stack = inspect.stack()[1 if check_stack is None else check_stack]
+            check_stack = inspect.stack()[1 if check_stack is None else check_stack] # TODO: delete or print this?
```
casperdcl, line 40: /CC @evgueni-ovtchinnikov

evgueni-ovtchinnikov: delete unused lines 40-43 please (not much help anyway as the error may be higher up the stack than level 1)

evgueni-ovtchinnikov: Re the purpose of `inspect.stack`: what I wanted was to get closer to the line that throws monkey wrench.

For example: running PET demos I get
```
`Error opening file acf.hs\n' exception caught at line 427 of /home/sirfuser/devel/buildVM/sources/SIRF/src/xSTIR/cSTIR/cstir.cpp
```
but actual failure happens in STIR function `ProjData::read_from_file` called either at line 694 or line 852 of SIRF's `stir_data_containers.h` depending on the storage scheme (file or memory).

PR #1400, opened 2026-05-12
tests: pass when missing torch

casperdcl:
Allows e.g. [SuperBuild `ctest` to pass](https://github.com/SyneRBI/SIRF-SuperBuild/pull/994) without installing `torch`.

### DIFF
--- a/src/torch/tests/test_null.py
+++ b/src/torch/tests/test_null.py
@@ -1,0 +1,4 @@
+def test_null():
+    """dummy test; always runs regardless of whether torch is found"""
+    # work-around for https://github.com/pytest-dev/pytest/pull/817
+    pass


KrisThielemans (approved):

paskino (changes requested):
Currently the reason why the SB failed was that `torch` wasn't installed on the CI so `pytest` didn't run any test; careful developers [decided](https://github.com/pytest-dev/pytest/pull/817) that in this situation you should fail catastrofically, but without a reasonable error message. 

We haven't understood the reason of the SB failure for a while, despite the hints in the coverage warning message.

Therefore, I would add this test to the `test_torch.py` with appropriate message regarding whether `torch` is installed or not.

casperdcl:
`test_torch.py` has the `importorskip` + lots of dependent imports & config at the top; refactoring it would be very painful (to allow a non-skipped test in  the same file).

Best to just keep the dummy test in a separate file.

